PR #24, opened 2019-05-08
Add put() to smart types exposing an operator&

sylveon:
This makes the interface feel more natural to C++/WinRT users, makes discovery of the feature easier (it shows up in Intellisense) as well as being explicit about the operation instead of implicit, so easier to notice when reading code because `operator&` normally isn't stateful.

### DIFF
--- a/include/wil/com.h
+++ b/include/wil/com.h
@@ -380,10 +380,21 @@ namespace wil
         //! apparent, prefer to scope variables as close to use as possible (generally avoiding use of the same com_ptr variable in successive calls to
         //! receive an output interface).
         //! @see addressof
-        pointer* operator&() WI_NOEXCEPT
+        pointer* put() WI_NOEXCEPT
         {
             reset();
             return &m_ptr;
+        }
+
+        //! Returns the address of the internal pointer (releases ownership of the pointer BEFORE returning the address).
+        //! The pointer is explicitly released to prevent accidental leaks of the pointer.  Coding standards generally indicate that
+        //! there is little valid `_Inout_` use of `IInterface**`, making this safe to do under typical use.  Since this behavior is not always immediately
+        //! apparent, prefer to scope variables as close to use as possible (generally avoiding use of the same com_ptr variable in successive calls to
+        //! receive an output interface).
+        //! @see addressof
+        pointer* operator&() WI_NOEXCEPT
+        {
+            return put();
         }
 
         //! Returns the address of the internal pointer (does not release the pointer; should not be used for `_Out_` parameters)
--- a/include/wil/com.h
+++ b/include/wil/com.h
@@ -380,6 +380,11 @@ namespace wil
         //! apparent, prefer to scope variables as close to use as possible (generally avoiding use of the same com_ptr variable in successive calls to
         //! receive an output interface).
         //! @see addressof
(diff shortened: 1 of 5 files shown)


ChrisGuzak:
nice. please add a test to demonstrate exercise this and demonstrate use. since this updates the implementation of the existing functionality and is so simple it could be argued that this is not needed but I want to establish the practice that _every change comes with tests_.

sylveon:
Where could I find existing tests for `operator&`?

jonwis:
Tests for resource types go into tests/ResourceTests.cpp ... Add a new test in the spirit of ResourceTests::TestScopeExit .

include/wil/com.h
```diff
@@ -380,12 +380,23 @@ namespace wil
         //! apparent, prefer to scope variables as close to use as possible (generally avoiding use of the same com_ptr variable in successive calls to
         //! receive an output interface).
         //! @see addressof
-        pointer* operator&() WI_NOEXCEPT
+        pointer* put() WI_NOEXCEPT
```
jonwis: Consider adding a little sample to the above comment:

```c++
/**
~~~~
// Call a method that returns a thing
STDAPI GetMuffin(IMuffin** muffin);
wil::com_ptr<IMuffin> myMuffin;
THROW_IF_FAILED(GetMuffin(myMuffin.put()));
~~~~
*/
```

include/wil/resource.h
```diff
@@ -292,13 +292,18 @@ namespace wil
             return storage_t::is_valid();
         }
 
-        pointer_storage *operator&() WI_NOEXCEPT
+        pointer_storage *put() WI_NOEXCEPT
         {
```
jonwis: As in com.h, consider a sample:

```c++
/**
~~~~
BOOL OpenOrCreateWaffle(PCWSTR name, HWAFFLE* handle);
wil::unique_any<HWAFFLE, decltype(&::CloseWaffle), ::CloseWaffle> waffle;
RETURN_IF_WIN32_BOOL_FALSE(OpenOrCreateWaffle(L"tasty.yum", waffle.put()));
~~~~
*/
```

dunhor:
For `com_ptr`, those tests are predominantly in `ComTests.cpp`. For `unique_*` types, those seem to be somewhat scattered, but primarily tested in `ResourceTests.cpp` and `wiTest.cpp` IIRC

sylveon:
I've added tests covering every type except `unique_com_token` because I couldn't even find tests for that.

tests/wiTest.cpp
```diff
@@ -989,6 +989,9 @@ TEST_CASE("WindowsInternalTests::UniqueHandle", "[resource][unique_any]")
 
         // address
         REQUIRE(*spMoveHandle.addressof() == handleValue);
+        REQUIRE(*spMoveHandle.put() == nullptr);
+        *spMoveHandle.put() = ::CreateEventEx(nullptr, nullptr, CREATE_EVENT_INITIAL_SET, 0);
```
ChrisGuzak: consider calling the W version directly, avoiding the macros

sylveon: The whole test suite uses the macros, the same line but for `operator&` (which I copied from) also uses the macro.

PR #31, opened 2019-05-13
constify variables

matbech:
This fixes the compiler warnings such as:
Warning	C26496	The variable '__hr' is assigned only once, mark it as const (con.4).


dunhor:
Probably an unpopular opinion, but I find the CppCoreChecks to be _way_ too pedantic. IMO we should just disable them, either completely or selectively.

ChrisGuzak:
this seems good to me, easy, safe and will eliminate this warning. Duncan, please give feedback on the checkers to the owners, I can connect you.

jonwis:
I've pushed https://github.com/microsoft/wil/tree/jonwis/more-consty which has the additional pragmas required to hush the `conditional is const, consider if constexpr` warnings inside the WIL macros.

include/wil/result_macros.h
```diff
@@ -569,24 +569,32 @@ WI_ODR_PRAGMA("WIL_FreeMemory", "0")
 #define __RFF_FN(MethodName)                                  __RFF_NS:: MethodName
 #endif
 // end-of-repeated fail-fast handling macros
+
+#if defined(_MSVC_LANG)
+#define __WI_SUPPRESS_4127_S __pragma(warning(push)) __pragma(warning(disable:4127))
```
BenJKuhn: Just curious, why not simplify with __pragma(warning(suppress:4127)), which wouldn't need an end tag?

jonwis: The [documentation for warning-suppress](https://docs.microsoft.com/en-us/cpp/preprocessor/warning?view=vs-2019) indicates that the `#pragma` applies to the _following_ line. The `__pragma` form appears to suppress it on both this line and the following line:

https://godbolt.org/z/12GZkp

```c++
const int kitten = 6;
const int poptart = 6;
const int muffin = 7;
void toast();
void foo() {
    __pragma(warning(suppress:4127)) if (kitten == poptart) { toast(); }
    if (muffin == poptart) { toast(); }
    if (muffin == kitten) { toast(); }
}
```

Only the `muffin == kitten` line shows an error, when `muffin == poptart` should also show an error.

PR #33, opened 2019-05-14
Specialize wstd::unique_ptr<> for UnmapViewOfFile

MicheleCicciottiWork:


dunhor:
Part of me likes the ability to specify a type (e.g. `char` or `wchar_t`) when mapping in a file, but it feels like it would be a nuisance for consumers who don't want that behavior. E.g. if all they want is an RAII type to free a mapped file that contains binary data, in which case a specialization of `unique_any` would be more appropriate for them.

sylveon:
Why not both?

include/wil/resource.h
```diff
@@ -3430,6 +3430,15 @@ namespace wil
         }
     };
 
+    struct mapview_deleter
```
dunhor: This can just be replaced with `function_deleter<...>`

dunhor:
That works too. E.g. similar to how there's both `unique_hlocal` and `unique_hlocal_ptr`

include/wil/resource.h
```diff
@@ -3454,6 +3463,12 @@ namespace wil
     template<typename T>
     using unique_virtualalloc_ptr = wistd::unique_ptr<T, virtualalloc_deleter>;
 
+    /** Manages a typed pointer allocated with MapViewOfFile
+    A specialization of wistd::unique_ptr<> that frees via UnmapViewOfFile(p).
+    */
+    template<typename T>
+    using unique_mapview_ptr = wistd::unique_ptr<T, mapview_deleter>;
```
jonwis: It'd be neat to also have something in win32_helpers.h which made `MapViewOfFile` easier, like:

```cpp
HRESULT map_file_view_nothrow(HANDLE file, DWORD access, const UINT64& offset, size_t size, unique_mapview& view)
{
    ULARGE_INTEGER offsetInt;
    offsetInt.QuadPart = offset;
    view.reset(::MapViewOfFile(file, access, offsetInt.LowPart, offsetInt.HighPart, size));
    RETURN_LAST_ERROR_IF_NULL(view);
    return S_OK;
}
unique_mapview map_file_view(HANDLE file, DWORD access, const UINT64& offset)
{
    unique_mapview view;
    THROW_IF_FAILED(map_file_view_nothrow(file, access, offset));
    return view;
}
```
And nothrow ... and maybe the `_ptr<T>` variant.

include/wil/resource.h
```diff
@@ -3454,6 +3463,12 @@ namespace wil
     template<typename T>
     using unique_virtualalloc_ptr = wistd::unique_ptr<T, virtualalloc_deleter>;
 
+    /** Manages a typed pointer allocated with MapViewOfFile
+    A specialization of wistd::unique_ptr<> that frees via UnmapViewOfFile(p).
+    */
+    template<typename T>
```
ChrisGuzak: seem this should default to void as that is the common case (in my impl of this that is what I used).
`template<typename T = void>`

PR #34, opened 2019-05-14
Add unique_hhook

sylveon:
This is for SetWindowsHookEx and UnhookWindowsHookEx.

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -4179,6 +4179,7 @@ namespace wil
     typedef unique_any<HACCEL, decltype(&::DestroyAcceleratorTable), ::DestroyAcceleratorTable> unique_haccel;
     typedef unique_any<HCURSOR, decltype(&::DestroyCursor), ::DestroyCursor> unique_hcursor;
     typedef unique_any<HWND, decltype(&::DestroyWindow), ::DestroyWindow> unique_hwnd;
+    typedef unique_any<HHOOK, decltype(&::UnhookWindowsHookEx), ::UnhookWindowsHookEx> unique_hhook;
 #endif // __WIL__WINUSER_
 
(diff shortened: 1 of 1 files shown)


include/wil/resource.h
```diff
@@ -4179,6 +4179,7 @@ namespace wil
     typedef unique_any<HACCEL, decltype(&::DestroyAcceleratorTable), ::DestroyAcceleratorTable> unique_haccel;
     typedef unique_any<HCURSOR, decltype(&::DestroyCursor), ::DestroyCursor> unique_hcursor;
     typedef unique_any<HWND, decltype(&::DestroyWindow), ::DestroyWindow> unique_hwnd;
+    typedef unique_any<HHOOK, decltype(&::UnhookWindowsHookEx), ::UnhookWindowsHookEx> unique_hhook;
```
dunhor: This should require a `WINAPI_FAMILY_PARTITION` check, but then again so should the ones above... I'd say don't worry about it; just looks like we've got some cleanup work to do later

BenJKuhn: HHOOK and HWND are both defined in app and desktop partitions, even though most APIs require desktop, so this is fine.

dunhor: Ahh, okay, I see now. This is all wrapped inside of an `#if defined(__WIL_WINBASE_) && !defined(__WIL_WINBASE_DESKTOP) && WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)` which already has the partition check

oldnewthing:
@sylveon Please also update [documentation](https://github.com/microsoft/wil/wiki/RAII-resource-wrappers).

sylveon:
@oldnewthing Thanks for the reminder, done.

PR #35, opened 2019-05-14
Conformance mode includes

MicheleCicciottiWork:
Added the appropriate `#if`s and `#include`s for `/permissive-`

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -8,6 +8,8 @@
 //    PARTICULAR PURPOSE AND NONINFRINGEMENT.
 //
 //*********************************************************
+#include <stdint.h> // SIZE_MAX
+
 #include "result_macros.h"
 #include "wistd_functional.h"
 #include "wistd_memory.h"
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -3738,6 +3738,7 @@ namespace wil
     template <typename T>
     using unique_hlocal_ptr = wistd::unique_ptr<T, hlocal_deleter>;
 
+#if defined(__PLACEMENT_NEW_INLINE)
     /** Provides `std::make_unique()` semantics for resources allocated with `LocalAlloc()` in a context that may not throw upon allocation failure.
     Use `wil::make_unique_hlocal_nothrow()` for resources returned from APIs that must satisfy a memory allocation contract that requires the use of `LocalAlloc()` / `LocalFree()`.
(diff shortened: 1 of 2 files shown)


include/wil/resource.h
```diff
@@ -3736,6 +3738,7 @@ namespace wil
     template <typename T>
     using unique_hlocal_ptr = wistd::unique_ptr<T, hlocal_deleter>;
 
+#if defined(__PLACEMENT_NEW_INLINE)
```
dunhor: This does not protect against multiple includes. E.g. including `resource.h`, then including `new`, then including `resource.h` again.

include/wil/result.h
```diff
@@ -11,6 +11,8 @@
 #ifndef __WIL_RESULT_INCLUDED
 #define __WIL_RESULT_INCLUDED
 
+#include <new> // operator new(size_t, void*)
```
dunhor: I suspect that this requires the ability to suppress (e.g. see `result_macros.h`)

include/wil/resource.h
MicheleCicciottiWork: It's not supposed to. What it does is preventing compilation errors in `/permissive-` mode when `<wil/resource.h>` is included before `<new>`

dunhor: What I'm saying is that it _needs_ to be able to support that.

include/wil/result.h
MicheleCicciottiWork: If we honored `WIL_SUPPRESS_NEW` in `<wil/result.h>`, like this:
```suggestion
#if !defined(WIL_SUPPRESS_NEW)
#include <new> // operator new(size_t, void*)
#endif
```
we would be unable to use placement `new` in `ProcessLocalStorageData<T>`, and in turn this would break `ProcessLocalStorage<T>`, which would break `g_processLocalData`. A possible alternative to `#include <new>` is reimplementing placement `new`  under `wistd`, since the implementation is trivial

MicheleCicciottiWork:
Nevermind, include guards won't work

dunhor:
Out of curiosity, what build configuration were you seeing errors?

include/wil/resource.h
ChrisGuzak: not obvious but wil/resource.h requires being included after the types it supports. and sometimes that means being included multiple times.

PR #37, opened 2019-05-14
Copy definition of placement new to <wil/wistd_memory.h>

MicheleCicciottiWork:
Elsewhere in WIL, placement `new` is used regardless of whether `<new>` is included or not. Since we can't safely `#include <new>`, we copy its definition of placement `new` (and placement `new[]`) into `<wil/wisdt_memory.h>`


MicheleCicciottiWork:
Nevermind. This library is more complex than I thought!

PR #43, opened 2019-05-17
Add and use wistd::invoke

### LINKED ISSUE #36: std::invoke for callables.
C++17 allows some pretty crazy things with templates. Observe the following:

```cpp
template <typename T>
struct member_parent { };

template<typename Object, typename Return, typename... Args>
struct member_parent<Return(Object:: *)(Args...)> {
	typedef Object type;
};

template<typename T>
using member_parent_t = typename member_parent<T>::type;

template<auto close_fn, typename token_t, token_t invalid_token = token_t()>
using com_token = wil::unique_com_token<member_parent_t<decltype(close_fn)>, token_t, decltype(close_fn), close_fn, invalid_token>;
```

This code allows to simplify writing a custom COM token:
```cpp
using app_visibility_sink_cookie = com_token<&IAppVisibility::Unadvise, DWORD>;
```

However, it can't be easily included in wil without depending on C++17 (we can't overload templated type aliases, C++ determined that would be too easy).

The code should work, but doesn't, because of this line: https://github.com/microsoft/wil/blob/ade910fb44f0b23c03111a2838c2148d326f5622/include/wil/resource.h#L1370

If that line where to be changed to `std::invoke(close_fn, m_source, m_token);`, this would work. However, I am aware that wil tries to avoid dependencies on the system standard library, and I am honestly unsure on how I would proceed to add more stuff to `wistd`.

Ideally, replacing all of the calls of function pointer template parameters like above by calls to `wistd::invoke` would allow more creativity in the use of wil.

sylveon:
Fixes #36

### DIFF
--- a/include/wil/wistd_functional.h
+++ b/include/wil/wistd_functional.h
@@ -516,9 +516,16 @@ namespace wistd     // ("Windows Implementation" std)
     swap_wil(function<_Rp(_ArgTypes...)>& __x, function<_Rp(_ArgTypes...)>& __y)
     {return __x.swap(__y);}
 
+    // std::invoke
+    template <class _Fp, class... _Args>
+    inline __WI_LIBCPP_INLINE_VISIBILITY
+    auto
+    invoke(_Fp &&__f, _Args &&...__args) -> decltype(__invoke(wistd::forward<_Fp>(__f), wistd::forward<_Args>(__args)...))
+    {return __invoke(wistd::forward<_Fp>(__f), wistd::forward<_Args>(__args)...);}
+
 #else // __WI_LIBCPP_CXX03_LANG
 
-#error wistd::function not implemented for pre-C++11
+#error wistd::function and wistd::invoke not implemented for pre-C++11
 
 #endif
 }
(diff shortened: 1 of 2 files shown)


include/wil/wistd_functional.h
```diff
@@ -516,9 +516,16 @@ namespace wistd     // ("Windows Implementation" std)
     swap_wil(function<_Rp(_ArgTypes...)>& __x, function<_Rp(_ArgTypes...)>& __y)
     {return __x.swap(__y);}
 
+    // std::invoke
```
dunhor: Please copy the implementation from libc++

sylveon: The libc++ implementation uses invoke_result_t, which requires C++14.

dunhor: Feel free to reduce that limitation as there's no requirement that `invoke_result` be limited to C++17, apart from that's when it was introduced to the standard. Could also use `__invoke_of`

sylveon: Fixed

ChrisGuzak:
update the description, "this set wil up to support [callables ](https://en.cppreference.com/w/cpp/named_req/Callable) instead of just functions in the resource cleanup"

sylveon:
Description of what?

ChrisGuzak:
the PR description.

ChrisGuzak:
I'm not github PR savvy, is there a description independent of the title?

sylveon:
Ah yes there is. Kinda pointless to change the description since it's merged now but I updated it anyways.

PR #46, opened 2019-05-21
Sync with OS copy

dunhor:
This is sync'd with commit 9ebe337fed9899a3ca4dae71e713fad8ba6e4b2a

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -2458,7 +2458,10 @@ namespace wil
     typedef unique_any_handle_null_only<decltype(&::CloseHandle), ::CloseHandle> unique_process_handle;
 
     typedef unique_struct<TOKEN_LINKED_TOKEN, decltype(&details::CloseTokenLinkedToken), details::CloseTokenLinkedToken> unique_token_linked_token;
+
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM)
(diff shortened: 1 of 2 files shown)


dunhor:
FYI: I'm planning on updating the NuGet once submitted

sylveon:
Would it also be possible to create a release on GitHub so that the copy on vcpkg can pin to that instead of a random commit?

dunhor:
I suppose we can. @ChrisGuzak any thoughts on a versioning scheme? I was just thinking of just using the date e.g. `20190521` or something like that. I can't really imagine that we'd have a need for `1.X`, `2.X`, etc. releases

ras0219-msft:
If you'd prefer that we follow a rolling-release/evergreen/live-at-head/master model, that's fine by vcpkg as well. We just need clear guidance :)

sylveon:
That'd actually be preferable for me, since I usually make changes to use them straight away in my project (I installed via vcpkg but then just symlinked the installation location to my local clone of wil so that I can git pull the latest changes myself).

ras0219-msft:
I didn't completely follow your setup, however our `--head` switch _might_ be useful for you:
```
vcpkg remove wil
vcpkg install wil --head
```
Will download and install the latest master of `wil` at any point in time.

As for the rolling release, we wouldn't silently upgrade underneath users, but we have ways to automate publishing updates to the vcpkg catalog.

dunhor:
WIL definitely fits more into the "live at head" model

PR #47, opened 2019-05-22
Update cppwinrt.h to use the new C++/WinRT 2.0 exception translation techniques

### LINKED ISSUE #42: Changing linker order will cause tests to fail
Change cpplatest source listing to:
```
target_sources(witest.cpplatest PUBLIC
    ${CMAKE_CURRENT_SOURCE_DIR}/../CommonTests.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../ComTests.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../FileSystemTests.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../main.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../ResourceTests.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../ResultTests.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../SafeCastTests.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../WistdTests.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../wiTest.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../CppWinRTTests.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../StlTests.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../TokenHelpersTests.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../UniqueWinRTEventTokenTests.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../WatcherTests.cpp
    ${CMAKE_CURRENT_SOURCE_DIR}/../WinRTTests.cpp
    )
```
compile and run tests. It will fail.

dunhor:
There's a few changes here:
* Updates WIL to use a global function pointer for `winrt::hresult_error` -> `HRESULT` translation for the WIL helpers (`ResultFromCaughtException`, etc.). This is done unconditionally and independently of the C++/WinRT version
* Updates WIL to set `winrt_to_hresult_handler` if the C++/WinRT version is >= 2.0. We still define `WINRT_EXTERNAL_CATCH_CLAUSE` since it's non-trivial to conditionalize that, but C++/WinRT should (soon) feel free to remove support for that macro.

I was also tempted to add support for `SupportedExceptions::Thrown` and `SupportedExceptions::ThrownOrAlloc` to match the behavior for `Platform::Exception`, but eventually decided not to since C++/WinRT seems to translate some errors to `std::*` exceptions. Currently this is only `std::bad_alloc` for `E_OUTOFMEMORY` - so in theory supporting `SupportedExceptions::ThrownOrAlloc` should be fine - but still opted to keep things as-is.

There's also a minor loss in functionality. `winrt::to_hresult` translates a few more `std::*` exceptions (e.g. `std::invalid_argument`); the difference is that now those will get mapped as `HRESULT_FROM_WIN32(ERROR_UNHANDLED_EXCEPTION)`. Shouldn't be hard to go in add support for these in WIL, if desired.

fixes #42 (well, technically only fixes it for newer SDK versions ;))

### DIFF
--- a/include/wil/cppwinrt.h
+++ b/include/wil/cppwinrt.h
@@ -21,7 +21,7 @@
 // C++/WinRT "2.0" this was accomplished by injecting the WINRT_EXTERNAL_CATCH_CLAUSE macro - that WIL defines below -
 // into its exception handler (winrt::to_hresult). Starting with C++/WinRT "2.0" this mechanism has shifted to a global
 // function pointer - winrt_to_hresult_handler - that WIL sets automatically when this header is included and
-// 'RESULT_SUPPRESS_STATIC_INITIALIZERS' is not defined.
+// 'CPPWINRT_SUPPRESS_STATIC_INITIALIZERS' is not defined.
 
 /// @cond
 namespace wil::details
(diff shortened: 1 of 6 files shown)


include/wil/cppwinrt.h
```diff
@@ -16,47 +16,70 @@
...
-// WIL and C++/WinRT use two different exception types for communicating HRESULT failures. Thus, both libraries need to
-// understand how to translate these exception types into the correct HRESULT values at the ABI boundary. C++/WinRT
-// accomplishes this by injecting the WINRT_EXTERNAL_CATCH_CLAUSE macro - that WIL defines below - into its exception
-// handler. WIL accomplishes this by detecting this file's inclusion in result_macros.h and modifies its behavior to
-// account for the C++/WinRT exception type.
+#ifdef CPPWINRT_VERSION
+// Prior to C++/WinRT "2.0" this header needed to be included before 'winrt/base.h' so that our definition of
+// 'WINRT_EXTERNAL_CATCH_CLAUSE' would get picked up in the implementation of 'winrt::to_hresult'. This is no longer
+// problematic, so only emit an error when using a version of C++/WinRT prior to 2.0
+static_assert(::wil::details::major_version_from_string(CPPWINRT_VERSION) >= 2,
```
jonwis: Instead of this string parsing, could cppwinrt produce a `CPPWINRT_VERSION_MAJOR` macro which is a numeric value 2 or similar? I guess that'd require a coordinated change between the repos ... will the version identification mechanism of cppwinrt ever change?

kennykerr: Right, it's one more thing we have to version check. This version string has been around for ages so it's rather convenient and avoids introducing yet another macro. And when C++ modules finally comes along we'll be completely macro free, so adding macros is probably not the best strategy in the long run.

dunhor: I'll let Kenny be the authority on which macros are present. But I agree that we should ideally future-proof ourselves for modules, in which case I guess we'll move to something like checking `winrt::version_string` (or integers, but neither are usable with the preprocessor). But by that time, hopefully we can remove all of the back-compat stuff.

PR #48, opened 2019-05-23
Update readme with vcpkg info

dunhor:
Folks in VS pointed out that the readme hadn't been updated to include information regarding WIL's availability with vcpkg.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,10 +33,19 @@ and some (such as the RAII resource wrappers) can even be used in kernel mode.
 
 This project is documented in [its GitHub wiki](https://github.com/Microsoft/wil/wiki). Feel free to contribute to it!
 
-# Consuming WIL via NuGet
-You can consume WIL via a NuGet package. To do so, follow the instructions on [nuget.org](https://www.nuget.org/packages/Microsoft.Windows.ImplementationLibrary).
-This package includes the header files under the [include](include) directory as well as a [.targets](packaging/nuget/Microsoft.Windows.ImplementationLibrary.targets)
-file.
+# Consuming WIL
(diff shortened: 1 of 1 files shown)


KindDragon:
@sylveon you can build latest commit using `vcpkg install <package> --head`

sylveon:
Nice, I wasn't aware of that. This should totally be mentioned in the readme then.

dunhor:
@KindDragon I'm a bit confused by the expected workflow when using `--head`. For one it doesn't seem possible to "upgrade" an existing package to reference HEAD (at least `vcpkg upgrade` doesn't seem to have any `--head`-equivalent argument and running `vcpkg install` again with `--head` doesn't seem to do it either). AFAICT you have to remove and re-install the package with `--head` for it to work. Not a terrible limitation, but not ideal either.

But what's more annoying is that even if I _do_ install a package with `--head`, I can't seem to figure out how to update it. If I run `vcpkg update` it tells me `Using local portfile versions. To update the local portfiles, use 'git pull'`. Okay, maybe it's doing some special voodoo that hooks into `git pull`, but running `git pull` just seems to no-op (so no voodoo). If I run `vcpkg upgrade` it tells me I need `--no-dry-run` which, if I re-run with that, will revert the package to the version that vcpkg has on record. AFAICT this also requires removing and re-installing the package.

Perhaps `upgrade` has a special argument that I'm missing, but there's no help topic for `upgrade`. Is this expected?

ras0219-msft:
`--head` is intended only as a temporary measure to enable authors to quickly ship fixes for users; the idea is that if someone reports an issue, you can say "Fixed on master -- use `vcpkg install --head` until vcpkg updates to the new version".

For most users, we will be pulling the master commits from here into our catalog regularly (with testing), so we'd recommend leaving the README as just `vcpkg install wil:x64-windows`.

dunhor:
Sounds good. I'll leave it as something for users to discover for themselves if they choose to use that functionality

PR #50, opened 2019-05-25
Add missing shared/weak pointers for hhook

sylveon:


include/wil/resource.h
```diff
@@ -4954,6 +4956,16 @@ namespace wil
     typedef weak_any<shared_cotaskmem_string> weak_cotaskmem_string;
 #endif // __WIL_OBJBASE_H_STL
 
+#if defined(_SHLOBJ_H_) && !defined(__WIL_SHLOBJ_H__) && WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+#define __WIL_SHLOBJ_H__
+    typedef unique_any<LPITEMIDLIST, decltype(&::ILFree), ::ILFree> unique_itemidlist;
```
ChrisGuzak: ILFree() is defined to call CoTaskMemFree() so it is not needed.

ChrisGuzak: LPITEMIDLIST is a problematic type. if you use [STRICT_TYPED_ITEMIDS](https://docs.microsoft.com/en-us/windows/desktop/api/shtypes/ns-shtypes-_itemidlist) it is re-defined as 3 distinct types, absolute, relative, child to create some type safety. all clients should use that to avoid common errors. Given that I'd like to see this support exclusive to that mode.

sylveon: In [the documentation](https://docs.microsoft.com/en-us/windows/desktop/api/shlobj_core/nf-shlobj_core-ilfree), this is only mentioned for Windows 2000, which makes the use of CoTaskMemFree effectively implementation-defined behavior on anything more recent.

sylveon: Actually, I might just remove that entirely, since `wil::unique_cotaskmem_ptr<ITEMIDLIST_*>` can be used instead.

sylveon:
I've rebased this pull request to only add the missing shared and weak pointers since using wil::unique_cotaskmem_ptr for ITEMIDLIST is a better idea. Please take a look again.

Issue #49, opened 2019-05-24
Add error handling for windows RPC APIs (that use SEH)


ChrisGuzak:
A feature request from internal windows teams.

RETURN_IF_RPC_FAILED/THROW_IF_RPC_FAILED/FAIL_FAST_IF_RPC_FAILED macros that would use a lambda to create a function scope that enables the use of the __try/__catch/__finally language extensions needed to consume RPC based APIs.

#define RETURN_IF_RPC_FAILED(hr) \
do \
{ \
HRESULT __hrRet = [&]() -> HRESULT \
{ \
RpcTryExcept \
{ \
HRESULT __hrRet = wil::verify_hresult(hr); \
if (FAILED(__hrRet)) \
{ \
__RETURN_HR_FAIL(__hrRet, #hr); \
} \
} \
RpcExcept(I_RpcExceptionFilter(RpcExceptionCode())) \
{ \
__RETURN_WIN32_FAIL(RpcExceptionCode(), #hr); \
} \
RpcEndExcept \
return S_OK; \
}(); \
} \
while (0, 0)

jonwis:
My github-fu is weak, but I started https://github.com/microsoft/wil/compare/user/jonwis/rpc with the method-wrapper variant.

I'm not happy with it yet. Feels like I should be able to use the new `wistd::invoke` to take both lambdas and methods-with-parameters:

```c++
// Lambda form:
wil::call_rpc_nothrow([&] { return MyRpcMethod(m_connection, a, b, c); });

// Parameters form:
wil::call_rpc_nothrow(MyRpcMethod, a, b,c);
```

I'm also struggling on the throwing 'with result' form. Seems like I shouldn't have to specify the GUID, and some traits-magic could figure it out:

```c++
// Current
GUID g = wil::call_rpc<GUID>(MethodThatReturnsGuid, a, b, c);

// Preferred
GUID g = wil::call_rpc(MethodThatReturnsGuid, a, b, c);
```

kennykerr:
[...]

PR #51, opened 2019-05-27
Add RPC helpers

jonwis:
This resolves issue #49.

See also [MSDN's page on RPC exceptions](https://docs.microsoft.com/en-us/windows/desktop/Rpc/exception-handling). Structured exceptions and C++ exceptions don't play well together.  This adds `wil::invoke_rpc` and `wil::invoke_rpc_result` to convert raised exceptions into HRESULTs, both throwing and nonthrowing variants.

Also included are `wil::unique_midl_ptr` and `wil::unique_midl_string` for working with types returned by RPC calls allocated via [MIDL_user_allocate](https://docs.microsoft.com/en-us/windows/desktop/Rpc/the-midl-user-allocate-function).


jonwis:
Question for readers - since this is tied to Win32 RPC, should this instead go into `win32_helpers.h` instead of adding a completely new header?

include/wil/resource.h
```diff
@@ -4747,6 +4747,40 @@ namespace wil
...
+#define __WIL__RPCNDR_H__
+
+    //! Function deleter for use with pointers allocated by MIDL_user_allocate
+    using midl_deleter = function_deleter<decltype(&::MIDL_user_free), MIDL_user_free>;
+
+    //! Unique-ptr holding a type allocated by MIDL_user_alloc or returned from an RPC invocation
+    template<typename T> using unique_midl_ptr = wistd::unique_ptr<T, midl_deleter>;
+
+    //! Unique-ptr for strings allocated by MIDL_user_alloc
+    using unique_midl_string = unique_midl_ptr<WCHAR>;
```
ChrisGuzak: WCHAR -> wchar_t

include/wil/resource.h
```diff
@@ -4747,6 +4747,40 @@ namespace wil
...
+#ifndef WIL_NO_ANSI_STRINGS
+    using unique_midl_ansistring = unique_midl_ptr<CHAR>;
```
ChrisGuzak: CHAR -> char

include/wil/rpc_helpers.h
```diff
@@ -0,0 +1,170 @@
...
+
+    For example, consider an RPC interface method defined in idl as:
+    ~~~
+    HRESULT GetKittenState([in, ref, string] const wchar_t* name, [out, retval] KittenState** state);
+    ~~~
+    To call this method, use:
+    ~~~
+    wil::unique_rpc_binding binding = // typically gotten elsewhere;
+    wil::unique_midl_ptr<KittenState> state;
+    HRESULT hr = wil::invoke_rpc_nothrow(GetKittenState, binding.get(), L"fluffy", wil::out_param(state));
```
ChrisGuzak: I see it would be nice to have invoke_rpc_nothrow on unique_rpc_binding
`binding.invoke_rpc_nothrow(GetKittenState, L"fluffy", wil::out_param(state));`

ChrisGuzak: use `state.put()` as that makes this a little simpler.

include/wil/rpc_helpers.h
```diff
@@ -0,0 +1,170 @@
...
+    ~~~
+    */
+    template<typename... TCall> HRESULT invoke_rpc_nothrow(TCall&&... args) WI_NOEXCEPT
+    {
+        RpcTryExcept
+        {
+            // Note: this helper type can be removed with C++17 enabled via
+            // 'if constexpr(wistd::is_same_v<void, result_t>)'
+            using result_t = typename wistd::__invoke_of<TCall...>::type;
+            HRESULT hr = details::call_adapter<result_t>::call(wistd::forward<TCall>(args)...);
```
ChrisGuzak: `const HRESULT hr ...`
but it seems this is better
```
RETURN_IF_FAILED(details::call_adapter...)
return S_OK;
```

include/wil/rpc_helpers.h
ChrisGuzak: it seems the function and its params should be together, so move the binding to the first param.

jonwis: I wish we had extension methods for this. 

Sadly, there's [no requirement that the binding handle be first](https://docs.microsoft.com/en-us/windows/desktop/Rpc/explicit-binding-handles). Our RPC implementation allows it anywhere.  In some RPC methods, the binding handle is even [implicit](https://docs.microsoft.com/en-us/windows/desktop/Rpc/implicit-binding-handles)

include/wil/rpc_helpers.h
```diff
@@ -0,0 +1,170 @@
...
+            template<typename... TArgs> static HRESULT call(TArgs&& ... args)
+            {
+                wistd::invoke(wistd::forward<TArgs>(args)...);
+                return S_OK;
+            }
+        };
+
+        // Some RPC exceptions are already HRESULTs. Others are in the regular Win32
+        // error space. If the incoming exception code isn't an HRESULT, wrap it.
+        constexpr static HRESULT map_rpc_exception(DWORD code)
```
dunhor: `static` -> `inline`. Or, since this function is `constexpr`, just remove `static`

include/wil/rpc_helpers.h
```diff
@@ -0,0 +1,170 @@
...
+    ~~~
+    To call this method, use:
+    ~~~
+    wil::unique_rpc_binding binding = // typically gotten elsewhere;
+    wil::unique_midl_ptr<KittenState> state;
+    HRESULT hr = wil::invoke_rpc_nothrow(GetKittenState, binding.get(), L"fluffy", wil::out_param(state));
+    RETURN_IF_FAILED(hr);
+    ~~~
+    */
+    template<typename... TCall> HRESULT invoke_rpc_nothrow(TCall&&... args) WI_NOEXCEPT
```
dunhor: Does this not give you a build error? You can't mix structured and C++ exceptions (I'm assuming that `RpcTryExcept` is essentially a `__try`, etc.?)

dunhor: Well that's interesting. It'll give me errors if I try and use `try...catch` in a function that uses `__try...__except`, but doesn't give the same errors - and works correctly - in a function that uses `noexcept`. I suppose that whatever flag they set on the function's handler info is compatible with the SEH version.

include/wil/rpc_helpers.h
jonwis: See above for why that's a problem; Microsoft's RPC implementation does not require the first parameter be the binding.

include/wil/rpc_helpers.h
jonwis: Fixed

include/wil/rpc_helpers.h
jonwis: Fixed

include/wil/rpc_helpers.h
jonwis: Fixed

include/wil/resource.h
jonwis: Fixed

include/wil/resource.h
jonwis: Fixed

oldnewthing:
@jonwis Please update wiki.

* Add `unique_midl_ptr` to [Supporting types](https://github.com/microsoft/wil/wiki/RAII-resource-wrappers#supporting-types)
* Add `unique_midl_string`, `unique_midl_ansistring` to [Unique-any handle types](https://github.com/microsoft/wil/wiki/RAII-resource-wrappers#available-unique_any-handle-types)
* Add new page for `rpc_helpers.h`.

jonwis:
Added to the types lists and created https://github.com/microsoft/wil/wiki/RPC-helpers

include/wil/rpc_helpers.h
```diff
@@ -0,0 +1,170 @@
...
+    ~~~
+    To call this method, use:
+    ~~~
+    wil::unique_rpc_binding binding = // typically gotten elsewhere;
+    GUID id;
+    HRESULT hr = wil::invoke_rpc_result_nothrow(id, GetKittenId, binding.get(), L"fluffy");
+    RETURN_IF_FAILED(hr);
+    ~~~
+    */
+    template<typename TResult, typename... TCall> HRESULT invoke_rpc_result_nothrow(TResult& result, TCall&&... args) WI_NOEXCEPT
```
oldnewthing: If the `TCall...` requires conversions to match the arguments of the invoked function, those conversions will take place **inside** the `RpcTryExcept`, and if the conversions encounter C++ exceptions, crazy things will happen: The `RpcExceptionFilter` will say "Sure, I'll handle that exception" and the function will simply return `0xE06d7363` instead of propagating the C++ exception.

We need to perform the conversions outside the `RpcTryExcept` so the RPC exception handler won't swallow them.

One option is to convert it ourselves. This means that if an exception does occur, we will `std::terminate` because this function is marked `noexcept`.

```cpp
template<typename TResult, typename... TExpectedArgs, typename... TActualArgs>
HRESULT invoke_rpc_result_nothrow(
    TResult& result,
    TResult(__stdcall*function)(TExpectedArgs...),
    TActualArgs... args)
{
    // C++ exception possible during conversion of arguments
    std::tuple<TExpectedArgs...> converted{ std::forward<TActualArgs>(args)... };
    RpcTryExcept
    {
        result = std::apply(function, std::move(converted));
        return S_OK;
    }
    RpcExcept(RpcExceptionFilter(RpcExceptionCode()))
    {
        RETURN_HR(details::map_rpc_exception(RpcExceptionCode()));
    }
    RpcEndExcept
}
```

Another option is to disallow conversions in the arguments and require the caller to preconvert them. This allows us to remain `WI_NOEXCEPT`, and might even be more in the spirit of WIL: If you want the `nothrow` version, you can't do things that might throw.

```cpp
template<typename TResult, typename... TArgs>
HRESULT invoke_rpc_result_nothrow(
    TResult& result,
    TResult(__stdcall*function)(TArgs...),
    TArgs... args) WI_NOEXCEPT
{
    RpcTryExcept
    {
        result = wistd::invoke(wistd::forward<TCall>(args)...);
        return S_OK;
    }
    RpcExcept(RpcExceptionFilter(RpcExceptionCode()))
    {
        RETURN_HR(details::map_rpc_exception(RpcExceptionCode()));
    }
    RpcEndExcept
}
```

It does create extra work on the caller to do the preconversion.

```cpp
extern HRESULT GetMuffin(IMuffin** muffin);

wil::com_ptr_nothrow<IMuffin> muffin;
// does not compile
invoke_rpc_result_nothrow(hr, GetMuffin, wil::out_param(muffin));

// you have to preconvert, which is cumbersome
invoke_rpc_result_nothrow(hr, GetMuffin, static_cast<IMuffin**>(wil::out_param(muffin)));
```

jonwis: Filed #59 to resolve.

PR #54, opened 2019-06-07
Sync with OS repo

dunhor:
Learning from past mistakes, I'll wait for a successful OS build before completing this time

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -4208,7 +4208,9 @@ namespace wil
     typedef shared_any<unique_hwinsta> shared_hwinsta;
 #endif // !defined(NOGDI) && !defined(NODESKTOP)
     typedef shared_any<unique_hwnd> shared_hwnd;
+#if !defined(NOUSER) && !defined(NOWH)
     typedef shared_any<unique_hhook> shared_hhook;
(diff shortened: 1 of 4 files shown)


include/wil/resource.h
```diff
@@ -4181,7 +4183,9 @@ namespace wil
     typedef unique_any<HACCEL, decltype(&::DestroyAcceleratorTable), ::DestroyAcceleratorTable> unique_haccel;
     typedef unique_any<HCURSOR, decltype(&::DestroyCursor), ::DestroyCursor> unique_hcursor;
     typedef unique_any<HWND, decltype(&::DestroyWindow), ::DestroyWindow> unique_hwnd;
+#if !defined(NOUSER) && !defined(NOWH)
```
sylveon: This should be added to {shared,weak}_hhook

dunhor: It's probably good to have more of these, but the definition of `unique_hhook` is guarded with the same guards as `unique_hwnd`, so it should be fine, at least so far as not adding any new build breaks

dunhor: 🤦🏻‍♂️I was looking at the wrong version of the file

PR #56, opened 2019-06-11
Enable '-fno-delayed-template-parsing' and other various conformance changes

### LINKED ISSUE #10: Enable '-fno-delayed-template-parsing' when building with Clang
There's a few outstanding issues that I've yet to address, so it can't be turned on just yet, but we should get this complete when we can. Both the "normal" and "cpplatest" tests build successfully with this set, but there's still work to do for the "app" and "noexcept" variants.

It may also be worth our while to address what we can with the `-fno-ms-compatibility` switch turned on while we're at it.

dunhor:
Primarily just sorts includes, adds `ifdefs`, etc.
fixes #10 

Also partially addresses #6. For _most_ uses, it is okay to build with `/experimental:preprocessor`; the fix here is mostly to replace `, __VA_ARGS__` with `, ##__VA_ARGS__` as that's the more portable way to elide providing an empty argument when `__VA_ARGS__` is missing. This works so long as `WI_ARGS_COUNT` is not used when given an empty `__VA_ARGS__`. I.e. using any `*_MSG` or `*_WITH_EXPECTED` may fail. This is due to the fact that it's impossible to have a missing `__VA_ARGS__` when there are no commas in the call (it gets interpreted as an empty argument). When `__VA_OPT__` support is available (i.e. >= C++20), this is no longer true since `__VA_OPT__` properly handles empty `__VA_ARGS__`. Once C++20 is the de-facto standard we can enable `/experimental:preprocessor` by default.


tests/CppWinRT20Tests.cpp
```diff
@@ -1,13 +1,13 @@
...
-#include "common.h"
-
 // Prior to C++/WinRT 2.0 this would cause issues since we're not including wil/cppwinrt.h in this translation unit.
 // However, since we're going to link into the same executable as 'CppWinRTTests.cpp', the 'winrt_to_hresult_handler'
 // global function pointer should be set, so these should all run successfully
 
 #include <winrt/base.h>
 #include <wil/result.h>
 
+#include "common.h"
```
dunhor: FYI: These movements are to ensure that all WIL headers are included first

PR #57, opened 2019-06-13
Add RPC context handle wrapper

jonwis:
Here's an rpc-specific expansion of `wil::unique_any` that wraps a context handle and ensures the call to any `Close` method (which is itself an RPC call) is properly guarded against infrastructure failures.

### DIFF
--- a/include/wil/rpc_helpers.h
+++ b/include/wil/rpc_helpers.h
@@ -158,7 +158,7 @@ namespace wil
     // HRESULT OpenFoo([in] handle_t binding, [out] FOO_CONTEXT*);
     // HRESULT UseFoo([in] FOO_CONTEXT context;
     // void CloseFoo([in, out] PFOO_CONTEXT);
-    using unique_foo_context = wil::unique_context_handle<FOO_CONTEXT, decltype(&CloseFoo), CloseFoo>;
+    using unique_foo_context = wil::unique_rpc_context_handle<FOO_CONTEXT, decltype(&CloseFoo), CloseFoo>;
     unique_foo_context context;
     RETURN_IF_FAILED(wil::invoke_rpc_nothrow(OpenFoo, m_binding.get(), context.put()));
     RETURN_IF_FAILED(wil::invoke_rpc_nothrow(UseFoo, context.get()));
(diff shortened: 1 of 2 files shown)


include/wil/rpc_helpers.h
```diff
@@ -132,6 +131,43 @@ namespace wil
         RpcEndExcept
     }
 
+    namespace details
+    {
+        // Provides an adapter around calling the context-handle-close method on an
+        // RPC interface, which itself is an RPC call.
+        template<typename TStorage, typename close_fn_t, close_fn_t close_fn>
```
oldnewthing: Will the `TStorage` always be a trivially-copyable thing no bigger than a pointer?

include/wil/rpc_helpers.h
```diff
@@ -132,6 +131,43 @@ namespace wil
...
+    // void CloseFoo([in, out] PFOO_CONTEXT);
+    using unique_foo_context = wil::unique_context_handle<FOO_CONTEXT, decltype(&CloseFoo), CloseFoo>;
+    unique_foo_context context;
+    RETURN_IF_FAILED(wil::invoke_rpc_nothrow(OpenFoo, m_binding.get(), context.put()));
+    RETURN_IF_FAILED(wil::invoke_rpc_nothrow(UseFoo, context.get()));
+    context.reset();
+    ~~~
+    */
+    template<typename TContext, typename close_fn_t, close_fn_t close_fn>
+    using unique_context_handle = unique_any<TContext, decltype(&details::rpc_closer_t<TContext, close_fn_t, close_fn>::Close), details::rpc_closer_t<TContext, close_fn_t, close_fn>::Close>;
```
oldnewthing: Perhaps `unique_rpc_context_handle` would be clearer. While the term "context handle" is used primarily for RPC, it is not exclusively used in that sense.

include/wil/rpc_helpers.h
jonwis: Pretty sure, RPC handles must be pointer-sized.

include/wil/rpc_helpers.h
jonwis: Good idea, will fix

PR #64, opened 2019-06-29
Various open issues - resource managers, permissive, and unique_variant

jonwis:


include/wil/resource.h
```diff
@@ -4812,10 +4812,10 @@ namespace wil
 #define __WIL_OBJBASE_H_
     using cotaskmem_deleter = function_deleter<decltype(&::CoTaskMemFree), ::CoTaskMemFree>;
 
-    template <typename T>
+    template <typename T = void>
     using unique_cotaskmem_ptr = wistd::unique_ptr<T, cotaskmem_deleter>;
 
-    template <typename T>
+    template <typename T = void>
```
dunhor: I'm not sure this is correct. `unique_array_ptr` seems to indicate that it's an array of `T`. Array of `void` doesn't seem like it would work out.

dunhor: I'm not sure this is correct. `unique_array_ptr` seems to indicate that it's an array of `T`. Array of `void` doesn't seem like it would work out.

Issue #27, opened 2019-05-11
Multiple files not /analyze clean


zx64:
Tested against 4361351 with local fix for #26.

Test file (in root of repo for simplicity)
```
#define WIN32_LEAN_AND_MEAN
#include <windows.h>
#include "include/wil/com.h"
#include "include/wil/registry.h"
#include "include/wil/filesystem.h"
#include "include/wil/win32_helpers.h"
int main() {}
```

```
> cl testwil.cpp /W4 /WX /permissive- /analyze
Microsoft (R) C/C++ Optimizing Compiler Version 19.20.27508.1 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

[see link below]
```
Compiler output: https://gist.github.com/zx64/563b366873fd8063bb72a44ec5839e52

ChrisGuzak:
yes, thanks for pointing this out. this will generate a lot of noise (already seen on our internal discussion alias) so we should prioritize this.

jonwis:
Started a stab at this with https://github.com/microsoft/wil/compare/jonwis/analysis-cleanup

jonwis:
Holy cats I have no idea what's going on in the "you changed the entire file!" in that branch. Will figure it out.

sylveon:
That's usually line endings. A .editorconfig should fix editors silently changing those.

jonwis:
Ok, down to a mysterious `*pNode` (maybe analysis failure) and a problem in `GetFileInfo`. Progress!

rysavyjan:
Run Code Analysis with VS 2019 (version 16.2.3) returns

```
1>------ Rebuild All started: Project: ConsoleApplication1, Configuration: Debug Win32 ------
1>ConsoleApplication1.cpp
1>C:\Users\Jan\source\repos\ConsoleApplication1\wil\result_macros.h(3616,1): error C2059:  syntax error: '<L_GRAMMAR_template_head_l>'
1>C:\Users\Jan\source\repos\ConsoleApplication1\wil\result_macros.h(3618,1): error C2143:  syntax error: missing ';' before '{'
1>C:\Users\Jan\source\repos\ConsoleApplication1\wil\result_macros.h(3618,1): error C2447:  '{': missing function header (old-style formal list?)
1>C:\Users\Jan\source\repos\ConsoleApplication1\wil\result_macros.h(3626,1): error C2059:  syntax error: ''template<''
[...]

PR #66, opened 2019-07-01
Clean up /analysis errors in wil headers, test code, and related

### LINKED ISSUE #65: Repo is missing a .gitattributes
Turns out the changes I've pushed (like #51) have had Unix-style LF endings instead of Windows-style CRLF endings.

I presume we need to add a `.gitattributes` containing:

```
# Disable CRLF-mapping for all files in the depot.
* -text
```

And then use [the sequence here](https://stackoverflow.com/questions/15641259/how-to-normalize-working-tree-line-endings-in-git) to renormalize broken endings.

jonwis:
Using VS2019's "Analzyze > Run Code Analysis > Run Code Analysis on {project}" turned up a lot of issues within the core WIL code.  This PR addresses as many as possible, including:

* Missing initialization of members and locals
* Incorrect or insufficient annotations on WIL infrastructure and in macros
* Missing `WI_NOEXCEPT` markers
* Missing null checks or flows that break analysis annotation assumptions
* Explicitly discard some returned temporaries
* Normalize endings in newly-added files
* Many many many code analysis annotations

Specifically, this fixes #65 and #27.


tests/common.h
```diff
@@ -251,7 +251,7 @@ namespace witest
         bool doesThrow = false;
         bool doesCrash = DoesCodeCrash([&]()
         {
-            doesThrow = DoesCodeThrow(callOp);
+            doesThrow = DoesCodeThrow(wistd::forward<Lambda>(callOp));
```
dunhor: Was this actually a warning? Forwarding lambdas that aren't copied/moved - while not necessarily a bad thing - is technically unnecessary.

tests/common.h
```diff
@@ -355,7 +355,7 @@ namespace witest
         TestFolder(const TestFolder&) = delete;
         TestFolder& operator=(const TestFolder&) = delete;
 
-        TestFolder(TestFolder&& other)
+        TestFolder(TestFolder&& other) WI_NOEXCEPT
```
dunhor: Did something really warn about a function _not_ being noexcept? I'm not arguing that this _shouldn't_ be noexcept, but I'm picturing a warning that says "function that does not throw should be declared as noexcept" which would be a terrible, terrible thing to have

tests/workarounds/wrl/wrl/implements.h
```diff
@@ -2153,7 +2153,10 @@ class RuntimeClass :
     RuntimeClass(const RuntimeClass&);
     RuntimeClass& operator=(const RuntimeClass&);
 protected:
-    HRESULT CustomQueryInterface(REFIID /*riid*/, _Outptr_result_nullonfailure_ void** /*ppvObject*/, _Out_ bool *handled)
+    HRESULT CustomQueryInterface(
```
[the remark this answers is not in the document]
dunhor: IMO don't bother with the tests at all. If you _do_ fix something here, you should apply the same fix to the same file in the OS repo since these files are intended to be temporary and deleted eventually.

include/wil/wistd_functional.h
```diff
@@ -262,7 +262,7 @@ namespace wistd     // ("Windows Implementation" std)
         static constexpr size_t __buffer_size = 13 * sizeof(void*);
 
         typedef __function::__base<_Rp(_ArgTypes...)> __base;
-        typename aligned_storage<__buffer_size>::type __buf_;
+        typename aligned_storage<__buffer_size>::type __buf_{};
```
dunhor: IMO this warning should just get suppressed. This has a runtime cost that's entierly unnecessary

include/wil/wistd_functional.h
```diff
@@ -303,12 +303,12 @@ namespace wistd     // ("Windows Implementation" std)
         __WI_LIBCPP_INLINE_VISIBILITY
         function(nullptr_t) WI_NOEXCEPT : __f_(0) {}
         function(const function&);
-        function(function&&);
+        function(function&&) WI_NOEXCEPT;
```
dunhor: Please remove. This should not be noexcept

include/wil/wistd_functional.h
```diff
@@ -303,12 +303,12 @@ namespace wistd     // ("Windows Implementation" std)
...
         function(nullptr_t) WI_NOEXCEPT : __f_(0) {}
         function(const function&);
-        function(function&&);
+        function(function&&) WI_NOEXCEPT;
         template<class _Fp, class = _EnableIfCallable<_Fp>>
         function(_Fp);
 
         function& operator=(const function&);
-        function& operator=(function&&);
+        function& operator=(function&&) WI_NOEXCEPT;
```
dunhor: Please remove. This should not be noexcept

include/wil/winrt.h
```diff
@@ -439,9 +430,9 @@ namespace wil
             struct type // T holder
             {
                 type() {};
-                type(T&& value) : m_value(wistd::forward<T>(value)) {};
+                type(T&& value) WI_NOEXCEPT : m_value(wistd::forward<T>(value)) {};
```
dunhor: Anything storing a `T` should not have its copy/move constructors/assignment operators explicitly marked as `noexcept`. It should, at best, conditionally declare them as `noexcept`

tests/common.h
jonwis: I'll try removing. There was a complaint about "unused unnamed temporary objects" that prompted this, but it's possible I was holding the report wrong.

tests/common.h
jonwis: The warning was like "this kind of method may not throw" ... not "this method should be noexcept."  https://en.cppreference.com/w/cpp/language/move_constructor has:

> To make strong exception guarantee possible, user-defined move constructors should not throw exceptions. For example, std::vector relies on std::move_if_noexcept to choose between move and copy when the elements need to be relocated.

But I don't know how to compile the language spec in my head so I don't know if there's a hard requirement.

I'm happy to roll back changes in test code as the goal is to get the library itself analyze-clean.

dunhor: If it's only warning about move constructors/assignment operators, that's not as bad since it _is_ a desirable property to have such operations be `noexcept`. That said, if it gave the same warning for `wistd::function`, that's a pretty embarrassing bug in the analyzer. Either that or it's warning for _all_ move constructors, which would be worse and effectively renders the warning useless and something that should always be disabled.

tests/common.h
dunhor: Could it possibly be related to the `(void)` you added to `DoesCodeThrow`? Before the change, this function will capture `callOp` as either an l-value or r-value reference and pass as an l-value reference to `DoesCodeThrow`. There shouldn't be any "unused unnamed temporary objects" created

PR #68, opened 2019-07-04
Fix: Increase version of process local storage identifier

matbech:
The internal layout of the Node struct has been changed in this commit:
https://github.com/microsoft/wil/commit/133201137cb5405cd62e730abca3dc8bd7efb19d#diff-e1b0b9bc94e49529ae997e1977a311f8L431

However, the version of the process local storage has not been increased. This PR increases the version from 2 to 3.


ChrisGuzak:
good fix. this bug is pernicious. when we failed to do this before many people wasted time diagnosing this problem. lets get it in quick.

jonwis:
Thanks for catching this. We should probably add a compile-time set of layout assertions using `offsetof` + `sizeof` and move the primitive types up & out of the structs so they can be tested against. The static-assert message (or nearby source) can contain specific instructions on how and when to update the ABI portions.

PR #72, opened 2019-07-08
Remove use of deprecated macros

oldnewthing:


include/wil/registry.h
```diff
@@ -49,7 +49,7 @@ namespace wil
     //     {
     //         // invalidate cached registry data here
     //     });
-    // RETURN_HR_IF_FALSE(E_OUTOFMEMORY, watcher);
+    // RETURN_HR_IF(E_OUTOFMEMORY, !watcher);
```
dunhor: `RETURN_IF_NULL_ALLOC` perhaps seems better for this one

oldnewthing: Doesn't work because `RETURN_IF_NULL_ALLOC` requires a pointer.

dunhor: Seems to work for me. I think it just requires something that's convertible to `bool`

oldnewthing: It wants something [convertible to a pointer](https://github.com/microsoft/wil/blob/6112ec16d8a09402ef980c5b951e17bda2e5898c/include/wil/result_macros.h#L607), but maybe there's a secret converter inside the watcher.

dunhor: Ahh, okay. `unique_registry_watcher_nothrow` is a `unique_any_t`, which [is comparable with `nullptr_t`](https://github.com/microsoft/wil/blob/6112ec16d8a09402ef980c5b951e17bda2e5898c/include/wil/resource.h#L340)

oldnewthing: Mystery solved. (I was worried that we constructed a temporary `unique_event_watcher_nothrow(nullptr)`, did the comparison, and then destructed the temporary.) Should I use `RETURN_IF_NULL_ALLOC`?

PR #75, opened 2019-07-21
Statically initialized SIDs

jonwis:
Provide a helper that constructs a valid _SID of a predefined size without costing an allocation.

### DIFF
--- a/include/wil/token_helpers.h
+++ b/include/wil/token_helpers.h
@@ -438,28 +438,61 @@ namespace wil
     }
 #endif // WIL_ENABLE_EXCEPTIONS
 
-    template<typename... Ts> constexpr auto static_sid(const SID_IDENTIFIER_AUTHORITY& authority, Ts&& ... subAuthorities)
+    namespace details
     {
-        struct FakeSid
+        template<size_t AuthorityCount> struct static_sid_t
         {
             BYTE Revision;
             BYTE SubAuthorityCount;
             SID_IDENTIFIER_AUTHORITY IdentifierAuthority;
-            DWORD SubAuthority[sizeof...(subAuthorities)];
+            DWORD SubAuthority[AuthorityCount];
 
             PSID get()
             {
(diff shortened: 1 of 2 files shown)


include/wil/token_helpers.h
```diff
@@ -438,6 +438,63 @@ namespace wil
...
+    namespace details
+    {
+        template<size_t AuthorityCount> struct static_sid_t
+        {
+            BYTE Revision;
+            BYTE SubAuthorityCount;
+            SID_IDENTIFIER_AUTHORITY IdentifierAuthority;
+            DWORD SubAuthority[AuthorityCount];
+
+            PSID get()
```
ChrisGuzak: given `PSID` is an alias for `void*` it is dangerous. nothing we can do about it here that I'm aware of, but FYI.

include/wil/token_helpers.h
```diff
@@ -438,6 +438,63 @@ namespace wil
...
+    Replaces AllocateAndInitializeSid by constructing a structure laid out like a PSID, but
+    returned like a value. The resulting object is suitable for use with any method taking PSID,
+    passed by "&the_sid" or via "the_sid.get()"
+    ~~~~
+    // Change the owner of the key to administrators
+    auto systemSid = wil::make_static_sid(SECURITY_NT_AUTHORITY, SECURITY_BUILTIN_DOMAIN_RID, DOMAIN_ALIAS_RID_ADMINS);
+    RETURN_IF_WIN32_ERROR(SetNamedSecurityInfo(keyPath, SE_REGISTRY_KEY, OWNER_SECURITY_INFORMATION, &systemSid, nullptr, nullptr, nullptr));
+    ~~~~
+    */
+    template<typename... Ts> constexpr auto make_static_sid(const SID_IDENTIFIER_AUTHORITY& authority, Ts&&... subAuthorities)
```
dunhor: Forwarding reference seems kind of overkill here if the arguments are expected to all be `DWORD`

PR #80, opened 2019-08-07
Fix noreturn compile error for clang

giuntnic:
Many of the functions called from the RETURN_IF_FAILED macro used noreturn annotation since in the failfast and exception case it isn't expected for them to return. In the case that a failfast or exception wasn't expected the functions would still retrun and had SAL annotation to express this. Clang doesn't pick up on the SAL resulting in a warning being generated. This change provides template overloads based on the return type so that in all cases the functions are properly annotated as noreturn or not. Additionally, the suppress action flag was factored into a template parameter for ReportFailure as that overrides no return behavior and setting it at runtime for a failfast or exception return would result in UB. Tested on VS 2017 and Clang 9.0.0

### DIFF
--- a/include/wil/result_macros.h
+++ b/include/wil/result_macros.h
@@ -2951,6 +2951,9 @@ namespace wil
 
             case SupportedExceptions::None:
                 return functor.Run();
+
+            case SupportedExceptions::Default:
+                WI_ASSERT(false);
             }
 
             WI_ASSERT(false);
@@ -3393,7 +3396,7 @@ namespace wil
 
             LogFailure(__R_FN_CALL_FULL, T, hr, message, needPlatformException,
                 debugString, ARRAYSIZE(debugString), callContextString, ARRAYSIZE(callContextString), &failure);
-
+__WI_SUPPRESS_4127_S
             if (T == FailureType::FailFast)
             {
                 WilFailFast(const_cast<FailureInfo&>(failure));
@@ -3415,6 +3418,7 @@ namespace wil
                 // Wil was instructed to throw, but doesn't have any capability to do so (global function pointers are not setup)
                 WilFailFast(const_cast<FailureInfo&>(failure));
             }
+__WI_SUPPRESS_4127_E
         }
 
         template<>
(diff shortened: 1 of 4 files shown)


include/wil/result_macros.h
```diff
@@ -1340,10 +1339,14 @@ namespace wil
...
                                _Out_ FailureInfo *failure) WI_NOEXCEPT;
-        _When_((type == FailureType::Exception) || (type == FailureType::FailFast), _Analysis_noreturn_) __declspec(noinline) inline void ReportFailure(__R_FN_PARAMS_FULL, FailureType type, HRESULT hr, _In_opt_ PCWSTR message = nullptr, ReportFailureOptions options = ReportFailureOptions::None);
-        _When_((type == FailureType::Exception) || (type == FailureType::FailFast), _Analysis_noreturn_) inline void ReportFailure_ReplaceMsg(__R_FN_PARAMS_FULL, FailureType type, HRESULT hr, _Printf_format_string_ PCSTR formatString, ...);
-        _When_((type == FailureType::Exception) || (type == FailureType::FailFast), _Analysis_noreturn_) __declspec(noinline) inline void ReportFailure_Hr(__R_FN_PARAMS_FULL, FailureType type, HRESULT hr);
-        _When_((type == FailureType::Exception) || (type == FailureType::FailFast), _Analysis_noreturn_) __declspec(noinline) inline HRESULT ReportFailure_CaughtException(__R_FN_PARAMS_FULL, FailureType type, SupportedExceptions supported = SupportedExceptions::Default);
+        template<FailureType, bool = false>
```
dunhor: >FailureType [](start = 17, length = 11)

The `FailureType` template parameter needs a name for the SAL below (assuming SAL can even handle template arguments)

giuntnic: Is there a way to validate SAL? Even without 'type' being defined this compiled without issue. I can't find any resources that mention SAL and template parameters. I'd imagine they probably wouldn't play nice together so lacking a way to validate the SAL I'll just remove it.

dunhor: The compiler doesn't really care about SAL (one of the many reasons I prefer not to use it). @jonwis may know more since he recently did some work to fix some of them up. I believe you ran run code analysis in VS or something like that

ChrisGuzak: I suspect SAL does not do template params.
can we move these to the specializations of the functions and not make them conditional?

include/wil/result_macros.h
```diff
@@ -3363,36 +3367,72 @@ namespace wil
             wchar_t debugString[2048];
             char callContextString[1024];
 
-            LogFailure(__R_FN_CALL_FULL, type, hr, message, needPlatformException,
+            LogFailure(__R_FN_CALL_FULL, T, hr, message, needPlatformException,
                 debugString, ARRAYSIZE(debugString), callContextString, ARRAYSIZE(callContextString), &failure);
+        }
+
+        _Use_decl_annotations_
```
ChrisGuzak: I suspect this won't work and hope it is not needed anyway

ChrisGuzak:
> This change requires a successful OS build before we can port it to GitHub

Duncan, can you work with Nicholas to make sure this happens and then approve?

dunhor:
Since many function signatures were kept intact, getting an OS build isn't as crucial any more. I think that the OS tests will still break since they'll be detouring the wrong function, but that's not as critical. Other than that, we'll probably hit breaks from publish size limits.

At the moment I'm trying to figure out why triggers for the CI build randomly broke (or figure out how to manually queue a build for a PR in the meantime)

dunhor:
Figured out how to manually queue. Build is still failing: https://dev.azure.com/msft-wil/Windows%20Implementation%20Library/_build/results?buildId=124

sylveon:
Seems to be caused by the compiler triggering a warning for `if (T == FailureType::FailFast)`, saying `if constexpr` can be used, and warnings being treated as error.

Since `if constexpr` is a C++17 feature, this warning can either be muted, or a `WIL_IF_CONSTEXPR` macro can be introduced to use regular if in case C++17 is not available.

giuntnic:
I was considering going the WIL_IF_CONSTEXPR route but the warning still fires on pre c++17 code which means there would need to be another workaround for that code (see this [example](https://godbolt.org/z/E5t_Rb)). Given that I'll probably just add the pragma to mute the error.

PR #83, opened 2019-08-12
Update root CMakeLists to include 'project()' to suppress CMake warning and remove '-Wno-invalid-noreturn' clang compiler option

dunhor:
Recent versions of CMake are now issuing a warning when the top-level CMakeLists.txt has no `project()` command, so adding one to suppress the warning.

Also using this PR as a means to try and diagnose what's wrong with the CI build triggers, so if you see random commits with whitespace changes, that's why

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
 cmake_minimum_required(VERSION 3.11)
+project(WIL)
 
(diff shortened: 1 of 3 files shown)


dunhor:
And of course the CI trigger works when _I_ try and test it...

PR #84, opened 2019-08-16
Add constructor unique_hglobal_locked(HGLOBAL)

petrsol:
I have added constructor which was missing to me. Without it I had to use STGMEDIUM structure and its member hGlobal to use unique_hglobal_locked.

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -5247,6 +5247,12 @@ namespace wil
             m_globalMemory = GlobalLock(medium.hGlobal);
         }
 
+        unique_hglobal_locked(HGLOBAL global) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(global)
+        {
+            // GlobalLock returns a pointer to the associated global memory block and that's what callers care about.
+            m_globalMemory = GlobalLock(global);
+        }
+
         // In the future, we could easily add additional constructor overloads such as unique_hglobal_locked(HGLOBAL) to make consumption easier.
 
         pointer get() const
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -5245,6 +5245,10 @@ namespace wil
         {
             // GlobalLock returns a pointer to the associated global memory block and that's what callers care about.
             m_globalMemory = GlobalLock(global);
+            if (!m_globalMemory)
+            {
+                release();
+            }
         }
 
         explicit unique_hglobal_locked(unique_hglobal& global) : unique_hglobal_locked(global.get())
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -5255,7 +5255,7 @@ namespace wil
         {
         }
 
-        explicit unique_hglobal_locked(_In_ STGMEDIUM& medium) : unique_hglobal_locked(medium.hGlobal)
+        explicit unique_hglobal_locked(STGMEDIUM& medium) : unique_hglobal_locked(medium.hGlobal)
         {
         }
 
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -5241,17 +5241,17 @@ namespace wil
     {
         unique_hglobal_locked() = delete;
 
-        unique_hglobal_locked(HGLOBAL global) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(global)
+        explicit unique_hglobal_locked(HGLOBAL global) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(global)
         {
             // GlobalLock returns a pointer to the associated global memory block and that's what callers care about.
             m_globalMemory = GlobalLock(global);
         }
 
-        unique_hglobal_locked(unique_hglobal &global) : unique_hglobal_locked(global.get())
+        explicit unique_hglobal_locked(unique_hglobal& global) : unique_hglobal_locked(global.get())
         {
         }
 
-        unique_hglobal_locked(_In_ STGMEDIUM &medium) : unique_hglobal_locked(medium.hGlobal)
+        explicit unique_hglobal_locked(_In_ STGMEDIUM& medium) : unique_hglobal_locked(medium.hGlobal)
         {
         }
 
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -5247,7 +5247,7 @@ namespace wil
             m_globalMemory = GlobalLock(global);
         }
 
-        unique_hglobal_locked(const unique_hglobal &global) : unique_hglobal_locked(global.get())
+        unique_hglobal_locked(unique_hglobal &global) : unique_hglobal_locked(global.get())
         {
         }
 
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -5241,19 +5241,19 @@ namespace wil
     {
         unique_hglobal_locked() = delete;
 
-        unique_hglobal_locked(_In_ STGMEDIUM& medium) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(medium.hGlobal)
-        {
-            // GlobalLock returns a pointer to the associated global memory block and that's what callers care about.
-            m_globalMemory = GlobalLock(medium.hGlobal);
-        }
-
         unique_hglobal_locked(HGLOBAL global) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(global)
         {
             // GlobalLock returns a pointer to the associated global memory block and that's what callers care about.
             m_globalMemory = GlobalLock(global);
         }
 
-        // In the future, we could easily add additional constructor overloads such as unique_hglobal_locked(HGLOBAL) to make consumption easier.
+        unique_hglobal_locked(const unique_hglobal &global) : unique_hglobal_locked(global.get())
+        {
+        }
+
+        unique_hglobal_locked(_In_ STGMEDIUM &medium) : unique_hglobal_locked(medium.hGlobal)
+        {
+        }
 
         pointer get() const
         {


include/wil/resource.h
```diff
@@ -5247,6 +5247,12 @@ namespace wil
             m_globalMemory = GlobalLock(medium.hGlobal);
         }
 
+        unique_hglobal_locked(HGLOBAL global) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(global)
+        {
+            // GlobalLock returns a pointer to the associated global memory block and that's what callers care about.
+            m_globalMemory = GlobalLock(global);
+        }
+
         // In the future, we could easily add additional constructor overloads such as unique_hglobal_locked(HGLOBAL) to make consumption easier.
```
sylveon: Remove this comment since you just added that constructor overload

include/wil/resource.h
```diff
@@ -5247,6 +5247,12 @@ namespace wil
             m_globalMemory = GlobalLock(medium.hGlobal);
```
sylveon: Make this constructor delegate to the HGLOBAL constructor to avoid code duplication:
```cpp
unique_hglobal_locked(_In_ STGMEDIUM& medium) : unique_hglobal_locked(medium.hGlobal)
{
}
```

include/wil/resource.h
petrsol: Yes, thanks, comment is away in my second commit.

include/wil/resource.h
petrsol: Good idea, thanks, done.

petrsol: I have added also constructor taking unique_hglobal& as you suggested.

I do not feel experienced enough to change get() method. It would break existing code if changed to default behaviour.

I have experimented with template<typename T = void>. But I have received only bunch of compiler errors so it is probably over my knowledge, sorry.

sylveon: Don't worry about modifying `get()` or adding templates, I can fix those by myself in another pull request.

include/wil/resource.h
```diff
@@ -5241,13 +5241,19 @@ namespace wil
...
-        unique_hglobal_locked(_In_ STGMEDIUM& medium) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(medium.hGlobal)
+        unique_hglobal_locked(HGLOBAL global) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(global)
         {
             // GlobalLock returns a pointer to the associated global memory block and that's what callers care about.
-            m_globalMemory = GlobalLock(medium.hGlobal);
+            m_globalMemory = GlobalLock(global);
         }
 
-        // In the future, we could easily add additional constructor overloads such as unique_hglobal_locked(HGLOBAL) to make consumption easier.
+        unique_hglobal_locked(const unique_hglobal &global) : unique_hglobal_locked(global.get())
```
sylveon: I would not take a const-reference, because it allows this code
```cpp
wil::unique_hglobal_locked lock{ GetHglobal() };
// HGLOBAL freed by unique_any's destructor, lock.get() will give you invalid memory
```

Instead, it should take a non-const reference, which will force you to keep the HGLOBAL alive:
```cpp
wil::unique_hglobal_locked lock{ GetHglobal() }; // This is a compiler error

auto hglobal{ GetHglobal() };
wil::unique_hglobal_locked lock{ hglobal };
// HGLOBAL not freed until hglobal goes out of scope. lock.get() gives you valid memory
```

petrsol: Yes, that's right, r-values are not welcomed here. ;-)

include/wil/resource.h
```diff
@@ -5241,13 +5241,19 @@ namespace wil
...
-        unique_hglobal_locked(_In_ STGMEDIUM& medium) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(medium.hGlobal)
+        unique_hglobal_locked(HGLOBAL global) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(global)
         {
             // GlobalLock returns a pointer to the associated global memory block and that's what callers care about.
-            m_globalMemory = GlobalLock(medium.hGlobal);
+            m_globalMemory = GlobalLock(global);
         }
 
-        // In the future, we could easily add additional constructor overloads such as unique_hglobal_locked(HGLOBAL) to make consumption easier.
+        unique_hglobal_locked(unique_hglobal &global) : unique_hglobal_locked(global.get())
```
jonwis: Maybe this should be `unique_hglobal const&` instead, as it's not mutating the inbound `unique_hglobal` ?

include/wil/resource.h
```diff
@@ -5241,13 +5241,19 @@ namespace wil
     {
         unique_hglobal_locked() = delete;
 
-        unique_hglobal_locked(_In_ STGMEDIUM& medium) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(medium.hGlobal)
+        unique_hglobal_locked(HGLOBAL global) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(global)
```
jonwis: Should this and the other ctors be marked `explicit` ?

include/wil/resource.h
jonwis: Not sure this is correct. [GlobalLock](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-globallock) says it increments the lock-count, and the object is destroyed when the lock count drops to zero. I think this also means `unique_hglobal_locked` needs to _not_ call `GlobalUnlock` if the `GlobalLock` fails.

So in your example:

```c++
HGLOBAL GetHGlobal();
wil::unique_hglobal_locked lock { GetHGlobal() }; // Success, explicit use of `HGLOBAL` ctor
// what GetHGlobal returned now has lock count +1

auto hglobal { GetHGlobal() }; // Type is HGLOBAL but without a strong reference on it (so "+0")
wil::unique_hglobal_locked lock2 { hglobal }; // Success, explicit use of `HGLOBAL` ctor
// And lock2 is holding is now holding a +1 lock count.  When it goes out of scope hglobal
// still contains what GetHGlobal would have returned; if it was previously locked that
// lock count remains.

File: include/wil/resource.h [hunk 6]
```diff
@@ -5241,13 +5241,19 @@ namespace wil
     {
         unique_hglobal_locked() = delete;
 
-        unique_hglobal_locked(_In_ STGMEDIUM& medium) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(medium.hGlobal)
+        unique_hglobal_locked(HGLOBAL global) : unique_any<void*, decltype(&::GlobalUnlock), ::GlobalUnlock>(global)
         {
             // GlobalLock returns a pointer to the associated global memory block and that's what callers care about.
-            m_globalMemory = GlobalLock(medium.hGlobal);
+            m_globalMemory = GlobalLock(global);
```
jonwis: If this GlobalLock fails, the global should not be unlocked.

include/wil/resource.h
```diff
@@ -5241,13 +5241,19 @@ namespace wil
...
-            m_globalMemory = GlobalLock(medium.hGlobal);
+            m_globalMemory = GlobalLock(global);
         }
 
-        // In the future, we could easily add additional constructor overloads such as unique_hglobal_locked(HGLOBAL) to make consumption easier.
+        unique_hglobal_locked(unique_hglobal &global) : unique_hglobal_locked(global.get())
+        {
+        }
+
+        unique_hglobal_locked(_In_ STGMEDIUM &medium) : unique_hglobal_locked(medium.hGlobal)
```
ChrisGuzak: while you are here
- remove the `_In_` SAL 
- align the `&` with the type .

This probably should be `const STGMEDIUM& medium`, but that might break some use.

include/wil/resource.h
ChrisGuzak: note, only movable memory has a lock count. new code should use movable as that was a win16 construct not useful on current platforms. Of course there is still code that uses the movable, notably OLE, so we need to support it. but code creating global handles should use `GMEM_FIXED` (`GPTR`)

include/wil/resource.h
ChrisGuzak: note, failure only occurs on invalid handle. the alloc step ensures that the resources are available.

include/wil/resource.h
sylveon: Clipboard related functions are still documented to require movable memory.

include/wil/resource.h
sylveon: That was a change I suggested, because it would allow to take a rvalue `unique_hglobal`, making the locked to memory invalid after the call to the constructor. A proper move constructor could be added that will also free the hglobal in the destructor, but that would complexify the class.

https://github.com/microsoft/wil/pull/84#discussion_r314894385

include/wil/resource.h
petrsol: I have tried it and when GlobalFree is called for locked global memory handle, content of memory is changed (in my case first 8 bytes) and part of memory become inaccessible too (debugger show ?? marks instead of content). Also another call to GlobalLock with original global memory handle has returned NULL. So I think it is not possible to use locked memory after calling GlobalFree because it is deallocated.

```
HGLOBAL mem;
GetHGlobalFromStream(memStream.get(), &mem);
void *m = GlobalLock(mem);  // here I see memory 'm' content OK
GlobalFree(mem);            // after this call memory 'm' is corrupted and partly inaccessible
void *m2 = GlobalLock(mem); // this returns NULL
```

include/wil/resource.h
petrsol: OK, I have added explicit to all three constructors.

include/wil/resource.h
petrsol: OK, I have added call to release() in case of GlobalLock returns NULL.

include/wil/resource.h
sylveon: `The GlobalFree function will free a locked memory object.`: https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-globalfree

include/wil/resource.h
petrsol: I have removed _In_ SAL and aligned the & with the type.

include/wil/resource.h
dunhor: No, don't add `const`

petrsol:
I'm confused, please help me. Is any action expected from me? Should I change anything according to @jonwis change request? Or should I just wait patiently?

sylveon:
Just wait, it's labor day today so they probably have a day off

PR #87, opened 2019-08-20
Simplify tests to build faster

### LINKED ISSUE #9: Tests take an excessive amount of time to build
The tests were copied mostly verbatim from the OS depot. These were already pretty bad in terms of build time, but undoubtedly made worse by the change over to Catch2. Most of the excessive build time comes from `ComTests.cpp`, likely through the excessive amount of template instantiation that it does. Some if not much if not most of that is likely redundant, and if we can reduce it down we certainly should. `WinRTTests.cpp` and `wiTest.cpp` are also possible offenders, but not nearly to the degree that the COM tests is. For now I've disabled Release builds on the CI machines to try and improve the state of things (15-ish min compile time for Debug compared to 50-ish min compile time for Release).

dunhor:
Predominantly modifies tests to move more stuff under the `WIL_EXHAUSTIVE_TEST` condition. Most of the things moved are mostly redundant (e.g. no need to test calling `QueryInterface` more than once). Other changes were to avoid doing things that were unnecessary (e.g. no need to test `com_ptr_t::operator&` or `com_raw_ptr` for all specializations). There's still room for improvement, but this is a good start.

Also changes Azure pipelines file to include relwithdebinfo now that we should be able to build everything within the allotted time limit.

Also adds the option to the init script to use `lld-link` in place of `link.exe`, though speed impact isn't that huge since template instantiation/Catch2 accounts for the vast majority of build time.

fixes #9

### DIFF
--- a/tests/ComTests.cpp
+++ b/tests/ComTests.cpp
@@ -944,25 +944,23 @@ static void TestSmartPointerConversion(const Ptr1& ptr1, const Ptr2& ptr2)
 template <typename IFace1, typename IFace2>
 static void TestPointerConversionCombination(IFace1* p1, IFace2* p2)
 {
+#ifdef WIL_ENABLE_EXCEPTIONS
+    TestSmartPointerConversion(wil::com_ptr<IFace1>(p1), wil::com_ptr_nothrow<IFace2>(p2));
+#endif
+    TestSmartPointerConversion(wil::com_ptr_failfast<IFace1>(p1), wil::com_ptr_nothrow<IFace2>(p2));
+    TestSmartPointerConversion(wil::com_ptr_nothrow<IFace1>(p1), wil::com_ptr_nothrow<IFace2>(p2));
+
 #ifdef WIL_EXHAUSTIVE_TEST
 #ifdef WIL_ENABLE_EXCEPTIONS
     TestSmartPointerConversion(wil::com_ptr<IFace1>(p1), wil::com_ptr<IFace2>(p2));
-    TestSmartPointerConversion(wil::com_ptr<IFace1>(p1), wil::com_ptr_failfast<IFace2>(p2));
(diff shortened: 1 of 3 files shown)


tests/ComTests.cpp
```diff
@@ -954,6 +956,7 @@ static void TestPointerConversionCombination(IFace1* p1, IFace2* p2)
 #endif
     TestSmartPointerConversion(wil::com_ptr_failfast<IFace1>(p1), wil::com_ptr_failfast<IFace2>(p2));
     TestSmartPointerConversion(wil::com_ptr_failfast<IFace1>(p1), wil::com_ptr_nothrow<IFace2>(p2));
+#endif // WIL_EXHAUSTIVE_TEST
 
 #ifdef WIL_ENABLE_EXCEPTIONS
     TestSmartPointerConversion(wil::com_ptr_nothrow<IFace1>(p1), wil::com_ptr<IFace2>(p2));
```
dunhor: >com_ptr_nothrow [](start = 36, length = 15)

In hindsight, it's probably best to switch up the type of `Ptr1`; I'll make the change

dunhor:
>     TestPointerConversion<IUnknown, IUnknown, ComObject>();

The additional ones with duplicate types are also probably redundant

---
Refers to: tests/ComTests.cpp:994 in 11269e5. [](commit_id = 11269e54b4a7d65659330b786d4cc4caaab887f0, deletion_comment = False)

dunhor:
Well, sadly relwithdebinfo takes ~2x the time to build that debug took before, so pipeline builds now take ~twice what they did before. But the good news is that I figured out how to increase the timeout, so going over an hour doesn't fail the build any more. I'll keep it like this for now; there's typically no significant urgency to submit changes

PR #93, opened 2019-09-19
Add wil::com_ptr::put_void and CLSID overloads to wil::CoCreateInstance

### LINKED ISSUE #92: Best practices for using com_ptr to create factory objects?
Hi,
I'm sorry for asking such a basic question, but I'm back to using COM APIs after a decade-long break and trying to adjust to modern best practices.

I'm working on some code that uses Direct2D and I'm looking at WIL to help me with that.

What are the recommended best practices for using `com_ptr` when calling functions like `D2D1CreateFactory` or `DWriteCreateFactory`? The former uses `void**` as the type of the output parameter, the latter `IUnknown**`. I've been following the examples in the docs for Direct2D and DirectWrite and using `reinterpret_cast` -- that seems to work well after upgrading to `com_ptr`, and looking at the source code there's nothing indicating that there's a better alternative for:
```
wil::com_ptr<IDWriteFactory2> pDWriteFactory2;
DWriteCreateFactory(
    DWRITE_FACTORY_TYPE_SHARED, 
    __uuidof(IDWriteFactory2), 
    reinterpret_cast<IUnknown**>(&pDWriteFactory2));
```
but I might be missing something... I think earlier incarnations of smart COM pointers had methods like `put_void` for getting the underlying pointer cast to `void**` for purposes like this. Just wanted to make sure there's no better way of doing things that I had somehow missed...

My other question is: how to use `wil::CoCreateInstance` to obtain a WIC factory object that implements `IWICImagingFactory2`? The first template argument expects a class, from which it can extract the CLSID using `__uuidof`, but in this case I only have the GUID `CLSID_WICImagingFactory2`. Is it a supported scenario, or should I not even bother and stick to the old way of doing things?

sylveon:
Fixes #92

### DIFF
--- a/include/wil/com.h
+++ b/include/wil/com.h
@@ -396,6 +396,13 @@ namespace wil
             return reinterpret_cast<void**>(put());
         }
 
+        //! Returns the address of the internal pointer casted to IUnknown** (releases ownership of the pointer BEFORE returning the address).
+        //! @see put
+        IUnknown** put_unknown() WI_NOEXCEPT
+        {
+            return reinterpret_cast<IUnknown**>(put());
+        }
+
         //! Returns the address of the internal pointer (releases ownership of the pointer BEFORE returning the address).
         //! The pointer is explicitly released to prevent accidental leaks of the pointer.  Coding standards generally indicate that
         //! there is little valid `_Inout_` use of `IInterface**`, making this safe to do under typical use.  Since this behavior is not always immediately
--- a/include/wil/com.h
+++ b/include/wil/com.h
@@ -376,9 +376,7 @@ namespace wil
 
         //! Returns the address of the internal pointer (releases ownership of the pointer BEFORE returning the address).
         //! The pointer is explicitly released to prevent accidental leaks of the pointer.  Coding standards generally indicate that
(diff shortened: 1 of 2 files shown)


include/wil/com.h
```diff
@@ -1846,46 +1858,90 @@ namespace wil
...
+    {
+        return CoCreateInstance<Interface, err_failfast_policy>(rclsid, dwClsContext);
+    }
+
     /** constructs a COM object using the class as the identifier (that has an associated CLSID) on a specific interface or IUnknown. */
     template<typename Class, typename Interface = IUnknown>
     wil::com_ptr_failfast<Interface> CoCreateInstanceFailFast(DWORD dwClsContext = CLSCTX_INPROC_SERVER)
     {
-        return CoCreateInstance<Class, Interface, err_failfast_policy>(dwClsContext);
+		return CoCreateInstanceFailFast<Interface>(__uuidof(Class), dwClsContext);
```
ChrisGuzak: fix indentation

sylveon: Thanks, resolved. It would be a good idea to add a `.editorconfig` file to the repo so that text editors automatically pick up the desired indentation style.

PR #94, opened 2019-09-20
Fix self-assignment in various types

oldnewthing:
Updated corresponding tests, or wrote missing ones. Added `wistd::exchange` from LLVM since it comes in handy when writing move assignment operators.

| Type                                   | Move self-assignment | Copy self-assignment |
|----------------------------------------|----------------------|----------------------|
| `com_ptr_t`                            | Fixed (Note 1)       |                      |
| `last_error_context`                   | Fixed                |                      |
| `details::string_maker<unique_string>` | Fixed                |                      |
| `details::shared_object`               | Fixed                | Fixed                |
| `vector_range::vector_iterator`        |                      | Fixed                |
| `iterable_range::iterable_iterator`    | (Note 2)             |                      |

Notes:

1. `wil::com_ptr_t` move assignment contained spurious assertion for self-assignment, which triggers on operations like `std::vector v; v.erase(v.begin(), v.begin());`. Code is correct as-is. Remove assertion.

2. `wil::iterable_range::iterable_iterator` move assignment accidentally destroys `m_element`, but that's okay because the next `operator*` will re-create it.

### DIFF
--- a/include/wil/com.h
+++ b/include/wil/com.h
@@ -303,7 +303,6 @@ namespace wil
         //! Move assign from a like `com_ptr_t` (releases current pointer, avoids AddRef/Release by moving the parameter).
         com_ptr_t& operator=(com_ptr_t&& other) WI_NOEXCEPT
         {
-            WI_ASSERT_MSG(this != wistd::addressof(other), "R-Values should be unique: self assignment is a bug");
             attach(other.detach());
             return *this;
         }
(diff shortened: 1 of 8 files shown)


ChrisGuzak:
I see the build failed but did not noticed this until I committed your change. can you prepare another change with the fix?

..\..\tests\ResourceTests.cpp(48,17): error: explicitly moving variable of type 'wil::last_error_context' to itself [-Werror,-Wself-move]
        error42 = std::move(error42);

dunhor:
FWIW you should just be able to add `-Wno-self-move` to `common_build_flags.cmake` (we already have `-Wno-self-assign-overloaded`)

PR #96, opened 2019-09-20
Re-apply Raymond's move assignment change

dunhor:
Add `-Wno-self-move` flag to get Clang build to succeed


dunhor:
I'm going to assume that Chris accidentally closed this instead of merging? Hasn't responded to the question yet, so re-opening for now

PR #108, opened 2019-11-05
Merge with OS source

### LINKED ISSUE #91: Update fail fast handling (FAIL_FAST, FAIL_FAST_IF, etc.) to use RoFailFastWithErrorContext to improve diagnosis of failures and Watson bucketing
There has been feedback that WIL FailFast should use stowed exceptions to get better blaming.  This indeed could be useful as part of FAIL_FAST_IF_FAILED calls.

In addition, the POF logic could also be useful as in the FAIL_FAIL_FAST_IF_FAILED calls that did RoOriginate, you can end up with a crash dump from the point where it actually failed.

We should rationalize what the best way to achieve this is (i.e. this could be just calling RoFailFastWithErrorContext as part of wil/result_originate.h if there is a stowed exception).

dunhor:
The most notable change is the removal of the `unique_hglobal&` constructor in `unique_hglobal_locked` (needed for a build break; the definition guards don't match up for the two - it's easy enough to work around by just calling `.get()`)

Fixes #91

### DIFF
--- a/include/wil/com.h
+++ b/include/wil/com.h
@@ -397,9 +397,9 @@ namespace wil
 
         //! Returns the address of the internal pointer casted to IUnknown** (releases ownership of the pointer BEFORE returning the address).
         //! @see put
-        IUnknown** put_unknown() WI_NOEXCEPT
+        ::IUnknown** put_unknown() WI_NOEXCEPT
         {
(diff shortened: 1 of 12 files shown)


dunhor:
Oh, right. Fixes #91

PR #115, opened 2020-02-10
Enable access to the last error value in last_error_context

NatiMiller:
Add an option to get the last error kept in object. (In order to be displayed by log messages.)

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -73,9 +73,9 @@ namespace wil
         }
 
         auto value() const
-		{
-			return m_error;
-		}
+        {
+            return m_error;
+        }
         
         last_error_context(last_error_context&& other) WI_NOEXCEPT
         {
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -72,6 +72,11 @@ namespace wil
         {
         }
 
+        DWORD get_last_error()
+		{
+			return m_error;
+		}
+        
         last_error_context(last_error_context&& other) WI_NOEXCEPT
         {
             operator=(wistd::move(other));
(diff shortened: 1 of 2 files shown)


sylveon:
I feel like `get_last_error()` is a misnomer because the actual last error might have changed since the object was created.

NatiMiller:
Hi Charles,
Function get_last_error() is required and is correct.
Sometimes user wants to get last error that was saved by object and was taken following specific action, even though the current last error is different.
The aim of this object is to save last error at specific time, while probably the real last error is different at time of destruction of this object.

dunhor:
I'll agree with @sylveon here. Something like `value()` seems like a better name for the function.

include/wil/resource.h
```diff
@@ -72,6 +72,11 @@ namespace wil
         {
         }
 
+        DWORD get_last_error()
+		{
```
dunhor: Spaces please. And ideally order after the destructor (or at least not sandwiched between two constructors)

NatiMiller: Agree.  Fix spaces , change location at file, and change name of function to get_saved_error.

NatiMiller: Agree.  Fix spaces , change location at file, and change name of function to get_saved_error.

include/wil/resource.h
```diff
@@ -72,6 +72,11 @@ namespace wil
         {
         }
 
+        auto value() const
+		{
```
ChrisGuzak, line 76: this is indented too much, perhaps using tabs instead of spaces

ChrisGuzak (changes requested):
Update the PR title to describe the functionality, e.g. "Enable access to the last error value in last_error_context"

ChrisGuzak (approved):

PR #123, opened 2020-04-01
Add exported/installed CMake target

### LINKED ISSUE #118: No clear directions for integrating WIL into CMake projects
Generally speaking, CMake libraries that are intended for use elsewhere will install an exports file - usually in `${INSTALL_DIR}/cmake/<project>Config.cmake` - in order to allow third party CMake projects to locate the library with `find_package(<project>)`. WIL does not appear to provide this CMake configuration file, and so it is unclear how the user is expected to integrate it into other CMake projects, and means that CMake projects that would otherwise use `find_package()` to pull in their dependencies now need to provide custom logic for WIL.

Now, it appears to me that CMake is not strictly used to package the library itself, but rather to do the packaging and testing tasks. I think the library would be well served with an [exported](https://visualgdb.com/tutorials/linux/cmake/find_package/) [interface library target](https://cmake.org/cmake/help/latest/command/add_library.html#interface-libraries) which we could then pull in much more easily.

CJKay:
This commit adds the `WIL` CMake interface library, which allows the
project to be exported and installed for consumption by other CMake
projects in the standard fashion.

The logic to create the NuGet package has been replaced with the
built-in CPack NuGet support introduced with CMake 3.13, which
automatically packages any files and directories installed by the
target. This replaces the `make_wil_nupkg` custom target with the
standard CPack `package` target, which builds the package with all the
configured package generators (for now that's just NuGet).

Note that the original NuGet specification file used the new license
style, whereas CPack still generates a NuGet specification with the
deprecated style. This will hopefully be resolved by future versions of
CPack.

Resolves #118.


dunhor:
It seems like CMake is failing to initialize on the build servers that create and sign the NuGet package. Here's the output:
```
-- The CXX compiler identification is MSVC 19.16.27038.0
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x86/cl.exe
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x86/cl.exe -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-3.12/Modules/WriteBasicConfigVersionFile.cmake:30 (message):
  Unknown keywords given to WRITE_BASIC_CONFIG_VERSION_FILE():
  "ARCH_INDEPENDENT"
Call Stack (most recent call first):
  C:/Program Files (x86)/Microsoft Visual Studio/2017/Enterprise/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-3.12/Modules/CMakePackageConfigHelpers.cmake:211 (write_basic_config_version_file)
  CMakeLists.txt:45 (write_basic_package_version_file)
```
The only thing that stands out is that it seems that these build machines are still using VS 2017 (something out of our control, unfortunately).

dunhor:
There were earlier problems with building the changes on the signing machines (see earlier comment of mine). I haven't checked recently, but I assume those issues still exist. Additionally, #137 from a while ago appears to have addressed the same problem, so there's likely some conflict resolution needed at the very least

CJKay:
Apologies for abandoning this patch on your lap, I've been unable to find the time recently to ensure it's kept up to date.

PR #132, opened 2020-04-13
SAL on ... is not valid

roywill:
Static Analysis fails for resource.h since SAL is not supported on ... argument lists.


jonwis:
Looks fine. Are there instances of SAL on variadic templates too?

PR #135, opened 2020-04-15
Update ifdefs to skip features unsupported by Vista

### LINKED ISSUE #134: resource.h doesn't compile with WINVER=0x0600
Technically, as per the "Update WINVER and _WIN32_WINNT" article from MSDN, targetting Windows Vista is an unsupported scenario, however in WIL headers it's implied to be the minimum supported version:

```cpp
// Minimal requirement set to Windows Vista
#if NTDDI_VERSION < NTDDI_VISTA 
#error WRL requires NTDDI_VERSION to be #defined at least to NTDDI_VISTA or greater
#endif
```

For `resource.h`, this is untrue however - attempting to include it and use it when targeting Vista throws a compilation error in `slim_event_t`, even if said feature is not used:
```
error C3861: 'QueryUnbiasedInterruptTime': identifier not found
```

Fix seems easy enough - this class is already guarded with the following preprocessor directive:
```
#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
```

and so extending it to exclude systems older than Windows 7 allows the code to compile:
```
#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP) && NTDDI_VERSION >= NTDDI_WIN7
```

EDIT:
This change is trivially PR-able, but does it require test coverage too? Windows 7 has test coverage, but since Vista is technically an unsupported target I would expect it not to require similar coverage and a best effort "support" might be sufficient.

CookiePLMonster:
Please check #134 for the rationale behind this PR.

Even though targetting anything older than Windows 7 is technically unsupported by Visual Studio nowadays, WIL itself claims to require the target version to be Vista or newer. This turned out to be untrue in the following cases:

1. `QueryUnbiasedInterruptTime` used in `slim_event_t` in `resource.h` - it was introduced in Windows 7, which makes slim events technically unsupported in Vista.
2. `FindStringOrdinal` in `filesystem.h` - `find_last_path_segment` seems to rely on it, and since it's a heavily used internal function in the entire WIL filesystem functionality, I figured it's best to make it unsupported on Vista. The alternative was to use `StrRChrIW` from `Shlwapi.h`, but that creates a dependency on `Shlwapi.lib` and it would most certainly not be welcome.

As of now, this PR **does** fix existing tests for Vista but **does not** add a new Vista test. I have those prepared locally, so in case they are needed (they should be, really - but I have noticed some concerns regarding inflated compilation times should tests get too extensive) I will commit them as well - my Vista tests are a copypaste of Windows 7 tests, but with `FileSystemTests` and `WatcherTests` removed, due to unsupported filesystem functionality.

Fixes #134

### DIFF
--- a/include/wil/filesystem.h
+++ b/include/wil/filesystem.h
@@ -31,6 +31,7 @@ namespace wil
         return wcsncmp(path, L"\\\\?\\", 4) == 0;
     }
 
+#if (_WIN32_WINNT >= _WIN32_WINNT_WIN7)
     //! Find the last segment of a path. Matches the behavior of shlwapi!PathFindFileNameW()
     //! note, does not support streams being specified like PathFindFileNameW(), is that a bug or a feature?
     inline PCWSTR find_last_path_segment(_In_ PCWSTR path)
@@ -51,6 +52,7 @@ namespace wil
         }
         return result;
     }
+#endif
 
     //! Determine if the file name is one of the special "." or ".." names.
     inline bool path_is_dot_or_dotdot(_In_ PCWSTR fileName)
@@ -83,7 +85,7 @@ namespace wil
         return false;
     }
 
-#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP) && (_WIN32_WINNT >= _WIN32_WINNT_WIN7)
 
     // PathCch.h APIs are only in desktop API for now.
 
@@ -957,7 +959,7 @@ namespace wil
         return result;
     }
 #endif // _CPPUNWIND
-#endif // WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+#endif // WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP) && (_WIN32_WINNT >= _WIN32_WINNT_WIN7)
 }
 
 #endif // __WIL_FILESYSTEM_INCLUDED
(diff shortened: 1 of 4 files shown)


include/wil/filesystem.h
```diff
@@ -83,7 +85,7 @@ namespace wil
         return false;
     }
 
-#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP) && (_WIN32_WINNT >= _WIN32_WINNT_WIN7)
```
dunhor, line 88: This seems like it excludes much more than it needs to

CookiePLMonster: I don't disagree - initially I wanted to only exclude functions which really relied on unsupported features, but I had an impression this would throw this header file into a spiral of ifdefs, as a lot of things in it seems to depend on each other, for example:

- `try_get_parent_path_range` and `GetFullPathNameW` rely on `find_last_path_segment`
- `CreateDirectoryDeepNoThrow` relies on `try_get_parent_path_range`
- `RemoveDirectoryRecursiveNoThrow` relies on `GetFullPathNameW`

I can of course spend some time ifdefing out functions in a more fine-grained fashion - but I am not convinced it makes sense. Both are fine by me.

dunhor: In general, we need to be careful about being too exclusionary for things like this. For example, because MSVC is historically so permissive, there's the chance that someone could be relying on a function that doesn't make use of any offending symbol. That said, I think we're likely fine here. (1) The union of projects building <= Vista and projects using WIL is likely very low, and (2) I _believe_ that the permissive issues generally apply to templates. Since there are offending functions that aren't templated, I have a strong feeling that anyone building <= Vista would receive build errors even under permissive mode.

Issue #131, opened 2020-04-09
Last error lost when calling LOG_*() macros


yves-dolce:
I've not tested all of the error logging macros but by looking at the underlying `ReportFailure<T>()`, you're not using your own `wil::last_error_context()` to preserve the last error.
In that function, this call is resetting the error to 0 for me:
`failure->pszModule = (g_pfnGetModuleName != nullptr) ? g_pfnGetModuleName() : nullptr;`

Is this by design (i.e. one doesn't pay for things she doesn't ask for) or you've never had a need for the last error code after logging it?

Thanks.

JamesWCCheng:
I've met the same issue.

If this by design, can you provide the intention of the design?

thanks

dunhor:
Probably just an oversight. I suspect nobody's noticed/cared since `LOG_*` is generally the "end of the road" for the error code. Feel free to submit a PR

PR #136, opened 2020-04-24
preventing the last error to be modified for LOG_LAST_ERROR* macros

yves-dolce:
This should fix https://github.com/microsoft/wil/issues/131

### DIFF
--- a/include/wil/result_macros.h
+++ b/include/wil/result_macros.h
@@ -664,7 +664,7 @@ WI_ODR_PRAGMA("WIL_FreeMemory", "0")
 
 // Always logs a known failure
 #define LOG_HR(hr)                                              __R_FN(Log_Hr)(__R_INFO(#hr) wil::verify_hresult(hr))
-#define LOG_LAST_ERROR()                                        __R_FN(Log_GetLastError)(__R_INFO_ONLY(nullptr))
+#define LOG_LAST_ERROR()                                        (wil::last_error_context{}, __R_FN(Log_GetLastError)(__R_INFO_ONLY(nullptr)))
 #define LOG_WIN32(win32err)                                     __R_FN(Log_Win32)(__R_INFO(#win32err) win32err)
 #define LOG_NTSTATUS(status)                                    __R_FN(Log_NtStatus)(__R_INFO(#status) status)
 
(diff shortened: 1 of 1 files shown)


dunhor:
My understanding is that the problem is specific to failure paths. For something like `LOG_LAST_ERROR`, this solution seems fine, but seems like unnecessary overhead for the ones that are conditional. I haven't dug too deeply into it, but seems like this is better addressed in either the `Log_*` or `ReportFailure_*` function(s)

include/wil/result_macros.h
```diff
@@ -664,7 +664,7 @@ WI_ODR_PRAGMA("WIL_FreeMemory", "0")
 
 // Always logs a known failure
 #define LOG_HR(hr)                                              __R_FN(Log_Hr)(__R_INFO(#hr) wil::verify_hresult(hr))
-#define LOG_LAST_ERROR()                                        __R_FN(Log_GetLastError)(__R_INFO_ONLY(nullptr))
+#define LOG_LAST_ERROR()                                        (wil::last_error_context{}, __R_FN(Log_GetLastError)(__R_INFO_ONLY(nullptr)))
```
ChrisGuzak, line 667: I'm concerned about code bloat. If this is moved into the Log_GetLast* functions the code can be shared.

ChrisGuzak: correctness should trump efficiency, so take my comment as a refinement.

jonwis (approved):

ChrisGuzak (approved):

dunhor:
FYI - this needed to be partially reverted. In particular, moving `last_error_context` broke building too much kernel code.

More specifically, there's a define - `__WIL_MIN_KERNEL` - that is set in some scenarios to disable some functionality in WIL for kernel components that can't use it/don't want it. This turns out to be an issue because there are some consumers who set this define (either intentionally or through inclusion of another header) and cannot build with the definition of `last_error_context`, and some who have this define set who cannot build without the definition of `last_error_context`. This ultimately boils down to a header inclusion order issue, which is too much of a mess to cleanly fix without risk of breaking things in other ways.

The change put the definition of `last_error_context` back in `resource.h` and then manually calls `SetLastError` in `result_macros.h`. This leads to a couple obvious differences:

1. The failfast/exception code paths will not restore the error
2. If the logging callback were to, say, throw an exception, there's no destructor to run and therefore the error will not be restored

(2) is unlikely to be an issue in practice since the callback should not be throwing exceptions (it may even be declared `noexcept`, I'm not sure). (1) could _maybe_ be a problem for the exception code path, however I'd be surprised if the exception throwing/unwinding code path was guaranteed to preserve the last error.

PR #137, opened 2020-05-02
cmake : Add install target to support consuming WIL through cmake.

p-groarke:
These changes allow a user to install WIL using cmake and consume it through available cmake mechanisms like `find_package`, `ExternalProject_Add` or `FetchContent`.

You can install it locally using the following command :
```bash
cmake .. -DCMAKE_INSTALL_PREFIX=install -DWIL_BUILD_TESTS=Off && cmake --build . && cmake --build . --target install
```

- Adds an `INTERFACE` target for the headers.
- Adds an install config to export the necessary target find info.
- Adds options to disable packaging and tests since you usually don't want to build those when consuming a lib (default on to prevent breaking existing builds).
- Fix the the path variable used in the tests cmake file to support consuming using FetchContent.

Unit tests were ran and passed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,12 @@
 cmake_minimum_required(VERSION 3.11)
 project(WIL)
 
+include(GNUInstallDirs)
+
 # Set by build server to speed up build/reduce file/object size
 option(FAST_BUILD "Sets options to speed up build/reduce obj/executable size" OFF)
-option(WIL_BUILD_PACKAGING "Sets options build the packaging, default on" On)
-option(WIL_BUILD_TESTS "Sets options build the unit tests, default on" On)
+option(WIL_BUILD_PACKAGING "Sets option to build the packaging, default on" On)
+option(WIL_BUILD_TESTS "Sets option to build the unit tests, default on" On)
 
 if (NOT DEFINED WIL_BUILD_VERSION)
     set(WIL_BUILD_VERSION "0.0.0")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,8 +1,12 @@
 cmake_minimum_required(VERSION 3.11)
 project(WIL)
 
+include(GNUInstallDirs)
```
p-groarke, line 4: Required by install.

CMakeLists.txt
```diff
@@ -17,5 +21,31 @@ else()
...
+if (${WIL_BUILD_PACKAGING})
+    add_subdirectory(packaging)
+endif()
+
+if (${WIL_BUILD_TESTS})
+    add_subdirectory(tests)
+endif()
+
+# Gather headers into an interface library.
+file(GLOB_RECURSE HEADER_FILES "${PROJECT_SOURCE_DIR}/include/${PROJECT_NAME}/*.h")
```
p-groarke, line 33: Adds the INTERFACE target.

CMakeLists.txt
```diff
@@ -17,5 +21,31 @@ else()
...
+add_library(${PROJECT_NAME} INTERFACE)
+
+# The interface's include directory.
```
p-groarke, line 36: The following installs all cmake required files to find the library.

tests/CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
-
-include(${CMAKE_SOURCE_DIR}/cmake/common_build_flags.cmake)
+include(${PROJECT_SOURCE_DIR}/cmake/common_build_flags.cmake)
```
p-groarke, line 1: `PROJECT_SOURCE_DIR` is required when running through ExternalProject or FetchContent.

Issue #114, opened 2020-01-30
SafeArray Support


carlos-zamora:
It'd be nice if there was a way to create safe arrays and destroy them when something goes wrong.

Doing something like this...
```
wil::unique_any<SAFEARRAY*, decltype(&::SafeArrayDestroy), ::SafeArrayDestroy>;
```

and making it easily convertible to a vector or array.

ChrisGuzak:
feel free to create a PR that adds this. I see several impls of this in the windows code, here is one to clone.

`using unique_safearray = wil::unique_any<SAFEARRAY*, decltype(&::SafeArrayDestroy), ::SafeArrayDestroy>;`

PR #145, opened 2020-08-22
Support for Safearrays!

dwcullop:
After reading issue #114 I thought that would be a good place for me to start contributing to WIL.  So, I've added two main class templates:  one for creating safearrays and one for accessing their data (ranged-for loops anyone?).  The safearray class can either be generic where you must supply the type or it can have a specific datatype be baked in, and there is a typedef for all the combinations of types and error policies.

I wanted to add some helper functions that perform the most common tasks, but thought it would be better to do that after getting some feedback on what those should look like.

I added a lot of sample usages and tried to follow the source documentation convention used by the other files.

All feedback is appreciated.

### DIFF
--- a/cmake/common_build_flags.cmake
+++ b/cmake/common_build_flags.cmake
@@ -45,16 +45,6 @@ if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
 
         # We don't want legacy MSVC conformance
         -fno-delayed-template-parsing
-
-        # NOTE: Windows headers not clean enough for us to realistically attempt to start fixing these errors yet. That
-        # said, errors that originate from WIL headers may benefit
-        # -fno-ms-compatibility
-        # -ferror-limit=999
-        # -fmacro-backtrace-limit=0
-
-        # -fno-ms-compatibility turns off preprocessor compatability, which currently only works when __VA_OPT__ support
-        # is available (i.e. >= C++20)
-        # -Xclang -std=c++2a
         )
 else()
     add_compile_options(
@@ -63,12 +53,6 @@ else()
 
         # wistd::function has padding due to alignment. This is expected
         /wd4324
-
-        # TODO: https://github.com/Microsoft/wil/issues/6
-        # /experimental:preprocessor
-
-        # CRT headers are not yet /experimental:preprocessor clean, so work around the known issues
-        # /Wv:18
 
         # Some tests have a LOT of template instantiations
         /bigobj
--- a/include/wil/Tracelogging.h
+++ b/include/wil/Tracelogging.h
@@ -743,7 +743,7 @@ public:
     __pragma(warning(push)) __pragma(warning(disable : 4127)) do \
     { \
         _tlgActivityDecl(activity) \
-            TraceLoggingWriteActivity(TraceLoggingType::Provider(), (name), _tlgActivityRef(activity).Id(), NULL, __VA_ARGS__); \
+            TraceLoggingWriteActivity(TraceLoggingType::Provider(), (name), _tlgActivityRef(activity).Id(), NULL, ##__VA_ARGS__); \
     } \
     while (0) \
     __pragma(warning(pop)) /// @endcond
@@ -1143,10 +1143,10 @@ private:
         return (m_sharedActivityData ? m_sharedActivityData->LockExclusive() : rwlock_release_exclusive_scope_exit());
     }
 
-    template <typename ActivityTraceLoggingType, typename TlgReflectorTag = _TlgReflectorTag_Param0IsProviderType>
-    class ActivityData : public _TlgActivityBase<ActivityData<ActivityTraceLoggingType, TlgReflectorTag>, keyword, level>
+    template <typename ActivityTraceLoggingTypeOther, typename TlgReflectorTagOther = _TlgReflectorTag_Param0IsProviderType>
+    class ActivityData : public _TlgActivityBase<ActivityData<ActivityTraceLoggingTypeOther, TlgReflectorTagOther>, keyword, level>
     {
-        using BaseTy = _TlgActivityBase<ActivityData<ActivityTraceLoggingType, TlgReflectorTag>, keyword, level>;
+        using BaseTy = _TlgActivityBase<ActivityData<ActivityTraceLoggingTypeOther, TlgReflectorTagOther>, keyword, level>;
         friend BaseTy;
         void OnStarted()
         {
@@ -1155,7 +1155,7 @@ private:
         {
         }
 
-        // SFINAE dispatching on presence of ActivityTraceLoggingType::CreateActivityId(_Out_ GUID& childActivityId, _In_opt_ const GUID* relatedActivityId)
+        // SFINAE dispatching on presence of ActivityTraceLoggingTypeOther::CreateActivityId(_Out_ GUID& childActivityId, _In_opt_ const GUID* relatedActivityId)
         template <typename ProviderType>
         auto CreateActivityIdByProviderType(int, _Out_ GUID& childActivityId)
             -> decltype(ProviderType::CreateActivityId(childActivityId, this->GetRelatedId()), (void)0)
@@ -1171,7 +1171,7 @@ private:
 
         void CreateActivityId(_Out_ GUID& childActivityId)
         {
-            CreateActivityIdByProviderType<ActivityTraceLoggingType>(0, childActivityId);
+            CreateActivityIdByProviderType<ActivityTraceLoggingTypeOther>(0, childActivityId);
         }
 
     public:
@@ -1221,7 +1221,7 @@ private:
 
         static TraceLoggingHProvider Provider()
         {
-            return ActivityTraceLoggingType::Provider();
+            return ActivityTraceLoggingTypeOther::Provider();
         }
 
         WI_NODISCARD bool NeedsStopped() const WI_NOEXCEPT
@@ -1299,15 +1299,17 @@ private:
 #define __WI_TraceLoggingWriteStart(activity, name, ...) \
     __pragma(warning(push)) __pragma(warning(disable : 4127)) do \
     { \
-        _tlgActivityDecl(activity) static const UINT64 _tlgActivity_Keyword = _tlgActivityRef(activity).Keyword; \
-        static const UINT8 _tlgActivity_Level = _tlgActivityRef(activity).Level; \
-        static const UINT64 _tlgActivityPrivacyTag = _tlgActivityRef(activity).PrivacyTag; \
+        _tlgActivityDecl(activity); \
+        using _tlg_Activity_t = wistd::remove_reference_t<decltype(activity)>; \
+        static constexpr const UINT64 _tlgActivity_Keyword = _tlg_Activity_t::Keyword; \
+        static constexpr const UINT8 _tlgActivity_Level = _tlg_Activity_t::Level; \
+        static constexpr const UINT64 _tlgActivityPrivacyTag = _tlg_Activity_t::PrivacyTag; \
         static_assert( \
-            _tlgActivity_Keyword == (_tlgActivity_Keyword _tlg_FOREACH(_tlgKeywordVal, __VA_ARGS__)), \
+            _tlgActivity_Keyword == (_tlgActivity_Keyword _tlg_FOREACH(_tlgKeywordVal, ##__VA_ARGS__)), \
             "Do not use TraceLoggingKeyword in TraceLoggingWriteStart. Keywords for START events are " \
             "specified in the activity type, e.g. TraceLoggingActivity<Provider,Keyword,Level>."); \
         static_assert( \
-            _tlgActivity_Level == (_tlgActivity_Level _tlg_FOREACH(_tlgLevelVal, __VA_ARGS__)), \
+            _tlgActivity_Level == (_tlgActivity_Level _tlg_FOREACH(_tlgLevelVal, ##__VA_ARGS__)), \
             "Do not use TraceLoggingLevel in TraceLoggingWriteStart. The Level for START events is " \
             "specified in the activity type, e.g. TraceLoggingActivity<Provider,Keyword,Level>."); \
         _tlgActivityRef(activity).zInternalStart(); \
@@ -1321,28 +1323,30 @@ private:
             TraceLoggingLevel(_tlgActivity_Level), \
             TelemetryPrivacyDataTag(_tlgActivityPrivacyTag), \
             TraceLoggingDescription("~^" _wiltlg_LSTRINGIZE(activity) L"^~"), \
-            __VA_ARGS__); \
+            ##__VA_ARGS__); \
     } \
     while (0) \
     __pragma(warning(pop))
 
 #define __WRITE_ACTIVITY_START(EventId, ...) \
     __TRACELOGGING_TEST_HOOK_ACTIVITY_START(); \
-    __WI_TraceLoggingWriteStart(*this, #EventId, __ACTIVITY_START_PARAMS(), __VA_ARGS__); \
+    __WI_TraceLoggingWriteStart(*this, #EventId, __ACTIVITY_START_PARAMS(), ##__VA_ARGS__); \
     EnsureWatchingCurrentThread()
 
 #define __WI_TraceLoggingWriteStop(activity, name, ...) \
     __pragma(warning(push)) __pragma(warning(disable : 4127)) do \
     { \
-        _tlgActivityDecl(activity) static const UINT64 _tlgActivity_Keyword = _tlgActivityRef(activity).Keyword; \
-        static const UINT8 _tlgActivity_Level = _tlgActivityRef(activity).Level; \
-        static const UINT64 _tlgActivityPrivacyTag = _tlgActivityRef(activity).PrivacyTag; \
+        _tlgActivityDecl(activity); \
+        using _tlg_Activity_t = wistd::remove_reference_t<decltype(activity)>; \
+        static constexpr const UINT64 _tlgActivity_Keyword = _tlg_Activity_t::Keyword; \
+        static constexpr const UINT8 _tlgActivity_Level = _tlg_Activity_t::Level; \
+        static constexpr const UINT64 _tlgActivityPrivacyTag = _tlg_Activity_t::PrivacyTag; \
         static_assert( \
-            _tlgActivity_Keyword == (_tlgActivity_Keyword _tlg_FOREACH(_tlgKeywordVal, __VA_ARGS__)), \
+            _tlgActivity_Keyword == (_tlgActivity_Keyword _tlg_FOREACH(_tlgKeywordVal, ##__VA_ARGS__)), \
             "Do not use TraceLoggingKeyword in TraceLoggingWriteStop. Keywords for STOP events are " \
             "specified in the activity type, e.g. TraceLoggingActivity<Provider,Keyword,Level>."); \
         static_assert( \
-            _tlgActivity_Level == (_tlgActivity_Level _tlg_FOREACH(_tlgLevelVal, __VA_ARGS__)), \
+            _tlgActivity_Level == (_tlgActivity_Level _tlg_FOREACH(_tlgLevelVal, ##__VA_ARGS__)), \
             "Do not use TraceLoggingLevel in TraceLoggingWriteStop. The Level for STOP events is " \
             "specified in the activity type, e.g. TraceLoggingActivity<Provider,Keyword,Level>."); \
         _tlgActivityRef(activity).zInternalStop(); \
@@ -1356,7 +1360,7 @@ private:
             TraceLoggingLevel(_tlgActivity_Level), \
             TelemetryPrivacyDataTag(_tlgActivityPrivacyTag), \
             TraceLoggingDescription("~^" _wiltlg_LSTRINGIZE(activity) L"^~"), \
-            __VA_ARGS__); \
+            ##__VA_ARGS__); \
     } \
     while (0) \
     __pragma(warning(pop))
@@ -1369,23 +1373,23 @@ private:
         auto& failure = *pFailure; \
         if (false, WI_IsAnyFlagSet(Keyword, (MICROSOFT_KEYWORD_TELEMETRY | MICROSOFT_KEYWORD_MEASURES | MICROSOFT_KEYWORD_CRITICAL_DATA))) \
         { \
-            __WI_TraceLoggingWriteStop(*this, #EventId, __ACTIVITY_STOP_TELEMETRY_FAILURE_PARAMS(failure), __VA_ARGS__); \
+            __WI_TraceLoggingWriteStop(*this, #EventId, __ACTIVITY_STOP_TELEMETRY_FAILURE_PARAMS(failure), ##__VA_ARGS__); \
         } \
         else \
         { \
-            __WI_TraceLoggingWriteStop(*this, #EventId, __ACTIVITY_STOP_TRACELOGGING_FAILURE_PARAMS(failure), __VA_ARGS__); \
+            __WI_TraceLoggingWriteStop(*this, #EventId, __ACTIVITY_STOP_TRACELOGGING_FAILURE_PARAMS(failure), ##__VA_ARGS__); \
         } \
     } \
     else \
     { \
         __TRACELOGGING_TEST_HOOK_ACTIVITY_STOP(nullptr, GetResult()); \
-        __WI_TraceLoggingWriteStop(*this, #EventId, __ACTIVITY_STOP_PARAMS(GetResult()), __VA_ARGS__); \
+        __WI_TraceLoggingWriteStop(*this, #EventId, __ACTIVITY_STOP_PARAMS(GetResult()), ##__VA_ARGS__); \
     } \
     IgnoreCurrentThread();
 
 // optional params are:  KeyWord, Level, PrivacyTags, Options
 #define __BEGIN_TRACELOGGING_ACTIVITY_CLASS(ActivityClassName, ...) \
-    class ActivityClassName final : public wil::ActivityBase<TraceLoggingType, __VA_ARGS__> \
+    class ActivityClassName final : public wil::ActivityBase<TraceLoggingType, ##__VA_ARGS__> \
     { \
     protected: \
         void StopActivity() WI_NOEXCEPT override \
@@ -1547,13 +1551,13 @@ public: \
 #define DEFINE_TAGGED_TRACELOGGING_EVENT(EventId, ...) \
     void EventId() \
     { \
-        __WI_TraceLoggingWriteTagged(*this, #EventId, _GENERIC_PARTB_FIELDS_ENABLED, __VA_ARGS__); \
+        __WI_TraceLoggingWriteTagged(*this, #EventId, _GENERIC_PARTB_FIELDS_ENABLED, ##__VA_ARGS__); \
     }
 #else
 #define DEFINE_TAGGED_TRACELOGGING_EVENT(EventId, ...) \
     void EventId() \
     { \
-        __WI_TraceLoggingWriteTagged(*this, #EventId, __VA_ARGS__); \
+        __WI_TraceLoggingWriteTagged(*this, #EventId, ##__VA_ARGS__); \
     }
 #endif
 
@@ -1562,13 +1566,13 @@ public: \
     void EventId(PCSTR correlationVector) \
     { \
         __WI_TraceLoggingWriteTagged( \
-            *this, #EventId, _GENERIC_PARTB_FIELDS_ENABLED, TraceLoggingString(correlationVector, "__TlgCV__"), __VA_ARGS__); \
+            *this, #EventId, _GENERIC_PARTB_FIELDS_ENABLED, TraceLoggingString(correlationVector, "__TlgCV__"), ##__VA_ARGS__); \
     }
 #else
 #define DEFINE_TAGGED_TRACELOGGING_EVENT_CV(EventId, ...) \
     void EventId(PCSTR correlationVector) \
     { \
-        __WI_TraceLoggingWriteTagged(*this, #EventId, TraceLoggingString(correlationVector, "__TlgCV__"), __VA_ARGS__); \
+        __WI_TraceLoggingWriteTagged(*this, #EventId, TraceLoggingString(correlationVector, "__TlgCV__"), ##__VA_ARGS__); \
     }
 #endif
 
@@ -1582,7 +1586,7 @@ public: \
             #EventId, \
             TraceLoggingValue(static_cast<VarType1>(wistd::forward<T1>(varName1)), _wiltlg_STRINGIZE(varName1)), \
             _GENERIC_PARTB_FIELDS_ENABLED, \
-            __VA_ARGS__); \
+            ##__VA_ARGS__); \
     }
 #else
 #define DEFINE_TAGGED_TRACELOGGING_EVENT_PARAM1(EventId, VarType1, varName1, ...) \
@@ -1590,7 +1594,7 @@ public: \
     void EventId(T1&& varName1) \
     { \
         __WI_TraceLoggingWriteTagged( \
-            *this, #EventId, TraceLoggingValue(static_cast<VarType1>(wistd::forward<T1>(varName1)), _wiltlg_STRINGIZE(varName1)), __VA_ARGS__); \
+            *this, #EventId, TraceLoggingValue(static_cast<VarType1>(wistd::forward<T1>(varName1)), _wiltlg_STRINGIZE(varName1)), ##__VA_ARGS__); \
     }
 #endif
 
@@ -1605,7 +1609,7 @@ public: \
             TraceLoggingValue(static_cast<VarType1>(wistd::forward<T1>(varName1)), _wiltlg_STRINGIZE(varName1)), \
             _GENERIC_PARTB_FIELDS_ENABLED, \
             TraceLoggingString(correlationVector, "__TlgCV__"), \
-            __VA_ARGS__); \
+            ##__VA_ARGS__); \
     }
 #else
 #define DEFINE_TAGGED_TRACELOGGING_EVENT_PARAM1_CV(EventId, VarType1, varName1, ...) \
@@ -1617,7 +1621,7 @@ public: \
             #EventId, \
             TraceLoggingValue(static_cast<VarType1>(wistd::forward<T1>(varName1)), _wiltlg_STRINGIZE(varName1)), \
             TraceLoggingString(correlationVector, "__TlgCV__"), \
-            __VA_ARGS__); \
+            ##__VA_ARGS__); \
     }
 #endif
 
@@ -1632,7 +1636,7 @@ public: \
             TraceLoggingValue(static_cast<VarType1>(wistd::forward<T1>(varName1)), _wiltlg_STRINGIZE(varName1)), \
             TraceLoggingValue(static_cast<VarType2>(wistd::forward<T2>(varName2)), _wiltlg_STRINGIZE(varName2)), \
             _GENERIC_PARTB_FIELDS_ENABLED, \
-            __VA_ARGS__); \
+            ##__VA_ARGS__); \
     }
 #else
 #define DEFINE_TAGGED_TRACELOGGING_EVENT_PARAM2(EventId, VarType1, varName1, VarType2, varName2, ...) \
@@ -1644,7 +1648,7 @@ public: \
             #EventId, \
             TraceLoggingValue(static_cast<VarType1>(wistd::forward<T1>(varName1)), _wiltlg_STRINGIZE(varName1)), \
             TraceLoggingValue(static_cast<VarType2>(wistd::forward<T2>(varName2)), _wiltlg_STRINGIZE(varName2)), \
-            __VA_ARGS__); \
+            ##__VA_ARGS__); \
     }
 #endif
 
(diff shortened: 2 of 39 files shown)


dwcullop:
> Maybe also add `unique_safearray<T>`?

There's already `unique_safearray` which is a no-throw version of the generic class that can handle any SAFEARRAY type (but you provide the type when necessary).

include/wil/safearrays.h
```diff
@@ -0,0 +1,897 @@
...
+        template<> struct var_traits<CURRENCY>                   { static constexpr auto vartype = VT_CY; };
+        template<> struct var_traits<DECIMAL>                    { static constexpr auto vartype = VT_DECIMAL; };
+        template<> struct var_traits<BSTR>                       { static constexpr auto vartype = VT_BSTR; };
+        template<> struct var_traits<LPUNKNOWN>                  { static constexpr auto vartype = VT_UNKNOWN; };
+        template<> struct var_traits<LPDISPATCH>                 { static constexpr auto vartype = VT_DISPATCH; };
+        template<> struct var_traits<VARIANT>                    { static constexpr auto vartype = VT_VARIANT; };
+
+        inline void __stdcall SafeArrayDestroy(SAFEARRAY* psa) WI_NOEXCEPT
+        {
+            __FAIL_FAST_ASSERT__(psa != nullptr);
```
dunhor, line 55: This seems unnecessary. Just let the call do a null deref

include/wil/safearrays.h
```diff
@@ -0,0 +1,897 @@
...
+            }
+            return vt;
+        }
+
+        inline ULONG __stdcall SafeArrayGetLockCount(SAFEARRAY* psa) WI_NOEXCEPT
+        {
+            return (psa != nullptr) ? psa->cLocks : 0U;
+        }
+
+        inline HRESULT __stdcall SafeArrayCreate(VARTYPE vt, UINT cDims, SAFEARRAYBOUND* sab, SAFEARRAY*& psa) WI_NOEXCEPT
```
dunhor, line 99: Don't use hungarian

include/wil/safearrays.h
```diff
@@ -0,0 +1,897 @@
...
+        template<> struct var_traits<VARIANT_BOOL>               { static constexpr auto vartype = VT_BOOL; };
+        template<> struct var_traits<DATE>                       { static constexpr auto vartype = VT_DATE; };
+        template<> struct var_traits<CURRENCY>                   { static constexpr auto vartype = VT_CY; };
+        template<> struct var_traits<DECIMAL>                    { static constexpr auto vartype = VT_DECIMAL; };
+        template<> struct var_traits<BSTR>                       { static constexpr auto vartype = VT_BSTR; };
+        template<> struct var_traits<LPUNKNOWN>                  { static constexpr auto vartype = VT_UNKNOWN; };
+        template<> struct var_traits<LPDISPATCH>                 { static constexpr auto vartype = VT_DISPATCH; };
+        template<> struct var_traits<VARIANT>                    { static constexpr auto vartype = VT_VARIANT; };
+
+        inline void __stdcall SafeArrayDestroy(SAFEARRAY* psa) WI_NOEXCEPT
```
dunhor, line 53: Specifying the calling convention seems unnecessary, or am I missing something?

include/wil/safearrays.h
```diff
@@ -0,0 +1,897 @@
...
+    public:
+        // Exception-based construction
+        safearraydata_t(pointer psa)
+        {
+            static_assert(wistd::is_same<void, result>::value, "this constructor requires exceptions; use the create method");
+            access(psa);
+        }
+        safearraydata_t() = default;
+        safearraydata_t(safearraydata_t&&) = default;
+        ~safearraydata_t() = default;
```
dunhor, line 247: Unnecessary

include/wil/safearrays.h
```diff
@@ -0,0 +1,897 @@
...
+        typedef value_type* value_pointer;
+        typedef const value_type* const_value_pointer;
+        typedef value_type& value_reference;
+        typedef const value_type& const_value_reference;
+        typedef value_type* iterator;
+        typedef const value_type* const_iterator;
+
+    public:
+        // Exception-based construction
+        safearraydata_t(pointer psa)
```
dunhor, line 240: Probably want this to be `explicit`

include/wil/safearrays.h
```diff
@@ -0,0 +1,897 @@
...
+    //! @tparam err_policy  Represents the error policy for the class (error codes, exceptions, or fail fast; see @ref page_errors)
+    template <typename T, typename err_policy = err_exception_policy>
+    class safearraydata_t
+    {
+    private:
+        typedef unique_any<SAFEARRAY*, decltype(&details::SafeArrayUnaccessData), details::SafeArrayUnaccessData, details::pointer_access_noaddress> safearray_unaccess_data;
+
+    public:
+        typedef typename err_policy::result result;
+        typedef typename safearray_unaccess_data::pointer pointer;
```
dunhor, line 229: This is not a dependent name, so the `typedef` is unnecessary

include/wil/safearrays.h
```diff
@@ -0,0 +1,897 @@
...
+        safearray_t(UINT cElements, LONG lowerBound = 0)
+        {
+            static_assert(wistd::is_same<void, result>::value, "this constructor requires exceptions; use the create method");
+            create<T>(cElements, lowerBound);
+        }
+
+        // Low-level, arbitrary number of dimensions
+        // Uses Explicit Type
+        template<typename T = element_t, typename wistd::enable_if<is_type_not_set<T>::value, int>::type = 0>
+        result create(VARTYPE vt, UINT cDims, SAFEARRAYBOUND* sab)
```
dunhor, line 391: I'd say follow STL naming convention and call these `reset`

include/wil/safearrays.h
```diff
@@ -0,0 +1,897 @@
...
+#error This header is not supported in kernel-mode.
+#endif
+
+#include <new.h> // new(std::nothrow)
+#include "wistd_type_traits.h"
+#include "resource.h" // unique_hkey
+
+namespace wil
+{
+#if defined( _OLEAUTO_H_ ) && !defined(__WIL_SAFEARRAY_)
```
dunhor, line 24: Given that this is a dedicated header file, the `_OLEAUTO_H_` check seems unnecessary and the file should include its dependencies

include/wil/safearrays.h
```diff
@@ -0,0 +1,897 @@
...
+//
+//    Copyright (c) Microsoft. All rights reserved.
+//    This code is licensed under the MIT License.
+//    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF
+//    ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+//    TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+//    PARTICULAR PURPOSE AND NONINFRINGEMENT.
+//
+//*********************************************************
+#ifndef __WIL_SAFEARRAYS_INCLUDED
```
dunhor, line 11: This has an include guard, but then uses the `resource.h`-style guards on definitions. E.g. you wouldn't be able to include this file twice to "turn on" shared/weak pointer support like you can for the `resource.h` header. IMO that's not a huge deal, but we shouldn't be attempting to do both. We should either make it includable multiple times or remove the definition guards (which likely means removing shared/weak support)

include/wil/safearrays.h
```diff
@@ -0,0 +1,897 @@
...
+
+namespace wil
+{
+#if defined( _OLEAUTO_H_ ) && !defined(__WIL_SAFEARRAY_)
+#define __WIL_SAFEARRAY_
+    /// @cond
+    namespace details
+    {
+        template<typename T>
+        struct var_traits{};
```
dunhor, line 30: The name `var_traits` is a bit generic and could apply to a number of scenarios and would probably benefit from a different name

include/wil/safearrays.h
dwcullop: I'll just remove the outer most guards and leave the two-phase includes, but you shouldn't have to include it twice if you include it last after all the other WIL headers.

include/wil/safearrays.h
dwcullop: It didn't used to be a dedicated header, so the transition may be incomplete.  I just looked at `filesystem.h` and it includes `PathCch.h` and `registry.h` includes `winreg.h`, so I guess it's okay to pull include `oleauto.h`.

include/wil/safearrays.h
dwcullop: I tried to use the same variable names as in the API documentation, which makes it easier to understand for people already familiar with the documentation, but I'll change it.

include/wil/safearrays.h
dwcullop: See [details::SetEvent](https://github.com/microsoft/wil/blob/3c00e7f1d8cf9930bbb8e5be3ef0df65c84e8928/include/wil/resource.h#L2328) in `resource.h`.  That's what I modeled it after.  It specifies the calling convention, presumably to make an identical (except for fail fast) version of the API.

include/wil/safearrays.h
dwcullop: Not sure to which line you're referring.  The move constructor is necessary for the `safearray_unaccess_data` member to move correctly.  The rest might not be strictly necessary, but they do show that I didn't forget about them, and certainly don't hurt anything.

include/wil/safearrays.h
dwcullop: Yes!  This absolutely should be explicit!  Will fix.

include/wil/safearrays.h
dwcullop: I had to add a bunch of those to make clang happy, but maybe that one isn't strictly necessary.  I'll remove it and find out.

include/wil/safearrays.h
dwcullop: I'm not sure about this... In STL that means you giving it a new pointer that you want to reset the smart pointer to.  These functions take parameters to create a new instance of the object.

Let's see if anyone else has an opinion on this because I'm on the fence.

include/wil/safearrays.h
dwcullop: I was originally going to re-use these for my VARIANT helper classes that I was going to do next, and in that case, they have more work to do and `var_traits<>` is absolutely a great name for them.  Since I'm not sure I'm going to do the VARIANT classes, I'll rename them and revisit the issue if/when I do.

I'll rename to `infer_safearray_type_traits`.

include/wil/safearrays.h
dwcullop: I'll remove the `nullptr` check from all of the API-wrappers.

include/wil/safearrays.h
dwcullop: There was also a constructor in `safearray_t` that took 2 parameters (but one has a default value) that also needed to be explicit as well.

include/wil/safearrays.h
dunhor: Sorry, I keep forgetting that Github's PR experience is a bit sub-optimal and doesn't show the highlighted range. I originally meant the destructor, however I realize now that none of the 5 (copy/move constructor/assignment and destructor) are necessary since that's the behavior of `unique_any<>`

include/wil/safearrays.h
dunhor: The issue isn't needing to include it twice, it's avoiding include order dependencies. That is, the following:
```c++
#include <wil/safearray.h>
#include <memory>
#include <wil/safearray.h>
```
Should be equivalent to:
```c++
#include <memory>
#include <wil/safearray.h>
```

include/wil/safearrays.h
dunhor: I suppose `assign` would perhaps be a better option

include/wil/safearrays.h
```diff
@@ -0,0 +1,889 @@
...
+        const_iterator begin() const WI_NOEXCEPT { WI_ASSERT(m_begin != nullptr); return m_begin; }
+        const_iterator end() const WI_NOEXCEPT { WI_ASSERT(m_begin != nullptr); return m_begin+m_size; }
+
+        // Old style iteration
+        ULONG size() const WI_NOEXCEPT { return m_size; }
+        WI_NODISCARD value_reference operator[](ULONG i) { WI_ASSERT(i < m_size); return *(m_begin + i); }
+        WI_NODISCARD const_value_reference operator[](ULONG i) const { WI_ASSERT(i < m_size); return *(m_begin +i); }
+
+        // Utilities
+        bool empty() const WI_NOEXCEPT { return m_size == ULONG{}; }
```
dunhor, line 270: Just compare to zero (or `0ul` if sign comparison is giving warnings, which I'd hope it wouldn't)

include/wil/safearrays.h
```diff
@@ -0,0 +1,889 @@
...
+            psa = ::SafeArrayCreate(vt, dims, sab);
+            RETURN_LAST_ERROR_IF_NULL(psa);
+            WI_ASSERT(vt == details::SafeArrayGetVartype(psa));
+            return S_OK;
+        }
+
+        inline HRESULT __stdcall SafeArrayDimElementCount(SAFEARRAY* psa, UINT dims, ULONG* result) WI_NOEXCEPT
+        {
+            __FAIL_FAST_ASSERT__(psa != nullptr);
+            __FAIL_FAST_ASSERT__(result != nullptr);
```
dunhor, line 106: Both of these asserts are unnecessary; you dereference both below. In general, asserting against null is discouraged

include/wil/safearrays.h
```diff
@@ -0,0 +1,889 @@
...
+        }
+        //! 1D Specialization of ubound
+        result ubound(LONG* result) const
+        {
+            WI_ASSERT(dims() == 1);
+            WI_ASSERT(result != nullptr);
+            return err_policy::HResult(::SafeArrayGetUBound(storage_t::get(), 1, result));
+        }
+        //! Indicates the total number of elements across all the dimensions
+        result count(ULONG* result) const
```
dunhor, line 467: `size` is more consistent with STL containers

include/wil/safearrays.h
```diff
@@ -0,0 +1,889 @@
...
+        //!         DoSomethingWithBSTR(bstr);
+        //!     }
+        //!     return S_OK;
+        //! }
+        //! ~~~~
+        //! @{
+
+        //! Copies the specified element and puts it into the specified index in the SAFEARRAY (1D version)
+        //! The SAFEARRAY will own any resources and will clean up when it is destroyed
+        template<typename T = element_t, typename wistd::enable_if<is_value_type<T>::value, int>::type = 0>
```
dunhor, line 543: The default type of `T` here and below seems misleading as it'll never actually come in to play. The type of the argument will always dictate what `T` is.

include/wil/safearrays.h
```diff
@@ -0,0 +1,889 @@
...
+        result put_element(LONG index, T val)
+        {
+            WI_ASSERT(sizeof(val) == elemsize());
+            WI_ASSERT(dims() == 1);
+            return err_policy::HResult(::SafeArrayPutElement(storage_t::get(), &index, val));
+        }
+        //! Retrieves a copy of the element at the specified index.  The caller owns this copy and must free
+        //! any resources with LPUNKNOWN->Release, SysFreeString, VariantClear, etc.
+        template<typename T = element_t, typename wistd::enable_if<is_type_set<T>::value, int>::type = 0>
+        result get_element(LONG index, T& val)
```
dunhor, line 560: Seems like this should be a `T`-returning function for exceptions/fail-fast. IIRC I believe `com_ptr` has support for similar things.

include/wil/safearrays.h
dwcullop: Some STL containers use `count`, but all of them use `size`, so I will change it.  I suppose "Count" came from the shim function "SafeArrayCountElements", so that will become "SafeArrayGetTotalSize".  I guess `dim_elements` should become `dim_size`, and it's shim should be "SafeArrayGetDimSize", which gives them a symmetry they didn't have before.

include/wil/safearrays.h
dwcullop: Yes, if T is provided, it will always use that T.  However, for the cases when element_t is not void, it will come into play because you don't have to provide the T at all.  I agree it is unusual to have a single template parameter change the shape of the class so dramatically.  I'm really implementing two classes:  one where the type is built-in and one where it isn't and always has to be provided when needed.

include/wil/safearrays.h
dwcullop: There has to be error handling because the Get operation could fail.  I could do a return T for `err_exception_policy` but I didn't because for some types, like BSTR or LPUNKNOWN, the caller would be responsible for cleaning up the resources and that isn't obvious and it makes it too hard to use correctly.

What I could do is add overloads the return the resource owned by appropriate WIL class (`wil::unique_bstr`) and only enable them for when the element type is provided and appropriate.  Actually, I can do that for all of the error policies with the extra "easy to use" ones for the exception based policy.

include/wil/safearrays.h
dwcullop: Yeah, I don't see a lot of other nullptr asserts in WIL, so I'll take them out.

include/wil/safearrays.h
dwcullop: I think I originally had the size type be a template parameter, which is why it ended up like that.  I have no problem changing it.

include/wil/safearrays.h
dunhor: I'm not sure I follow. `T` is the type of the second argument in the function. The caller must provide the second argument which will dictate what `T` is. The default here has no effect.

include/wil/safearrays.h
dwcullop: I think they are necessary.  `safearraydata_t` isn't a `unique_any` type because it needs additional data members (pointer and size).  It **has** a `unique_any` type that ensures `SafeArrayUnaccessData` is called properly and that requires the class to be "move only", so I defaulted the ones that are okay and deleted the ones that aren't.

Which do you think I should remove?

include/wil/safearrays.h
dwcullop: I think I had the size original as a size_type typedef, which is why I wrote it like that.  No issue changing it.

include/wil/safearrays.h
dwcullop: I am not sure what part of `com_ptr` you mean.  Can you please cite a line number, @dunhor ?

dunhor: Hopefully this works. [Here is an example from com_ptr](https://github.com/microsoft/wil/blob/3c00e7f1d8cf9930bbb8e5be3ef0df65c84e8928/include/wil/com.h#L487)

That's a good point for pointer/resource-owning types. If memory serves me correctly, there's a map-raw-to-smart-type helper, though I'm not sure if that would apply here (e.g. I'm unsure if it handles `BSTR`, etc.), though I'm not entirely sure where that lives. It probably wouldn't be _too_ terrible to have a separate one for this purpose, I think.

include/wil/safearrays.h
dwcullop: I see what you're saying.  You're only objecting to the `= element_t` part.  It's there so that all the functions are the same and easier to visually validate.  If you think it's misleading, I'll look at removing it.

include/wil/safearrays.h
dwcullop: Okay, I can do that for the non-error code case.  The mapping to raw pointers (currently) only works for Interface smart pointers.  It would be better to extend that to work with any of the generic types so that we can directly support other BSTR / VARIANT wrappers, but I'm working just supporting the WIL wrappers for now.

dwcullop: That was a lot more work than I thought it would be, but it will be very nice for people who want to use put_element/get_element (although using access_data() is going to be way more performant).

include/wil/safearrays.h
dunhor: Right. Members affect default constructors/assignment operators in much the same - if not the exact same - way as base classes. E.g. see https://godbolt.org/z/cb77s1. In general, you should either define (actually define, not default) all five or let the compiler do the correct thing for you. E.g. see [the rule of five](https://en.cppreference.com/w/cpp/language/rule_of_three)

include/wil/safearrays.h
dunhor: Correct. Declaring the default suggests that you did so for a functional reason. Since the default will never have an effect, it gives the impression that you may have made a mistake. E.g. perhaps you intended for the argument to be `element_t`.

include/wil/safearrays.h
```diff
@@ -0,0 +1,1084 @@
...
+        //! ~~~~
+        template<typename T = element_t, typename wistd::enable_if<is_type_set<T>::value, int>::type = 0>
+        WI_NODISCARD safearraydata_t<T> access_data() const
+        {
+            static_assert(wistd::is_same<void, result>::value, "this method requires exceptions or failfast");
+            return safearraydata_t<T>{ storage_t::get() };
+        }
+
+        //! Special Get Functions
+#if defined(__WIL_OLEAUTO_H_)
```
dunhor, line 780: This won't work for member functions. You are now introducing an include order dependency which should be avoided

include/wil/safearrays.h
```diff
@@ -0,0 +1,1084 @@
...
+            {
+                ULONGLONG elements = 1;
+                for (UINT i = 0; i < psa->cDims; ++i)
+                {
+                    // Determine the total size of the safearray by multiplying
+                    // the sizes of each dimension together
+                    elements *= psa->rgsabound[i].cElements;
+                    if (elements > ULONG_MAX)
+                    {
+                        RETURN_HR(HRESULT_FROM_WIN32(ERROR_ARITHMETIC_OVERFLOW));
```
dunhor, line 121: This seems kind of excessive and seems to be the only thing stopping something as simple as `size` from returning its result (as opposed to being an out param). Why not just make it a `size_t`?

include/wil/safearrays.h
dwcullop: 1) What won't work for member functions?  I don't understand that part of your comment.
2) I'm not introducing an include order dependency.  That dependency was already there.  People are told to include the system headers they're going to use first, and the WIL will customize itself based on the included headers.  I only want to include that functionality if WIL has supplied the types that I need.

Since (thanks to feedback on this PR) I am now in a separate header and am including the other headers I need, I will make sure to include `OleAuto.h` before including `resource.h`.  That way, if `resource.h` was included before `OleAuto.h` was included the first time, it will pick up these definitions the 2nd time around (assuming all the conditions are met).  Since I don't want to care about those conditions, I just want to add the functionality if they were met, testing for `__WIL_OLEAUTO_H_` will accomplish that.

TL;DR:  I'll include the headers I need in the right order and that will eliminate any order dependencies.

include/wil/safearrays.h
dwcullop: I really wanted to do that, but `size_t` is platform dependent, so we can't assume the size, but it will be 32-bit on some systems, which won't work.  I could define a `size_type` that is always 64-bit, but that solve anything because even a 64-bit value could overflow when multiplying an arbitrary number of 32-bit values.

include/wil/safearrays.h
```diff
@@ -0,0 +1,1106 @@
...
+    //! This class works even if the SAFEARRAY is multi-dimensional by treating it as a large single-dimension array,
+    //! but doesn't support access via a multi-dimensional index.
+    //! NOTE: This class does not manage the lifetime of the SAFEARRAY itself.  See @ref safearray_t.
+    //! ~~~~
+    //! HRESULT copy_to_bstr_vector(SAFEARRAY* psa, std::vector<wil::unique_bstr>& result)
+    //! {
+    //!     auto data = wil::safearraydata_nothrow<BSTR>{};
+    //!     RETURN_IF_FAILED(data.access(psa));
+    //!     result.reserve(result.size() + data.size());
+    //!     for(BSTR& bstr : data)
```
ChrisGuzak, line 170:
```suggestion
    //!     for (BSTR& bstr : data)
```

tests/SafearrayTests.cpp
```diff
@@ -0,0 +1,1574 @@
...
+//static const GUID IID_IAmForTesting =
+//{ 0x5d80ec64, 0x6694, 0x4f49, { 0xb0, 0xb9, 0xcc, 0xaa, 0x65, 0x46, 0x7d, 0x12 } };
+
+interface __declspec(uuid("5D80EC64-6694-4F49-B0B9-CCAA65467D12"))
+    IAmForTesting : public IUnknown
+{
+    STDMETHOD(GetID)(LONG* pOut) = 0;
+};
+
+class TestComObject : virtual public IAmForTesting,
```
ChrisGuzak, line 106: can you use `winrt::implements` here instead to avoid the IUnknown boilerplate? use WINRT_CUSTOM_MODULE_LOCK, get_module_lock to enable global object counting.

dwcullop: Not at all familiar with WinRT but am willing to learn.  It will just take a little while to make that change.

dwcullop: I converted the class to use WinRT for it's COM implementation and then realized that `winrt::implements` requires C++17 for the features that it uses, in which case, I'd rather keep it as-is.

dwcullop:
Apologies for the delay.  I completely forgot I was working on this.  I've resolved the merge conflicts.  Please let me know if there's anything else I need to do to complete this.

magol:
@jonwis @dunhor @ChrisGuzak Any progress?

dunhor:
Sorry for the delay. I'll take a look through for anything glaring, but will defer to @ChrisGuzak or others more familiar with `SAFEARRAY`s & their usage.

Also note that the entire repo has been re-formatted. Since all your changes are in new files, that likely won't cause you major merge conflicts, just note that you'll need to run `clang-format` on the new code.

include/wil/safearrays.h
```diff
@@ -0,0 +1,1105 @@
...
+
+#include <new.h> // new(std::nothrow)
+#include "wistd_type_traits.h"
+#include <OleAuto.h>
+#include "resource.h"
+
+namespace wil
+{
+#if !defined(__WIL_SAFEARRAY_)
+#define __WIL_SAFEARRAY_
```
dunhor, line 24: This should be removed and changed to just a normal include guard

include/wil/safearrays.h
```diff
@@ -0,0 +1,1105 @@
...
+    /// @cond
+    namespace details
+    {
+        template<typename T> struct infer_safearray_type_traits{};
+        template<> struct infer_safearray_type_traits<char>                       { static constexpr auto vartype = VT_I1; };
+        //template<> struct infer_safearray_type_traits<short>                      { static constexpr auto vartype = VT_I2; };
```
dunhor, line 30: What's the reason for this and `double` to be commented out? Does `I2` & `R8` not exist? Comments about why they are not present are more helpful than commented out code

include/wil/safearrays.h
dunhor: I suspect that was just because of copy/paste, but with `extern "C"` functions, you never know... Do you get compiler/linker errors with it removed?

include/wil/safearrays.h
dunhor: I assume that all of the memory for `SAFEARRAY`s must be in virtual memory, no? If that's the case, it's impossible for the total size to exceed `size_t` unless the data is corrupt

include/wil/safearrays.h
```diff
@@ -0,0 +1,1105 @@
...
+        typedef resource_policy<SAFEARRAY*, decltype(&details::SafeArrayDestroy), details::SafeArrayDestroy, details::pointer_access_all> safearray_resource_policy;
+    }
+    /// @endcond
+
+    typedef unique_any<SAFEARRAY*, decltype(&details::SafeArrayUnlock), details::SafeArrayUnlock, details::pointer_access_noaddress> safearray_unlock_scope_exit;
+    typedef unique_any<SAFEARRAY*, decltype(&details::SafeArrayUnaccessData), details::SafeArrayUnaccessData, details::pointer_access_noaddress> safearray_unaccess_scope_exit;
+
+    //! Guarantees a SafeArrayUnlock call on the given object when the returned object goes out of scope
+    //! Note: call SafeArrayUnlock early with the reset() method on the returned object or abort the call with the release() method
+    WI_NODISCARD inline safearray_unlock_scope_exit SafeArrayUnlock_scope_exit(SAFEARRAY* psa) WI_NOEXCEPT
```
dunhor, line 143: Shouldn't this be named more to indicate that it's calling "lock"? E.g. `LockSafeArray_scope_exit` or something?

(I'm also not a fan of including `_scope_exit` in the name, but if that's an established naming pattern, so be it)

include/wil/safearrays.h
```diff
@@ -0,0 +1,1105 @@
...
+    template<typename T>
+    WI_NODISCARD inline safearray_unaccess_scope_exit SafeArrayAccessData_scope_exit(SAFEARRAY* psa, T*& p) WI_NOEXCEPT
+    {
+        details::SafeArrayAccessData(psa, p);
+        return safearray_unaccess_scope_exit(psa);
+    }
+
+    //! Class that facilitates the direct access to the contents of a SAFEARRAY and "unaccessing" it when done
+    //! It enables treating the safearray like an regular array and/or performing a ranged-for on the contents.
+    //! Accessing a safearray increases it's lock count, so attempts to destroy the safearray will fail until it is unaccessed.
```
dunhor, line 160:
```suggestion
    //! Accessing a safearray increases its lock count, so attempts to destroy the safearray will fail until it is unaccessed.
```

include/wil/safearrays.h
```diff
@@ -0,0 +1,1105 @@
...
+    //! This class works even if the SAFEARRAY is multi-dimensional by treating it as a large single-dimension array,
+    //! but doesn't support access via a multi-dimensional index.
+    //! NOTE: This class does not manage the lifetime of the SAFEARRAY itself.  See @ref safearray_t.
```
dunhor, line 163: If it's non-owning, the name should indicate that better. In C++, the names `view` (non-modifiable) or `span` (modifiable) are typically used to indicate that

include/wil/safearrays.h
dwcullop: I'm not sure that's true for 32-bit compilations.

include/wil/safearrays.h
dwcullop: I2 and R8 are tricky to support because they're the same as other types as far as the compiler is concerned.  `short` is the same as `BSTR` and `double` is the same as `DATE`.

Up at the top Chris is suggesting using some dedicated enums for each type that could be confused so they can be explicit.  I'll look into that.

include/wil/safearrays.h
dwcullop: I followed the naming convention used in other places, which is to name the methods and the RAII classes they return after the specific action that is performed (with `_scope_exit` appended).

Examples:
- `semaphore_release_scope_exit`  // Releases the semaphore at scope exit
- `event_reset_scope_exit`        // Invokes Reset Event at scope exit

But you're not wrong that this version is also doing a lock on construction, but at Scope Exit, it unlocks.

include/wil/safearrays.h
dwcullop: The reason is SAFEARRAY uses ULONG for a size.  So, when computing the total size, we have to make sure we don't overflow a ULONG, so I use a ULONGLONG, check for overflow again ULONG_MAX, and then downcast to a ULONG after confirming it's safe.

tests/SafearrayTests.cpp
dwcullop: Unless there's no requirement to support pre-C++17 anymore... Then sure.

include/wil/safearrays.h
dwcullop: It is owning.  When SafeArrayAccessData is invoked, it creates a refcount and a pointer that this class owns until SafeArrayUnaccessData is invoked.

I think calling it view or span would be convey the wrong idea about what is going on.

include/wil/safearrays.h
dunhor: Again, comments about _why_ these types are not present are more helpful than leaving in commented out code. I see the `DATE`/`double` conflict, however `BSTR` is defined as a pointer and shouldn't conflict with `short` - did you mean something else?

But now that I know why they are not present, what's the behavior if I try and use `double` with this code? Does it give me a compilation error? Does it succeed, but silently give me incorrect behavior because values are interpreted incorrectly? Or does it end up working correctly? This is all information that is more useful to have commented in the code than commented out lines like this.

include/wil/safearrays.h
dunhor: The `SAFEARRAY` type seems to use `ULONG` to store the size of each dimension, however I don't see anything in the documentation that would indicate that the product of dimensions can't exceed the bounds of `ULONG`. So, I checked the definition of `SafeArrayCreate` and it appears to use `ULONG` to calculate the size to allocate (despite `IMalloc` accepting a `size_t` argument...).

So `ULONG` is probably fine here, although theoretically the API could be updated in the future to allow for larger allocations. One piece of feedback is to consider using the existing safe math functions that exist in `intsafe.h` instead of rolling your own, but otherwise I concede here.

include/wil/safearrays.h
dunhor: > Name that function with Lock and the RAII class with unlock.

Yeah, that's basically what I was suggesting (with an additional side note that the trailing `_scope_exit` in the function name is excessive). I see that some functions have it, but some others don't (e.g. `wil::CoInitializeEx` etc.)

include/wil/safearrays.h
dunhor: Sorry, by non-owning, I meant that it doesn't own the `SAFEARRAY` object (i.e. not responsible for calling `SafeArrayDestroy`). Alternative naming suggestions would be to include `ref`, `reference`, or `lock` into the data type name. `safearraydata` doesn't quite convey its ownership or responsibility.

include/wil/safearrays.h
```diff
@@ -0,0 +1,1105 @@
...
+        {
+            static_assert(wistd::is_same<void, result>::value, "this constructor requires exceptions; use the create method");
+            access(psa);
+        }
+        safearraydata_t() = default;
+        safearraydata_t(safearraydata_t&&) = default;
+        ~safearraydata_t() = default;
+        safearraydata_t& operator=(safearraydata_t&&) = default;
+        safearraydata_t(const safearraydata_t&) = delete;
+        safearraydata_t& operator=(const safearraydata_t&) = delete;
```
dunhor, line 250:
```suggestion
        safearraydata_t() = default;
```
Just to be clearer with my suggestion above

include/wil/safearrays.h
```diff
@@ -0,0 +1,1105 @@
...
+    //!     RETURN_IF_FAILED(data.access(psa));
+    //!     result.reserve(result.size() + data.size());
+    //!     for (BSTR& bstr : data)
+    //!     {
+    //!         result.emplace_back(wil::make_bstr_nothrow(bstr));
+    //!         RETURN_HR_IF_NULL(E_OUTOFMEMORY, result.rbegin()->get());
+    //!     }
+    //!     return S_OK;
+    //! }
+    //! ~~~~
```
dunhor, line 177: Does Doxygen handle this correctly? I'd think you'd need another set of `~~~` indicating the start of a new example

include/wil/safearrays.h
```diff
@@ -0,0 +1,1105 @@
...
+    class safearraydata_t
+    {
+    public:
+        typedef typename err_policy::result result;
+        typedef safearray_unaccess_scope_exit::pointer pointer;
+        typedef T value_type;
+        typedef value_type* value_pointer;
+        typedef const value_type* const_value_pointer;
+        typedef value_type& value_reference;
+        typedef const value_type& const_value_reference;
```
dunhor, line 234:
```suggestion
        typedef T value_type;
        typedef value_type* pointer;
        typedef const value_type* const_pointer;
        typedef value_type& reference;
        typedef const value_type& const_reference;
```
Am I missing something? Is there a reason to deviate from the typical naming conventions used in the STL and other libraries? You can replace the uses of `pointer` below with just `SAFEARRAY*` - you shouldn't need a typedef for it (nor do you really need to reference its type in such an obscure way via `safearray_unaccess_scope_exit::pointer`)

include/wil/safearrays.h
```diff
@@ -0,0 +1,1105 @@
...
+        using is_pointer_type = wistd::disjunction<wistd::is_same<T, BSTR>,
+                                                        wistd::is_same<T, LPUNKNOWN>,
+                                                        wistd::is_same<T, LPDISPATCH>>;
+
+        template<typename T>
+        using is_type_not_set = wistd::is_same<T, void>;
+        template<typename T>
+        using is_type_set = wistd::negation<is_type_not_set<T>>;
+        template <typename T>
+        using is_value_type = wistd::conjunction<wistd::negation<is_pointer_type<T>>, is_type_set<T>>;
```
dunhor, line 358: Can these just be `static constexpr bool`s? That would be much simpler to read

include/wil/safearrays.h
```diff
@@ -0,0 +1,1105 @@
...
+        typedef typename storage_t::pointer pointer;
+        typedef element_t elemtype;
+
+    public:
+        // forward all base class constructors...
+        template <typename... args_t>
+        explicit safearray_t(args_t&&... args) WI_NOEXCEPT : storage_t(wistd::forward<args_t>(args)...) {}
+
+        // Exception-based construction
+        template<typename T = element_t, typename wistd::enable_if<is_type_not_set<T>::value, int>::type = 0>
```
dunhor, line 377:
```suggestion
        template<typename T = element_t, wistd::enable_if_t<is_type_not_set<T>::value, int> = 0>
```
Use the alias template throughout. Less text and much simpler to read

magol:
@dwcullop Any progress to this?

PR #148, opened 2020-09-21
Fix C++20 Intellisense errors

sylveon:
module is now a keyword, so without these changes we get Intellisense errors (and possibly down the line compiler errors):

image

### DIFF
--- a/include/wil/win32_helpers.h
+++ b/include/wil/win32_helpers.h
@@ -297,7 +297,7 @@ namespace wil
     /** Retrieves the fully qualified path for the file containing the specified module loaded
     by a given process. Note GetModuleFileNameExW is a macro.*/
     template <typename string_type, size_t initialBufferLength = 128>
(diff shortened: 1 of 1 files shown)


dunhor:
This is an intellisense bug. `module` is _conditionally_ treated as a keyword. E.g. see https://godbolt.org/z/PhKePE

sylveon:
Oops

dunhor:
If you haven't already, be sure to submit an issue at https://developercommunity.visualstudio.com. I'm sure we aren't the only ones using `module` as an identifier :)

rysavyjan:
@sylveon, did you report this problem to https://developercommunity.visualstudio.com?

I'm unable to find it there and we would like to resolve this problem. For now we are renaming module parameter to hModule.

sylveon:
I totally forgot doing it! If you give me a bit I'll do that

sylveon:
Filled feedback: https://developercommunity2.visualstudio.com/t/intellisense-error-when-using-module-as-variable-n/1332749

rysavyjan:
Thank you, upvoted.

PR #149, opened 2020-09-21
Add coroutine-based async IO helpers

sylveon:
This still is missing unit test coverage and documentation, but I'm opening this as a draft to collect comments/feedback before working on that.

This adds a new header, `async.h`, which includes methods that help C++20 consumers do asynchronous IO.

Here's a small example snippet:

```cpp
#include <wil/async.h>
#include <winrt/Windows.Foundation.h>
#include <fmt/format.h>

winrt::Windows::Foundation::IAsyncAction ExampleMethod(HANDLE file)
{
    fmt::memory_buffer buf;
    for (std::uint_fast32_t i = 0; i < 1000001; ++i)
    {
        buf.clear();
        fmt::format_to(buf, "{}\n", i);
        co_await wil::WriteFileAsync(file, buf.data(), static_cast<DWORD>(buf.size()), nullptr);
    }
}

int main()
{
    wil::unique_hfile file(CreateFile(L"aaaaa.txt", GENERIC_WRITE, 0, nullptr, OPEN_ALWAYS, FILE_FLAG_OVERLAPPED, nullptr));
    if (file)
    {
        const auto action = ExampleMethod(file.get());

        while (!action.Completed())
        {
            MsgWaitForMultipleObjectsEx(0, nullptr, INFINITE, QS_ALLINPUT, MWMO_ALERTABLE | MWMO_INPUTAVAILABLE);
        }
    }
}
```

### DIFF
--- a/include/wil/async.h
+++ b/include/wil/async.h
@@ -1,0 +1,139 @@
+//*********************************************************
+//
+//    Copyright (c) Microsoft. All rights reserved.
+//    This code is licensed under the MIT License.
+//    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF
+//    ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+//    TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+//    PARTICULAR PURPOSE AND NONINFRINGEMENT.
+//
+//*********************************************************
+#ifndef __WIL_ASYNC_INCLUDED
+#define __WIL_ASYNC_INCLUDED
+
+#include <version>
+#ifndef __cpp_lib_coroutine
+#error This header requires C++20 coroutines.
+#endif
+
+#include <coroutine>
+#include <Windows.h>
+#include "resource.h"
+#include "result.h"
+
+namespace wil
+{
+    namespace details
+    {
+        struct operation_info
+        {
+            OVERLAPPED overlapped = { };
+
+            bool completed = false;
+            HRESULT error = S_OK;
+
+            std::coroutine_handle<> coro;
+        };
+
+        template<typename err_policy>
+        struct async_file_operation
+        {
+            unique_process_heap_ptr<operation_info> info = nullptr;
+            HRESULT setup_error = S_OK;
+
+            bool await_ready() const noexcept
+            {
+                return FAILED(setup_error) || info->completed;
+            }
+
+            typename err_policy::result await_resume() const
+            {
+                if (FAILED(setup_error))
+                {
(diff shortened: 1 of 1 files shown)


include/wil/async.h
```diff
@@ -0,0 +1,139 @@
...
+        {
+            // this is not used by ReadFileEx so we'll use it.
+            op.info->overlapped.hEvent = bytes_read;
+
+            ULARGE_INTEGER offset_large;
+            offset_large.QuadPart = offset;
+            op.info->overlapped.Offset = offset_large.LowPart;
+            op.info->overlapped.OffsetHigh = offset_large.HighPart;
+
+            if (::ReadFileEx(handle, buffer, bytes_to_read, &op.info->overlapped, details::OverlappedCallback))
```
smaillet-ms, line 100: While I'm all for the idea of this PR, there's a problem here in that it only works if, like in your sample, the caller making the async request actually blocks in an alertable state. The callback for the completion ONLY happens ON the thread that made the original call to ReadFileEx AND only while in a alertable state. That's a restriction with a lot of surprise attached to it. If the call stack doing the co_await never blocks in an alertable state (Surprisingly common as not all synch objects support this and the default is not alertable for most that do)

I'm not sure if documentation alone is enough to resolve that and if there could be subtle issues that complicate things. That is, in local dev loop testing everything completes synchronously, but in the field it doesn't subtle behavior differences could cause difficult to track bugs. That said, though it wouldn't actually be worse than what you can get into using the APIs raw...

sylveon: I could rename the functions to `wil::ReadFileEx` and `wil::WriteFileEx`, that would make it clearer that the alertable/threading behavior matches the native APIs with the same name. I am not very familiar with IOCPs and there are even more ways to do async IO (like `RegisterWaitForSingleObject`).

As a note to myself, this is missing some `GetOverlappedResult` calls.

smaillet-ms: Rename would help clarify this particular case. Though, I'm not sure that wrapping ReadFileEx/WriteFileEx is really the way to go for Async IO in WIL. As you say there are a lot of ways to do ASYNC IO in Win32, each having advantages. IOCompletion ports are another option. There is also the Thread Pool IO support APIs that actually use IO Completion ports under the hood but hide that in a a way that likely fits the coroutine co_await style paradigm better. (see: [CreateThreadPoolIo](https://docs.microsoft.com/en-us/windows/win32/api/threadpoolapiset/nf-threadpoolapiset-createthreadpoolio) and related APIs)

sylveon:
Closing this because I abandonned the original project that lead to this implementation. Feel free to pick it up.

PR #153, opened 2020-10-08
Add support for alertable waits on wil::unique_event

### LINKED ISSUE #152: wil event RAII types don't support alertable waits
wil::event_t relies on the wil::handle_wait function which uses WaitForSingleObjectEx with a hardcoded alertable flags set to FALSE, this prevents the use of wait for async I/O scenarios. Forcing apps to "get" the raw hjandle and call WaitForSIngleObjectEx directly. This is inconsistent with the mutex, semaphore and other objects that do have the alertable flag exposed.

Possible solution:
Add the alretable flag with a default of FALSE to wil::handle_wait() AND the event_t::wait() so that existing apps get the same functionality they always got, but new code can specify the wait.

sylveon:
Fixes #152

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -2561,10 +2561,10 @@ namespace wil
     }
 
     // Waits on the given handle for the specified duration
-    inline bool handle_wait(HANDLE hEvent, DWORD dwMilliseconds = INFINITE) WI_NOEXCEPT
+    inline bool handle_wait(HANDLE hEvent, DWORD dwMilliseconds = INFINITE, BOOL bAlertable = FALSE) WI_NOEXCEPT
     {
         DWORD status = ::WaitForSingleObjectEx(hEvent, dwMilliseconds, FALSE);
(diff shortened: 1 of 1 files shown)


smaillet-ms (approved):

smaillet-ms (approved):

smaillet-ms:
I Approved 👍, but I don't think my vote matters much in this repo 😁

sylveon:
Certainly matters more than mine would, although we both missed I wasn't passing `bAlertable` originally 😂

smaillet-ms:
Yeah, embarrassed I missed that one! Though, I also missed that there's no tests associated with this change either. Not sure what the process is for testing things in the GitHub version of WIL. (I'm more used to the internal version)

sylveon:
I couldn't find any test covering alertable waits on mutex or semaphores either, and am unfamiliar with how to proceed with testing such a thing

dunhor (approved):

PR #166, opened 2020-12-20
Holiday omnibus issue-addressing

### LINKED ISSUE #38: PathCch.h include in win32_helpers.h
PathCch.h is included in win32_helpers.h:
https://github.com/microsoft/wil/blob/master/include/wil/win32_helpers.h#L18

However, the functions declared in this header are only available on Windows 8 or higher. An include guard could be used to cover this case:
```
#if (_WIN32_WINNT >= _WIN32_WINNT_WIN8)
#include <PathCch.h>
#endif
```
But after quickly inspecting win32_helpers.h, I couldn't find any references to the symbols in PathCch.h and it compiles without any errors in my project without this include.

If the PathCch.h include has no function, I propose it should be removed.

### LINKED ISSUE #44: wil/winrt.h requires time.h to be included to use it
Raymond pointed this out here: https://github.com/microsoft/wil/wiki/Windows-Runtime-helpers

this should be fixed.

### LINKED ISSUE #154: SAL annotations for wil::mapview_deleter::operator() are incorrect
The current implementation generates an error when running code analysis on code that uses wil::unique_mapview_ptr<T>.

In particular the annotations on mapview_deleter::operator() are incorrect it currently has (_Pre_opt_valid _Frees_ptr_opt) however that doesn't match the requirements and declaration of UnMapViewOfFile, which doesn't accept NULL (e.g. the "opt" portion of the annotations is erroneous))

FYI the error is:
`packages\Microsoft.Windows.ImplementationLibrary.1.0.200902.2\include\wil\resource.h(3456): warning C6387: 'p' could be '0':  this does not adhere to the specification for the function 'UnmapViewOfFile'. `

jonwis:
The implementation of `wistd::unique_ptr` ensures that an invalid value (like `nullptr`) is never passed to the `_deleter` methods, here:

https://github.com/microsoft/wil/blob/adae42be8f45bdbf0fa50aca000fd8bf2063fff5/include/wil/wistd_memory.h#L568-L573

The `_opt_` annotations are irrelevant and can be removed, which fixes #154 

* Also adds a compilation test for `wil::unique_mapview_ptr<>` so untyped mapped views can be held.
* Removes a spurious inclusion of `pathcch.h` in `win32_helpers.h` which is no longer needed (fixes #38)
* Add an inclusion of `windows.foundation.h` to `winrt.h` to get `Windows::Foundation::DateTime` defined (fixes #44)

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -1298,10 +1298,7 @@ namespace wil
     {
         template<typename T> void operator()(_Frees_ptr_opt_ T* toFree) const
         {
-            if (toFree)
-            {
-                TDeleter(toFree);
-            }
+            TDeleter(toFree);
         }
     };
 
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -3438,7 +3438,7 @@ namespace wil
     struct process_heap_deleter
     {
         template <typename T>
-        void operator()(_Pre_opt_valid_ _Frees_ptr_opt_ T* p) const
+        void operator()(_Pre_valid_ _Frees_ptr_ T* p) const
         {
             details::FreeProcessHeap(p);
         }
@@ -3447,7 +3447,7 @@ namespace wil
     struct virtualalloc_deleter
     {
         template<typename T>
-        void operator()(_Pre_opt_valid_ _Frees_ptr_opt_ T* p) const
+        void operator()(_Pre_valid_ _Frees_ptr_ T* p) const
         {
             ::VirtualFree(p, 0, MEM_RELEASE);
         }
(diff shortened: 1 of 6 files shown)


include/wil/resource.h
```diff
@@ -1298,7 +1298,10 @@ namespace wil
     {
         template<typename T> void operator()(_Frees_ptr_opt_ T* toFree) const
         {
-            TDeleter(toFree);
+            if (toFree)
```
ChrisGuzak, line 1301: many deleter functions, like `LocalFree` accept `nullptr`, so this adds an unneeded test, growing the amount of code generated. should we differentiate deleters that do or don't accept null so we can optimize the `LocalFree` case?

jonwis: I think you're right. Everybody using the standard deleters must have this problem fixed already (in that the various `function_deleter` specializations use methods that are OK with null-ness.  The `mapview_deleter` annotation was just wrong and has been fixed while this method is left alone.

jonwis:
See also https://github.com/MicrosoftDocs/sdk-api/pull/610 which updates the documentation for `HeapFree` since it's _not_ in fact undefined behavior to pass `nullptr`.

ChrisGuzak (approved):

PR #167, opened 2020-12-23
Add noexcept on non-throwing CoCreateInstance methods

sylveon:
This prevents a superfluous IDE warning when using them in a `noexcept` function:
image


jonwis (approved):
Thanks! Are there other `...NoThrow` or `...FailFast` methods that need the attribution as well?

sylveon:
I can find a few that seem to be missing from a quick search but trying to find them all and validate that noexcept is actually safe to put is a bit tedious. Probably safer to do that as they crop up?

PR #168, opened 2020-12-24
Use std::bit_cast when available and add more noexcept in win32_helpers.h

sylveon:
Sort of a followup to #167!

I encountered the same issue with `GetModuleInstanceHandle` and decided to do that whole header since it's fairly short.

As a bonus, I enabled constexpr and totally UB-free `wil::filetime::from_int64` for consumers which use C++20.

However, I find it rather weird that `to_int64` uses bit shifts due to alignment differences, while right below `from_int64` asserts that the alignment is the same and does a `reinterpret_cast`. Is there any particular reason for this difference? Can I make both use bit shifts to make the API constexpr all the time?

### DIFF
--- a/include/wil/win32_helpers.h
+++ b/include/wil/win32_helpers.h
@@ -19,18 +19,19 @@
 #include <objbase.h>
 
 // detect std::bit_cast
-// should replace with __WI_LIBCPP_CONSTEXPR_AFTER_CXX20 once compilers
-// correctly signal C++20 in __cplusplus
 #ifdef __has_include
 #  if __has_include(<version>)
 #    include <version>
-#    ifdef __cpp_lib_bit_cast
+#    if __cpp_lib_bit_cast >= 201806L
 #      include <bit>
-#      define WI_CONSTEXPR_BIT_CAST __WI_LIBCPP_CONSTEXPR
-#    else
-#      define WI_CONSTEXPR_BIT_CAST inline
 #    endif
 #  endif
+#endif
+
+#if __cpp_lib_bit_cast >= 201806L
+#  define __WI_CONSTEXPR_BIT_CAST constexpr
+#else
+#  define __WI_CONSTEXPR_BIT_CAST inline
 #endif
 
 #include "result.h"
@@ -73,7 +74,7 @@ namespace wil
     {
         constexpr unsigned long long to_int64(const FILETIME &ft) WI_NOEXCEPT
         {
-#ifdef __cpp_lib_bit_cast
+#if __cpp_lib_bit_cast >= 201806L
             return std::bit_cast<unsigned long long>(ft);
 #else
             // Cannot reinterpret_cast FILETIME* to unsigned long long*
@@ -82,9 +83,9 @@ namespace wil
 #endif
         }
 
-        WI_CONSTEXPR_BIT_CAST FILETIME from_int64(unsigned long long i64) WI_NOEXCEPT
+        __WI_CONSTEXPR_BIT_CAST FILETIME from_int64(unsigned long long i64) WI_NOEXCEPT
         {
-#ifdef __cpp_lib_bit_cast
+#if __cpp_lib_bit_cast >= 201806L
             return std::bit_cast<FILETIME>(i64);
 #else
             static_assert(sizeof(i64) == sizeof(FILETIME), "sizes don't match");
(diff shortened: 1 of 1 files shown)


ChrisGuzak:
> Is there any particular reason for this difference?

not that I'm aware of. I recall @oldnewthing was involved with the cast to shift change. and I assume that was done consistently. perhaps we had a bad merge.

oldnewthing:
The original code used casts for both conversions. However, that resulted in alignment faults in `to_int64` because `FILETIME` is 4-byte aligned but `int64` is 8-byte aligned. The cast in `from_int64` is still valid because we are casting from strict to weak, which is allowed. I changed only what was necessary at the time, namely, fixing the alignment faults. If changing the `from_int64` from casting to shifting has benefits, then go for it.

ChrisGuzak (approved):

sylveon:
Looking at some compiler output, it seems that `bit_cast` and `reinterpret_cast` both have the same codegen on x64, while a bitshift results in worse assembly. On x86, everything is the same however.

https://godbolt.org/z/6vxGqo

So keeping the `reinterpret_cast` is fine, then.

For `to_int`, `memcpy` reaches the same assembly as `bit_cast` while the bitshift is a bit worse, so there is gain to be made by using `memcpy` (however it would require making it non-`constexpr`, which has the slight change of breaking consumers).

https://godbolt.org/z/qM57Y7

include/wil/win32_helpers.h
```diff
@@ -18,6 +18,21 @@
 #include <PathCch.h>
 #include <objbase.h>
 
+// detect std::bit_cast
+// should replace with __WI_LIBCPP_CONSTEXPR_AFTER_CXX20 once compilers
```
dunhor, line 22: In general, feature-test macros are meant to replace most C++ version macros

include/wil/win32_helpers.h
```diff
@@ -18,6 +18,21 @@
...
 
+// detect std::bit_cast
+// should replace with __WI_LIBCPP_CONSTEXPR_AFTER_CXX20 once compilers
+// correctly signal C++20 in __cplusplus
+#ifdef __has_include
+#  if __has_include(<version>)
+#    include <version>
+#    ifdef __cpp_lib_bit_cast
+#      include <bit>
+#      define WI_CONSTEXPR_BIT_CAST __WI_LIBCPP_CONSTEXPR
```
dunhor, line 29: In general, names with the `WI_` prefix are intended to be consumable outside of WIL. The prefix `__WI_` - disregarding the fact that this violates the standard, we know - is typically used for internal macros.

That said, since the the implementation of the function dramatically changes based off feature presence, this definition doesn't really seem necessary. See later comment

include/wil/win32_helpers.h
```diff
@@ -56,45 +71,53 @@ namespace wil
...
+            return std::bit_cast<unsigned long long>(ft);
+#else
             // Cannot reinterpret_cast FILETIME* to unsigned long long*
             // due to alignment differences.
             return (static_cast<unsigned long long>(ft.dwHighDateTime) << 32) + ft.dwLowDateTime;
+#endif
         }
 
-        inline FILETIME from_int64(unsigned long long i64)
+        WI_CONSTEXPR_BIT_CAST FILETIME from_int64(unsigned long long i64) WI_NOEXCEPT
```
dunhor, line 85: Seems like this could pretty easily be split into two definitions to avoid the need for the macro and keep things relatively clean. e.g.
```c++
#if __cpp_lib_bit_cast >= 201806L
constexpr FILETIME from_int64(unsigned long long i64) WI_NOEXCEPT
{
    // New definition
}
#else
inline FILETIME from_int64(unsigned long long i64) WI_NOEXCEPT
{
    // Old definition
}
#endif
```

include/wil/win32_helpers.h
dunhor: There's also the issue that this macro goes undefined if `__has_include` is not defined or `<version>` is unavailable.

include/wil/win32_helpers.h
sylveon: The `add` function right below this one could be constexpr if the macro is kept.

include/wil/win32_helpers.h
```diff
@@ -18,6 +18,22 @@
 #include <PathCch.h>
 #include <objbase.h>
 
+// detect std::bit_cast
+#ifdef __has_include
+#  if __has_include(<version>)
```
dunhor, line 23: Mostly nitpicking at this point, but this could be simplified to just `#if __has_include(<bit>)` since that header (also) defines the feature test macro

dunhor (approved):

include/wil/win32_helpers.h
sylveon: I was thinking in case there is standard library with a `<bit>` which doesn't have `std::bit_cast` but supports other functions from that header, we would avoid bringing it in entirely only to not use it.

PR #171, opened 2021-01-05
[installation] Fix install headers

JackBoosY:
Since case is sensitive in linux, and `PROJECT_NAME` is uppercase `WIL`:
https://github.com/microsoft/wil/blob/3b559727eceb622320a5a3540b8280fa2e1cb4ee/CMakeLists.txt#L2
And the header file directory name is lowercase `wil`:
_https://github.com/microsoft/wil/tree/master/include/wil_

This will cause a folder not found error when installing the header file directory in linux:
https://github.com/microsoft/wil/blob/3b559727eceb622320a5a3540b8280fa2e1cb4ee/CMakeLists.txt#L51

Fixes this.

Related: https://github.com/microsoft/vcpkg/pull/15438

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,4 +48,4 @@ install(EXPORT ${PROJECT_NAME}_targets
 )
 
 # Install the headers at a standard cmake location.
-install(DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/include/${PROJECT_NAME}" DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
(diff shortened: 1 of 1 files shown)


ChrisGuzak (commented):
what about the other uses of ${`PROJECT_NAME`}? why do they remain using the macro instead of just converting all to `wil`?

I'm not CMake savvy, maybe this is obvious to those that are.

JackBoosY:
@ChrisGuzak If you use `PROJECT_NAME` as the target name, there is no problem using it in other parts of the CMakeLists.
The key to this issue is that it involves special rules in unix.

ChrisGuzak (approved):

PR #173, opened 2021-01-23
wil/cppwinrt.h should consume winrt::winrt_throw_hresult_handler to log cppwinrt exceptions that are thrown

### LINKED ISSUE #172: wil/cppwinrt.h should consume winrt::winrt_throw_hresult_handler to log cppwinrt exceptions that are thrown
There is currently a gap when combining cppwinrt usage with WIL and it's logging functionality. The gap is that cppwinrt exceptions that come out of winrt::throw_hresult are invisible to WIL. This is problematic in a few ways so it would be best to fill this gap so that these libraries interop better.

To achieve that an extension point is being added to winrt::throw_hresult. This extension point will be nullptr in normal usage. When <wil/cppwinrt.h> is included it should wire the extension point with a function pointer that can take the exception information and log it appropriately.

See [cppwinrt PR 843](https://github.com/microsoft/cppwinrt/pull/843) for the corresponding change.

dmachaj:
Fixes #172

This change is being made to improve the interop between cppwinrt and WIL (particularly WIL's logging abilities for errors). It consumes a new hook from the `winrt::throw_hresult` method via `<wil/cppwinrt.h>` by filling it in with a function pointer. That function pointer will allow WIL's logging abilities to gain knowledge of the exceptions that are originated within cppwinrt.  When both changes are combined WIL is able to observe and log cppwinrt exceptions that previously flew past invisibly.

The core of this change is the new `winrt_throw_hresult` method which will be called from `winrt::throw_hresult` when things are set up properly.  This method forwards along the error information in a way roughly equivalent to `LOG_HR`.  The contract allows for the usual set of information (file name, function name, line number, error code, return address).  At this time only the return address and error code are actually able to be provided by cppwinrt.  Hopefully that changes in the future with [C++20 source_location](https://en.cppreference.com/w/cpp/utility/source_location).

The other piece of this is hooking up the new extension point when the cppwinrt version is new enough.  This is done via an `if constexpr` check in `WilInitialize_CppWinRT`.  If the version is new enough then it is hooked up.  If the version is too old then it would be a linker error and it compiles away into nothingness.

The trickiest part was refactoring the `constexpr` version checking method to expose both the major and minor versions instead of just the major version.  The minor version is actually the 3rd number in the versioning scheme so this isn't the prettiest method.  Unfortunately all of the usual STL string helpers are unavailable from a `constexpr` method so this must be done manually.

## How was the change tested?
This change was tested in a cppwinrt blank project that calls `winrt::check_hresult` with a WIL logger active.  Previous to both of these changes it never saw the error.  I pulled the latest NuGet from cppwinrt with that portion of this change and rebuilt.  Same result.  I then made the <wil/cppwinrt.h> changes and got that into the blank project and confirmed that it is properly logging the error as expected.  I also ran a `ninja` build and `runtests.cmd`.

### DIFF
--- a/include/wil/cppwinrt.h
+++ b/include/wil/cppwinrt.h
@@ -28,7 +28,8 @@
 namespace wil::details
 {
     // Since the C++/WinRT version macro is a string...
-    inline constexpr int major_version_from_string(const char* versionString)
+    // For example: "2.0.210122.3"
+    inline constexpr int version_from_string(const char* versionString)
     {
         int result = 0;
         auto str = versionString;
@@ -39,6 +40,34 @@ namespace wil::details
         }
 
         return result;
+    }
+
+    // Since the C++/WinRT version macro is a string...
+    // For example: "2.0.210122.3"
+    inline constexpr int major_version_from_string(const char* versionString)
+    {
+        return version_from_string(versionString);
+    }
+
+    inline constexpr int minor_version_from_string(const char* versionString)
+    {
+        auto str = versionString;
+        int dotCount = 0;
+        while ((*str != '\0'))
+        {
+            if (*str == '.')
+            {
+                ++dotCount;
+            }
+
+            ++str;
+            if (dotCount == 2)
+            {
+                break;
+            }
+        }
+
+        return version_from_string(str);
     }
 }
 /// @endcond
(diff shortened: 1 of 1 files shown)


kennykerr (approved):
Looks good to me, but I'd wait for Duncan to sign off on the WIL side of things.

dunhor (approved):

PR #174, opened 2021-02-01
Sync changes with OS repo

dunhor:


dunhor:
For external reference, there are instances of files named `version` - that are not headers - that are in some include paths, hence the change there.

PR #178, opened 2021-03-16
Check for ERROR_ALREADY_EXISTS for the second CreateDirectoryW call.

jlaanstra:
Handle the case where multiple threads are trying to create the same directory and the second call to CreateDirectoryW can also return ERROR_ALREADY_EXISTS.

### DIFF
--- a/include/wil/filesystem.h
+++ b/include/wil/filesystem.h
@@ -113,7 +113,7 @@ namespace wil
     {
         if (::CreateDirectoryW(path, nullptr) == FALSE)
         {
-            DWORD const lastError = ::GetLastError();
+            DWORD lastError = ::GetLastError();
             if (lastError == ERROR_PATH_NOT_FOUND)
             {
                 size_t parentLength;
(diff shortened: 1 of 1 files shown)


include/wil/filesystem.h
```diff
@@ -122,9 +122,16 @@ namespace wil
                     wistd::unique_ptr<wchar_t[]> parent(new (std::nothrow) wchar_t[parentLength + 1]);
                     RETURN_IF_NULL_ALLOC(parent.get());
                     RETURN_IF_FAILED(StringCchCopyNW(parent.get(), parentLength + 1, path, parentLength));
-                    CreateDirectoryDeepNoThrow(parent.get()); // recurs
+                    RETURN_IF_FAILED(CreateDirectoryDeepNoThrow(parent.get())); // recurs
+                }
+                if (::CreateDirectoryW(path, nullptr) == FALSE)
```
ChrisGuzak, line 127: consider a comment
`// handle racing threads both creating the same directory`

ChrisGuzak (approved):

PR #179, opened 2021-03-29
add wil::WaitForDebuggerPresent

ChrisGuzak:
This makes common debugging work flows simpler. Put a call to this function in your process init code and now the IFEO configuration can be set to cause the process to wait for the debugger to attach.

```
    // Waits for a debugger to attach to the current process based on registry configuration.
    //
    // Example:
    //     HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\explorer.exe
    //         WaitForDebuggerPresent=1
    //
    // REG_DWORD value of
    //     missing or 0 -> don't break
    //     1 -> wait for the debugger, continue execution once it is attached
    //     2 -> wait for the debugger, break here once attached.
```

also
- make `wil::GetModuleFileNameW` param optional, `nullptr` means current process
- add exception based version of `wil::QueryFullProcessImageNameW`

Both of these APIs can be used to get the current process file path but in my test `wil::GetModuleFileNameW(nullptr)` was about 10x faster, ~4ms for 1000 iterations vs ~60ms.

### DIFF
--- a/include/wil/win32_helpers.h
+++ b/include/wil/win32_helpers.h
@@ -496,7 +496,7 @@ namespace wil
             {
                 if (configValue == 2)
                 {
-                    __debugbreak(); // debugger attached, SHIFT+F11 to return to the caller
+                    DebugBreak(); // debugger attached, SHIFT+F11 to return to the caller
                 }
                 return; // debugger now attached, continue executing
             }
--- a/include/wil/win32_helpers.h
+++ b/include/wil/win32_helpers.h
@@ -437,6 +437,14 @@ namespace wil
         return result;
     }
 
+    template <typename string_type = wil::unique_cotaskmem_string, size_t stackBufferLength = 256>
+    string_type QueryFullProcessImageNameW(HANDLE processHandle, DWORD flags = 0)
+    {
+        string_type result;
+        THROW_IF_FAILED((wil::QueryFullProcessImageNameW<string_type, stackBufferLength>(processHandle, flags, result)));
+        return result;
+    }
+
 #endif
 
     /** Retrieve the HINSTANCE for the current DLL or EXE using this symbol that
--- a/include/wil/win32_helpers.h
+++ b/include/wil/win32_helpers.h
@@ -422,7 +422,7 @@ namespace wil
     }
 
     template <typename string_type = wil::unique_cotaskmem_string, size_t initialBufferLength = 128>
-    string_type GetModuleFileNameW(HMODULE module)
+    string_type GetModuleFileNameW(HMODULE module = nullptr)
     {
         string_type result;
         THROW_IF_FAILED((wil::GetModuleFileNameW<string_type, initialBufferLength>(module, result)));
--- a/include/wil/win32_helpers.h
+++ b/include/wil/win32_helpers.h
@@ -15,6 +15,7 @@
 #include <sysinfoapi.h> // GetSystemTimeAsFileTime
 #include <libloaderapi.h> // GetProcAddress
 #include <Psapi.h> // GetModuleFileNameExW (macro), K32GetModuleFileNameExW
+#include <winreg.h>
 #include <objbase.h>
 
 // detect std::bit_cast
@@ -422,7 +423,7 @@ namespace wil
     }
 
     template <typename string_type = wil::unique_cotaskmem_string, size_t initialBufferLength = 128>
-    string_type GetModuleFileNameW(HMODULE module = nullptr)
+    string_type GetModuleFileNameW(HMODULE module = nullptr /* current process module */)
     {
         string_type result;
         THROW_IF_FAILED((wil::GetModuleFileNameW<string_type, initialBufferLength>(module, result)));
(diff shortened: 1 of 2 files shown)


include/wil/win32_helpers.h
```diff
@@ -437,6 +438,73 @@ namespace wil
...
+    {
+        auto filePath = wil::GetModuleFileNameW<wil::unique_cotaskmem_string>();
+        if (auto lastSlash = wcsrchr(filePath.get(), L'\\'))
+        {
+            const auto fileName = lastSlash + 1;
+            auto keyPath = wil::str_concat<wil::unique_cotaskmem_string>(LR"(SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\)",
+                fileName);
+            DWORD value{}, sizeofValue = sizeof(value);
+            if (::RegGetValueW(HKEY_LOCAL_MACHINE, keyPath.get(), valueName,
+#ifdef RRF_SUBKEY_WOW6464KEY
```
jonwis, line 462: This has been defined since _WIN32_WINNT_WINTHRESHOLD ... Do we have a stance on minimum supported builds for wil?

include/wil/win32_helpers.h
```diff
@@ -437,6 +438,73 @@ namespace wil
...
+            if (configValue == 0)
+            {
+                return; // not configured, don't wait
+            }
+
+            if (IsDebuggerPresent())
+            {
+                if (configValue == 2)
+                {
+                    __debugbreak(); // debugger attached, SHIFT+F11 to return to the caller
```
jonwis, line 499: Should this be `::DebugBreak()`, which is compatible with clang?

include/wil/win32_helpers.h
```diff
@@ -437,6 +438,73 @@ namespace wil
...
+    inline DWORD GetCurrentProcessExecutionOption(PCWSTR valueName, DWORD defaultValue = 0)
+    {
+        auto filePath = wil::GetModuleFileNameW<wil::unique_cotaskmem_string>();
+        if (auto lastSlash = wcsrchr(filePath.get(), L'\\'))
+        {
+            const auto fileName = lastSlash + 1;
+            auto keyPath = wil::str_concat<wil::unique_cotaskmem_string>(LR"(SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\)",
+                fileName);
+            DWORD value{}, sizeofValue = sizeof(value);
+            if (::RegGetValueW(HKEY_LOCAL_MACHINE, keyPath.get(), valueName,
```
jonwis, line 461: Should this fall back from HKCU to HKLM? Changing a systemwide setting isn't always possible or good. IIRC standard users can debug their _own_ processes, and only need elevation to debug somebody else's.

jonwis (approved):

include/wil/win32_helpers.h
```diff
@@ -437,6 +438,73 @@ namespace wil
...
+    //
+    // Example:
+    //     HKLM\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\explorer.exe
+    //         WaitForDebuggerPresent=1
+    //
+    // REG_DWORD value of
+    //     missing or 0 -> don't break
+    //     1 -> wait for the debugger, continue execution once it is attached
+    //     2 -> wait for the debugger, break here once attached.
+    inline void WaitForDebuggerPresent(bool checkRegistryConfig = true)
```
oldnewthing, line 485: Mysterious boolean parameter. Somebody who reads along and sees

```cpp
wil::WaitForDebuggerPresent(false);
```

is not going to immediately understand what `false` means. As written, it sounds like it means "Don't wait for the debugger." You're pretty much forcing people to go back and read the implementation to figure out what that `bool` does.

Suggest two functions:

* `wil::WaitForDebuggerPresent()` - waits unconditionally
* `wil::WaitForDebuggerPresentIfConfigured` - waits if configured in HKLM

include/wil/win32_helpers.h
ChrisGuzak: the test made me do this. they compile targeting Win7 and this would not compile without the conditional.

include/wil/win32_helpers.h
```diff
@@ -422,7 +423,7 @@ namespace wil
     }
 
     template <typename string_type = wil::unique_cotaskmem_string, size_t initialBufferLength = 128>
-    string_type GetModuleFileNameW(HMODULE module)
+    string_type GetModuleFileNameW(HMODULE module = nullptr /* current process module */)
```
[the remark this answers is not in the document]
ChrisGuzak: I don't want to take that on now. but would be open to it. in my experience I've never see the case were a non-null module handle is used. if that correctly represents usage then this problem would be a rare one.

include/wil/win32_helpers.h
```diff
@@ -437,6 +438,73 @@ namespace wil
         return result;
     }
 
+    template <typename string_type = wil::unique_cotaskmem_string, size_t stackBufferLength = 256>
+    string_type QueryFullProcessImageNameW(HANDLE processHandle = GetCurrentProcess(), DWORD flags = 0)
```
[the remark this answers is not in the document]
ChrisGuzak: nullptr produces E_HANDLE (GetCurrentProcess() does not return null), so that bug is not exposed here. I'm going to keep this.

include/wil/win32_helpers.h
ChrisGuzak: `wil::WaitForDebuggerPresent(false /* checkRegistryConfig */);` is how I'd write that case. mitigating the confusion.

include/wil/win32_helpers.h
ChrisGuzak: I recalled that DebugBreak() would leave you a frame deeper, motivating __debugbreak(), but I just tried this and it leaves you on the line after the statement in the VS debugger. so I'll take this one.

oldnewthing (approved):

dunhor (approved):

ysc3839:
The `Image File Execution Options` supports filtering full path (From [here](https://superuser.com/a/1106241)). Although this option is not documented, there is a [document](https://docs.microsoft.com/en-us/microsoft-365/security/defender-endpoint/troubleshoot-exploit-protection-mitigations?view=o365-worldwide) mention it. Should we add support for this?

PR #182, opened 2021-04-20
Bring back "trivial" cleanup calls for most resources

jonwis:
Some experimentation shows that use of `wistd::invoke` (see #43 ) results in a small but measurable size increase in certain DLLs for resources that have trivial C-style cleanup functions. Investigations showed that the inliner was unable to see through the `wistd::invoke(close_fn, value)` call and resolve it to just `close_fn(value)`, resulting in an indirect call.

This change supports both direct method calls and "invoke-ables", getting the best of both worlds. Direct methods are returned to their `close_fn(value)` and invoke-ables continue to use `wistd::invoke` to resolve the calls.

Also added some compilation & runtime tests to verify behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Some things that WIL includes to whet your appetite:
   API HANDLEs, HWNDs, and other resources and resource handles with
   [RAII](https://en.cppreference.com/w/cpp/language/raii) semantics.
 - [`include/wil/win32_helpers.h`](include/wil/win32_helpers.h)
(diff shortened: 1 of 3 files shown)


jonwis:
In working on this, I tried out some of the compiler throughput test flags. Adding this new base adds a fractional amount of compilation time.  I'm considering instead putting this as a 'using' into the type body and returning the 'close' methods.

Generally, compilation throughput of WIL is a concern since we use it _everywhere_ in Windows. Some profiling and analysis work might be in order.

ChrisGuzak (approved):

PR #185, opened 2021-06-03
Add exception macro for gcc and clang

longnguyen2004:
`gcc` and `clang` don't define `_CPPUNWIND`, but they define `__EXCEPTIONS`, so I added a check for it.

### DIFF
--- a/include/wil/common.h
+++ b/include/wil/common.h
@@ -264,7 +264,7 @@ WI_ODR_PRAGMA("WIL_KERNEL_MODE", "0")
 #if (defined(_CPPUNWIND) || defined(__EXCEPTIONS)) && !defined(WIL_SUPPRESS_EXCEPTIONS)
 /** This define is automatically set when exceptions are enabled within wil.
 It is automatically defined when your code is compiled with exceptions enabled (via checking for the built-in
-_CPPUNWIND flag) unless you explicitly define WIL_SUPPRESS_EXCEPTIONS ahead of including your first wil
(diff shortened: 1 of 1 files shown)


include/wil/common.h
```diff
@@ -261,7 +261,7 @@ WI_ODR_PRAGMA("WIL_KERNEL_MODE", "1")
 WI_ODR_PRAGMA("WIL_KERNEL_MODE", "0")
 #endif
 
-#if defined(_CPPUNWIND) && !defined(WIL_SUPPRESS_EXCEPTIONS)
+#if defined(_CPPUNWIND) || defined(__EXCEPTIONS) && !defined(WIL_SUPPRESS_EXCEPTIONS)
```
justanotheranonymoususer, line 264: Parentheses?

sylveon: Yeah, this breaks WIL_SUPPRESS_EXCEPTIONS on MSVC

longnguyen2004: I'll fix it soon.

longnguyen2004:
How is this PR going?

jonwis (approved):

jonwis:
Do the existing tests use clang to compile & validate? Shouldn't they have caught this?

dunhor:
> Do the existing tests use clang to compile & validate? Shouldn't they have caught this?

The tests use Clang in MSVC-compat mode (i.e. 'clang-cl'). My guess is that this is intended to address using clang/gcc in non-compat mode.

dunhor (approved):

PR #188, opened 2021-06-18
Adding COM apartment variable (wil/com_apartment_variable.h)

ChrisGuzak:
Apartment variables are stored in the current COM apartment, every apartment has unique storage. When the apartment runs down the variables are released. This avoids the problems of using COM objects in global variables and enables caching of information relative to the apartment. 

```cpp
wil::apartment_variable<std::wstring> loggerName;

std::wstring GetNewLoggerName() { /* ... */ }

void LogThings() {
    auto& name = loggerName.get_or_create(&GetNewLoggerName);
    log_to_file(name);
    if (log_file_getting_big())
    {
        name = GetNewLoggerName(); // roll over to new log file
    }
}

void UseLogIfPresent() {
   auto name = loggerName.get_if();
   if (name) log_to_file(*name);
}

void StartNewLogger() {
     loggerName.set(GetNewLoggerName());
}

wil::apartment_variable<int> widgetCount;

void UpdateWidgetCounter()
{
    ++widgetCount.get_or_create();
}
```


include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,230 @@
...
+
+            auto observer = winrt::make<ApartmentObserver>();
+            unsigned long long id{};
+            APARTMENT_SHUTDOWN_REGISTRATION_COOKIE cookie{}; // leaked, COM run down releases all instances
+            winrt::check_hresult(RoRegisterForApartmentShutdown(observer.get(), &id, &cookie));
+            FAIL_FAST_IF(apartmentId != id);
+
+            auto result = createFn();
+            details::s_apartmentStorage.insert({ apartmentId,
+                decltype(details::s_apartmentStorage)::value_type::second_type { { variableKey, result } } });
```
oldnewthing, line 108: Dunno if this is clearer or more confusing:

```cpp
                decltype(details::s_apartmentStorage[0]) { { variableKey, result } } });
```

but do you even have to say the type?

```cpp
details::s_apartmentStorage.insert({ apartmentId, { { variableKey, result } } });
```

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,230 @@
...
+    // rundown the variable before the apartment exits.
+    template<typename F>
+    auto get_for_current_com_apartment(const F& createFn) -> std::invoke_result_t<F>
+    {
+        static_assert(std::is_function_v<F>, "createFn must be a function");
+
+        unsigned long long apartmentId{};
+        FAIL_FAST_IF_FAILED(RoGetApartmentIdentifier(&apartmentId));
+
+        const auto variableKey = std::addressof(createFn);
```
oldnewthing, line 59: I'm surprised this compiles. The address of `createFn` is a `R (*)()` which is not implicitly convertible to `void*`.

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,230 @@
...
+        inline static std::unordered_map<unsigned long long, std::unordered_map<const void*, std::any>> s_apartmentStorage;
+    }
+
+    // Get the apartment variable, returning the value stored in the per apartment
+    // storage or use the create function to create and store it in the current apartment.
+    // This variables lifetime is tied to the COM apartment lifetime and will
+    // be safely cleaned up. Call reset_for_current_com_apartment if you want to
+    // rundown the variable before the apartment exits.
+    template<typename F>
+    auto get_for_current_com_apartment(const F& createFn) -> std::invoke_result_t<F>
```
oldnewthing, line 52: Could be simplified to

```cpp
template<typename F, std::enable_if_t<std::is_function_v<F>, int> = 0>
auto get_for_current_com_apartment(F const* createFn) -> decltype(createFn())
{
   ...
}
```

also enforces that `createFn` be callable with zero parameters.

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,230 @@
...
+            auto variable = variables.find(variableKey);
+            assert(variable != variables.end());
+            std::ignore = std::any_cast<std::invoke_result_t<F>>(variable->second);
+#endif
+            return result;
+        }
+    }
+
+    // Remove the variable including its entry in the map. This has no
+    // effect if the variable was not stored based on a call to get_for_current_com_apartment
```
oldnewthing, line 123: I don't see why this restriction is necessary. It means that people are going to have a dummy get before the reset.

```cpp
auto DefaultCurrentWidget() { return std::_weak_ptr<Widget>(); }

void SetCurrentWidgetForThread(std::shared_ptr<Widget> const& widget)
{
    // dummy get_() call to populate the entry so we can reset it
    std::ignore = wil:get_for_current_com_apartment(DefaultCurrentWidget);

    wil::reset_for_current_com_apartment(DefaultCurrentWidget, widget);
}
```

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,230 @@
...
+        auto storage = details::s_apartmentStorage.find(apartmentId);
+        if (storage != details::s_apartmentStorage.end())
+        {
+            return storage->second.size();
+        }
+        return 0;
+    }
+}
+
+#endif // __WIL_COM_APARTMENT_VARIABLE_INCLUDED
```
oldnewthing, line 230: Here's another pattern: The `get` function returns a reference to the storage, not a copy. (We might even default to using the constructor instead of a custom maker function.)

```cpp
template<typename Storage>
auto& get_for_current_com_apartment()
{
   add an entry if necessary
   return a reference to the s_apartmentStorage[apartmentId][key];
}
```

The `try` version returns a pointer to the storage, or `nullptr` if it doesn't exist.
```cpp
template<typename Storage>
auto try_get_for_current_com_apartment()
{
    if (no entry exists) return nullptr;
    else return a pointer to the s_apartmentStorage[apartmentId][key];
}

```

(Returning a reference or pointer to the value storage is legal because `unordered_map` guarantees that the only thing that invalidates a key or value reference/pointer is the erasure of that key/value pair.)

This allows you to use non-copyable types with `get_for_current_com_apartment`. It also lets you do things like

```cpp
    // another widget for this thread
    ++wil::get_for_current_com_apartment<WidgetCounterStorage>();
```

or

```cpp
    // if we are the current widget, then say that there is no current widget any more
    auto currentWidget = wil::try_get_for_current_com_apartment<std::shared_ptr<Widget>>();
    if (currentWidget && *currentWidget == this) *currentWidget = nullptr;
```

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,230 @@
...
+        {
+            auto& variables = storage->second;
+            auto variable = variables.find(variableKey);
+            if (variable != variables.end())
+            {
+                return std::any_cast<std::invoke_result_t<F>>(variable->second);
+            }
+            else
+            {
+                auto result = createFn();
```
oldnewthing, line 74: Calling creation function under the lock is asking for trouble if the creation function itself calls `get_for_current_com_apartment`!

```cpp
auto GetDefaultWidgetSite() { return WidgetSite(); }

auto GetDefaultWidget() {
    return make_shared<Widget>(wil::get_for_current_com_apartment(GetDefaultWidgetSite));
}
```

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,230 @@
...
+#include <wil/result_macros.h>
+
+#ifndef WIL_ENABLE_EXCEPTIONS
+#error This header requires exceptions
+#endif
+
+namespace wil
+{
+    // Determine if apartment variables are supported in the current process context.
+    // Prior to https://microsoft.visualstudio.com/OS/_git/os.2020/pullrequest/5750493 the
```
sylveon, line 30: Maybe add a description for non-Microsoft readers?

jonwis:
Yet another option would be to declare a specific type like `wil::apartment_variable<T>` which you declare and then use like:

```c++
wil::apartment_variable<std::wstring> loggerName;
std::wstring GetNewLoggerName() { /* ... */ }
void LogThings() {
   auto name = loggerName.get_or_default(&GetNewLoggerName);
}
void UseLogIfPresent() {
    std::optional<std::wstring> ifSet = loggerName.get();
    if (ifSet) {
        /* .. */
    }
}
void Update() {
    loggerName.set(L"c:\\kittens.txt");
}

Where internally, apartment_variable<> keeps a `static std::atomic<unsigned long>` that it pulls from a module-wide `std::atomic<unsigned long>` which is the variable high-water mark.

Minimally, there's a lot of code in the template method that can't be comdat-folded but doesn't depend on the `T` template parameter. Could it be hoisted into a non-templatized helper to reduce code duplication?
```

oldnewthing:
I like @jonwis's `wil::apartment_variable<T>`. The apartment variable's address becomes the key to the map: `std::unordered_map<wil::details::apartment_variable_base*, std::any>`.

Expanding on Jon's suggestion:
```cpp
template<typename T>
struct wil::apartment_variable
{
    apartment_variable() = default;
    ~apartment_variable(); // clears all objects across all apartments

    // non-copyable, non-assignable
    apartment_variable(apartment_variable const&) = delete;
    void operator=(apartment_variable const&) = delete;

    // get current value or throw if no value has been set
    T& get_existing();

    // get current value or default-construct one on demand
    ​T& get_or_create();

    // get current value or custom-construct one on demand
    T& get_or_create(T (*creator)());
 
    // get pointer to current value or nullptr if no value has been set
    T* get_if();

    // replace or create the current value
    void set(T value);

    // remove any current value
    void clear();
};
```

The `get` and `get_if` methods return a reference/pointer to the actual value, so it can be modified in place.

Usage patterns:

```cpp
wil::apartment_variable<std::wstring> loggerName;

std::wstring GetNewLoggerName() { /* ... */ }

void LogThings() {
    auto& name = loggerName.get_or_create(&GetNewLoggerName);
    log_to_file(name);
    if (log_file_getting_big())
    {
        name = GetNewLoggerName(); // roll over to new log file
    }
}

void UseLogIfPresent() {
   auto name = loggerName.get_if();
   if (name) log_to_file(*name);
}

void StartNewLogger() {
     loggerName.set(GetNewLoggerName());
};

wil::apartment_variable<int> widgetCount;

void UpdateWidgetCounter()
{
    ++widgetCount.get_or_create();
}
```

As for comdat folding: Pretty much everything can be hoisted into this base class:

```cpp
struct wil::apartment_variable_base
{
    apartment_variable_base() = default;
    ~apartment_variable_base();

    // non-copyable, non-assignable
    apartment_variable_base(apartment_variable_base const&) = delete;
    void operator=(apartment_variable_base const&) = delete;

    // get current value or throw if no value has been set
    std::any& get_existing();

    // get current value or default-construct one on demand
   ​ std::any& get_or_create();

    // get current value or custom-construct one on demand
    std::any& get_or_create(std::any (*creator)());
 
    // get pointer to current value or nullptr if no value has been set
    std::any* get_if();

    // replace or create the current value
    void set(std::any value);

    // remove any current value
    void clear();
};
```

Specialized version would perform last-second conversion to/from `std::any`.

```cpp
template<typename T>
struct wil::apartment_variable : apartment_variable_base
{
    using base = apartment_variable_base;

    // get current value or throw if no value has been set
    T& get_existing() { return std::any_cast<T>(base.get_existing()); }

    // get current value or default-construct one on demand
    ​T& get_or_create() { return std::any_cast<T>(base.get_or_create()); }

    // get current value or custom-construct one on demand
    T& get_or_create(T (*creator)())
    { return std::any_cast<T>(base.get_or_create([creator]() { return std::any(creator()))); }
 
    // get pointer to current value or nullptr if no value has been set
    T* get_if() { return std::any_cast<T>(base.get_if()); }

    // replace or create the current value
    void set(T value) { return base.set(value); }

    // remove any current value
    using base::clear;
}
```

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,230 @@
...
+        unsigned long long apartmentId{};
+        return RoGetApartmentIdentifier(&apartmentId) != HRESULT_FROM_WIN32(ERROR_API_UNAVAILABLE);
+    }
+
+    namespace details
+    {
+        inline static winrt::slim_mutex s_lock;
+        // Apartment id -> variable storage. Variables are stored using the address of
+        // the function that produces the value as the key.
+        inline static std::unordered_map<unsigned long long, std::unordered_map<const void*, std::any>> s_apartmentStorage;
```
oldnewthing, line 43: At DLL unload, this will try to free all the objects, most of which belong to some other apartment. Is there some way to make it clearer to people that their objects need to be agile (even though it is apartment-local)?

Actually, even agility won't be good enough, because the agile wrapper will try to marshal to the original thread, which could itself be blocked on the loader lock.

So we need to figure out how to deal with apartment-local objects that are still around when the DLL unloads.

oldnewthing, line 43: We need to bypass destruction of this map at process shutdown, because the objects in the map may be references to DLLs that have already unloaded. (And certainly we would be destroying them after COM has already shut down.)

include/wil/com_apartment_variable.h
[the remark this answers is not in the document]
ChrisGuzak: right, but it only folds identical functions and identical functions will produce the same result. I suspect if the address is taken it won't fold them. but if it does no harm.

include/wil/com_apartment_variable.h
ChrisGuzak: thanks, I struggled to get this to compile and see now what you suggested works just fine.

include/wil/com_apartment_variable.h
ChrisGuzak: I found `std::enable_if_t` was not needed, the lambda and function both fail to compile without it. given that I also removed the `static_assert` (that intended to give a better error message than enable_if but now is not needed).

include/wil/com_apartment_variable.h
ChrisGuzak: did you mean to comment on the strict version? this was the open design question, I took your feedback, assuming it was on the strict version to choose the non-strict implementation (this one). I also chose the strict version with `newValue`. 

let me know if I understood you correclty.

include/wil/com_apartment_variable.h
ChrisGuzak: RoRegisterForApartmentShutdown and the related APIs is what the comment is referring to.

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,230 @@
...
+//
+//*********************************************************
+#ifndef __WIL_COM_APARTMENT_VARIABLE_INCLUDED
+#define __WIL_COM_APARTMENT_VARIABLE_INCLUDED
+
+#include <unordered_map>
+#include <any>
+#include <type_traits>
+#include <Unknwn.h>
+#include <winrt/base.h>
```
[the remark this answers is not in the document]
ChrisGuzak: yes, updated

include/wil/com_apartment_variable.h
ChrisGuzak: `ApartmentObserver` runs down the map, freeing all objects at apartment shutdown. that is the thing that enables this design. Maybe I mis-understand the case you are thinking of but by DLL unload time `s_apartmentStorage` will be empty.

ChrisGuzak: see above, `s_apartmentStorage` will be empty at DLL unload time.

include/wil/com_apartment_variable.h
ChrisGuzak: true, but the deadlock will be deterministic and easy to diagnose. I'm going to choose the simpler design as it stands unless you feel strongly.

include/wil/com_apartment_variable.h
ChrisGuzak: I took your first suggestion. on the second, my thoughts are callers that want to support empty will store a null value and treat it specially. My intended use for this feature will be to store the crawl scope manager and I plan to store the HRESULT from the attempts to get it and the pointer, if retrieved. that will enable treating the cases where the crawl scope manager can't be retrieved due to the service being disabled distinctly from other failures (out of memory).

ChrisGuzak:
I like @jonwis's wil::apartment_variable<T> too and I really appreciate you filling in the implementation details. I'm going to take a run at that version.

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,377 @@
...
+            inline static winrt::slim_mutex s_lock;
+            // Apartment id -> variable storage.
+            // Variables are stored using the address of the global variable as the key.
+            inline static std::unordered_map<unsigned long long,
+                std::unordered_map<wil::details::apartment_variable_base<test_hook>*, std::any>>
+                s_apartmentStorage;
+
+            apartment_variable_base()
+            {
+                // TODO: ensure this is a global, investigate cases where address is not reliable due to re-use as stack variables.
```
oldnewthing, line 211: If the destructor cleans up all copies, then it's okay for this to be a stack or heap variable.

Requiring this to be a global means that you can't have per-instance apartment variables, which would also be useful.

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,377 @@
...
+                        }
+                    };
+
+                    test_hook::RegisterForApartmentShutdown(winrt::make<ApartmentObserver>().get());
+
+                    auto variables = s_apartmentStorage.insert({ apartmentId,
+                        { { this, creator() } } }).first;
+                    return variables->second.begin()->second; // new map entry is first
+                }
+            }
```
oldnewthing, line 275: Need to call the `creator()` outside the lock to avoid reentrancy deadlock. Can also factor out the "create outside the lock" code as well as the "find current apartment" code.

```cpp
            using variables_map = std::unordered_map<wil::details::apartment_variable_base<test_hook>*, std::any>>;
            variables_map* get_current_apartment_variables()
            {
                const auto apartmentId = test_hook::GetApartmentId();

                auto storage = s_apartmentStorage.find(apartmentId);

                if (storage != s_apartmentStorage.end())
                {
                    return &storage->second;
                }
                return nullptr;
            }

            variables_map* ensure_current_apartment_variables()
            {
                auto variables = get_current_apartment_variables();
                if (variables)
                {
                    return variables;
                }

                struct ApartmentObserver : public winrt::implements<ApartmentObserver, IApartmentShutdown>
                {
                    void STDMETHODCALLTYPE OnUninitialize(unsigned long long apartmentId) noexcept override
                    {
                        // This code runs at apartment rundown so be careful to avoid deadlocks by
                        // extracting the variables under the lock then release them outside.
                        auto variables = [apartmentId]()
                        {
                            auto lock = winrt::slim_lock_guard(s_lock);
                            return s_apartmentStorage.extract(apartmentId);
                        }();
                    }
                };
                test_hook::RegisterForApartmentShutdown(winrt::make<ApartmentObserver>().get());

                return &s_apartmentStorage.insert({ apartmentId, {} }).first->second;
            }

            std::any& get_or_create(const std::function<std::any()>& creator)
            {
                variables_map* variables;

                { // scope for lock
                    auto lock = winrt::slim_lock_guard(s_lock);
                    variables = ensure_current_apartment_variables();

                    auto variable = variables->find(this);
                    if (variable != variables->end())
                    {
                        return variable->second;
                    }
                } // drop the lock

                // create the object outside the lock to avoid reentrancy deadlock
                auto value = creator();

                auto insert_lock = winrt::slim_lock_guard(s_lock);
                // The insertion may fail if creator() recursively caused itself to be created,
                // in which case we return the existing object and the falsely-created one is discarded.
                return variables->insert({ this, std::move(value) }).first->first;
            }
```

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,377 @@
...
+                    auto variable = variables.find(this);
+                    if (variable != variables.end())
+                    {
+                        return &(variable->second);
+                    }
+                }
+                return nullptr;
+            }
+
+            // replace or create the current value
```
oldnewthing, line 295: Comment says it can create, but code fails fast if value does not exist.

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,377 @@
...
+                    return variables->second.begin()->second; // new map entry is first
+                }
+            }
+
+            // get pointer to current value or nullptr if no value has been set
+            std::any* get_if()
+            {
+                auto lock = winrt::slim_lock_guard(s_lock);
+
+                auto storage = s_apartmentStorage.find(test_hook::GetApartmentId());
```
oldnewthing, line 282: These and other methods could take advantage of the new `get_current_apartment_variables` helper method.

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,377 @@
...
+        }
+
+        // get current value or custom-construct one on demand
+        T& get_or_create(T(*creator)())
+        {
+            return std::any_cast<T&>(base::get_or_create([creator]()
+            {
+                return std::any(creator());
+            }));
+        }
```
oldnewthing, line 362: Since you're using `std::function`, you can simplify to just

```cpp
    return std::any_cast<T&>(base::get_or_create(creator));
```
and let covariance do the work.

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,377 @@
...
+
+        // get pointer to current value or nullptr if no value has been set
+        T* get_if() { return std::any_cast<T>(base::get_if()); }
+
+        // replace or create the current value
+        void set(T&& value) { return base::set(std::forward<T>(value)); }
```
oldnewthing, line 368: Incorrect use of universal reference.

```suggestion
        template<typename U> void set(U&& value) { return base::set(std::forward<U>(value)); }

```

tests/ComApartmentVariableTests.cpp
```diff
@@ -0,0 +1,244 @@
...
+{
+    auto coUninit = mock_platform::CoInitializeExForTesting(COINIT_MULTITHREADED);
+
+    std::ignore = g_v1.get_or_create(fn);
+
+    wil::apartment_variable<int, mock_platform> v1; // 00000020E63CF6F4, should be a global
+
+    REQUIRE(v1.get_if() == nullptr);
+    REQUIRE(v1.get_or_create(fn) == 42);
+    v1.set(43);
```
oldnewthing, line 173: Need to test setting by lvalue.

```cpp
int value = 44;
v1.set(value);
```

would have caught incorrect universal reference.

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,234 @@
...
+
+            static void RegisterForApartmentShutdown(IApartmentShutdown* observer)
+            {
+                unsigned long long id{};
+                APARTMENT_SHUTDOWN_REGISTRATION_COOKIE cookie{}; // leaked, COM run down releases all instances
+                THROW_IF_FAILED(RoRegisterForApartmentShutdown(observer, &id, &cookie));
+            }
+        };
+
+        template<typename test_hook = apt_var_platform>
```
jonwis, line 58: Maybe call this `apartment_source` instead?

include/wil/win32_helpers.h
```diff
@@ -20,7 +20,7 @@
 
 // detect std::bit_cast
 #ifdef __has_include
-#  if __has_include(<bit>)
+#  if (__cplusplus > 202000) && __has_include(<bit>)
```
jonwis, line 23: 👍

sylveon: `/Zc:__cplusplus` isn't the default in MSVC or Visual Studio yet so I think this change should also check for `_MSVC_LANG` as well.

```suggestion
#  if (_MSVC_LANG >= 202002L || __cplusplus >= 202002L) && __has_include(<bit>)

```

include/wil/com_apartment_variable.h
ChrisGuzak: thanks for the impl. updated

include/wil/com_apartment_variable.h
ChrisGuzak: "forwarding reference" ;-). Fixed.

include/wil/com_apartment_variable.h
ChrisGuzak: its not a source of apartments

ChrisGuzak: other ideas? I'm going to won't fix.

oldnewthing: This is dependency injection, so `dependency` would be a not unreasonable name.

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,243 @@
...
+                            return s_apartmentStorage.extract(apartmentId);
+                        }();
+                    }
+                };
+                test_hook::RegisterForApartmentShutdown(winrt::make<ApartmentObserver>().get());
+                return &s_apartmentStorage.insert({ test_hook::GetApartmentId(), {} }).first->second;
+            }
+
+            // get current value or custom-construct one on demand
+            std::any& get_or_create(const std::function<std::any()>& creator)
```
oldnewthing, line 125: Can avoid the overhead of `std::function` by using a custom lightweight functor.

```cpp
struct any_maker_base
{
    std::any (*adapter)(void*);
    void *inner;

    std::any operator()()
    {
        return adapter(inner);
    }
};

template<typename T>
struct any_maker : any_maker_base
{
    any_maker()
    {
        adapter = [](auto) -> std::any { return T{}; };
    }

    any_maker(T (*maker)())
    {
        adapter = [](auto maker) -> std::any { return reinterpret_cast<T(*)()>(maker)(); };
        inner = reinterpret_cast<void*>(maker);
    }

    template<typename F>
    any_maker(F&& f)
    {
        adapter = [](auto maker) -> std::any { return reinterpret_cast<F*>(maker)[0](); };
        inner = std::addressof(f);
    }
};
```

The `any_maker` also supports an arbitrary functor, which is handy in the case of apartment variables that are not global variables.

Then we get
```cpp
std::any& get_or_create(const any_maker& creator)
{
   ... as before ...
}
```

The `apartment_variable` methods would become

```cpp
        T& get_or_create()
        {
            return base::get_or_create(details::any_maker<T>());
        }

        // get current value or custom-construct one on demand
        template<typename F>
        T& get_or_create(F&& f)
        {
            return std::any_cast<T&>(base::get_or_create(details::any_maker<T>(std::forward<F>(f))));
        }
```

include/wil/com_apartment_variable.h
ChrisGuzak: `wil::shutdown_aware_object` provides this

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,418 @@
...
+        struct apartment_variable_base
+        {
+            inline static winrt::slim_mutex s_lock;
+
+            struct apartment_variable_storage
+            {
+                apartment_variable_storage(apartment_variable_storage&& other) noexcept = default;
+                apartment_variable_storage(const apartment_variable_storage& other) = delete;
+
+                apartment_variable_storage(typename test_hook::shutdown_type&& cookie_) : cookie(std::move(cookie_))
```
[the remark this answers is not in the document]
ChrisGuzak: that would support copy and I don't want that (cookies are not copyable).

include/wil/com_apartment_variable.h
```diff
@@ -0,0 +1,418 @@
...
+
+    // Apartment variables enable storing COM objects safely in globals
+    // (objects with static storage duration) by creating a unique copy
+    // in each apartment and managing their lifetime based on apartment rundown
+    // notifications.
+    // They can also be used for automatic or dynamic storage duration but those
+    // cases are less common.
+    // This type is also useful for storing references to apartment affine objects.
+    //
+    // Note, that apartment variables hosted in a COM DLL need to
```
brialmsft, line 370: Missing a word at the end of this line. Possibly "modify"?

tests/ComApartmentVariableTests.cpp
```diff
@@ -0,0 +1,427 @@
...
+    // Enable the test hook to force losing the race
+    inline static constexpr unsigned long AsyncRundownDelayForTestingRaces = 1; // enable test hook
+    inline static std::unordered_map<unsigned long long, std::vector<wil::com_ptr<IApartmentShutdown>>> m_observers;
+};
+
+auto fn() { return 42; };
+auto fn2() { return 43; };
+
+wil::apartment_variable<int, mock_platform> g_v1;
+wil::apartment_variable<int> g_v2;
```
brialmsft, line 115: I'm a bit confused by these. g_v2 appears to be completely unreferenced. g_v1 is referenced only from TestApartmentVariableAllMethods, and only minimally there (and uses mock_platform even in the UseRealPlatformRunAllTests case). Are these just meant to smoke-test the global variable-specific stuff, or are they potentially incomplete (or am I just missing something)?

brialmsft (approved):

tests/ComApartmentVariableTests.cpp
ChrisGuzak: smoke test of global variables.

jonwis (approved):
Having `v.set(m)` fail if it's the first call on the thread is a little tricksy, so you have to always say `v.get_or_create() = m`

ChrisGuzak:
Now documented [here](https://github.com/microsoft/wil/wiki/WinRT-and-COM-wrappers#wilapartment_variable)

Issue #189, opened 2021-06-28
Add wil::unique_environment_block to resource.h


m417z:
Add an RAII wrapper that frees an environment block allocated with `CreateEnvironmentBlock`. Something similar to:

    using unique_environment_block = wil::unique_any<void*, decltype(&::DestroyEnvironmentBlock), ::DestroyEnvironmentBlock>;

sylveon:
Should be pretty trivial to PR

PR #192, opened 2021-07-12
Added unique_environment_block to resource.h

m417z:
Fixes https://github.com/microsoft/wil/issues/189

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -5361,6 +5361,11 @@ namespace wil
     typedef unique_any<HTHEME, decltype(&::CloseThemeData), ::CloseThemeData> unique_htheme;
 #endif // __WIL_INC_UXTHEME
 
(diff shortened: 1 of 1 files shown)


include/wil/resource.h
```diff
@@ -5361,6 +5361,11 @@ namespace wil
     typedef unique_any<HTHEME, decltype(&::CloseThemeData), ::CloseThemeData> unique_htheme;
 #endif // __WIL_INC_UXTHEME
 
+#if defined(_INC_USERENV) && !defined(__WIL_INC_USERENV) && !defined(WIL_KERNEL_MODE)
```
sylveon, line 5364: I think you need a check for `WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_SYSTEM)` as well.

m417z: Done, thanks.

sylveon (approved):
LGTM, I'd add the shared_ and weak_ variants as well but they may not be very useful for environment blocks.

include/wil/resource.h
jonwis: I look forward to the day when we can get rid of that entirely. :|

jonwis (approved):

dunhor (approved):

PR #199, opened 2021-08-06
wil::resume_foreground addresses many issues with winrt::resume_foreground

oldnewthing:
The C++/WinRT built-in `resume_foreground()` support is problematic.

`co_await resume_foreground(CoreDispatcher)` has no return value. If the resumption of the coroutine could not be enqueued, the coroutine simply hangs, resulting in a memory leak (from accumulated suspended coroutines that never complete), while giving the developer no indication that the work they wanted to do will never be done.

`co_await resume_foreground(DispatcherQueue)` returns `false` if the resumption of the coroutine could not be enqueued and the thread switch therefore failed. In practice, nobody checks the return value, meaning that the code after the `co_await` is running on the wrong thread. (This flaw was copy/paste'd into `Microsoft.UI.Dispatching.co_await.h`.)

Furthermore, just because the coroutine continuation was enqueued doesn't mean that it will actually run. The work item may end up simply abandoned, for example if the dispatcher thread exits abruptly.

To fix all this, we provide alternate versions `wil::resume_foreground()` which have these properties:

* If the thread-switch is successful, the `co_await` completes normally.
* If the thread-switch is unsuccessful, the `co_await` completes with an exception on an unspecified thread.

Another problem with C++/WinRT's `winrt::resume_foreground()` is simply that it's in C++/WinRT. It really belongs in an outside library, since C++/WinRT should be focused on the projection. Putting them in C++/WinRT forces dispatcher implementors to have to modify C++/WinRT whenever they want to change something.

Implementation details
======================

The implementation is tricky because `wil/cppwinrt.h` is included before any other C++/WinRT headers, yet it wants to access types from headers that may be included after it. It cannot `#include` those headers directly, because they may not exist.

The solution is to forward-declare the types, and then use them as dependent types in template expansions, so that the members aren't resolved until the second phase of template name lookup, at which point the full definitions should have been provided by the calling app.

We declare a traits type `dispatcher_traits` and specialize it for each of the dispatchers we support. This also serves as a convenient hook for the unit test: We can just invent our own custom dispatcher, and as long as it has a `dispatcher_traits`, we can use it with `wil::resume_foreground`.

In order to be able to call `RunAsync` or `TryEnqueue` before the methods have been declared, we make them dependent names in `dispatcher_RunAsync` and `dispatcher_TryEnqueue`, so that the name lookup happens at invoke time rather than definition time.

To detect abandoned work, we use a custom delegate which detects that it has been destructed without ever having been invoked, in which case it resumes the coroutine so it can raise an exception with error code `ERROR_NO_TASK_QUEUE`. (We ignore the return value from `Scheduler::Schedule` because a successful return value does not guarantee that the work item will run.)

The abandonment could take place on any thread, because it might occur after the work was successfully queued, and then before it can run, the destination thread exits. The work item is now stuck in limbo: The destination thread no longer exists, and the source thread may not exist either. It has no choice but to resume on some other thread.

The state of the suspended coroutine is kept in a `dispatched_handler_state`. Our `m_state` pointer is non-null if we are the "active" `dispatcher_handler`. The pointer is moved when the move-only `dispatcher_handler` is moved, so only the instance that lives inside the delegate has a non-null pointer, and even that instance nulls out its pointer before resuming the coroutine, so that we never have a pointer to a destructed state.

The `handle` is non-null if the coroutine is still suspended. `orphaned` is set to `true` if the handler discovers that it was never invoked. Upon resumption, we raise an exception if we realize we were orphaned.

There is a subtlety in the `await_suspend`: We need to keep an explicit reference to the work item, so that it doesn't destruct prematurely. In the case where the Queue function throws an exception, we null out the coroutine handle in the state so that the handler's destructor won't try to report the failure: The exception in the Queue function puts the the coroutine into an error state, making it no longer resumable.

A bonus side effect of the replacement `wil::resume_foreground` is that its `await_suspend` returns `void` rather than `bool`, thereby avoiding a heap corruption coroutine code generation bug in older versions of Visual Studio.

One of the problematic pieces is figuring out whether coroutine handles live in the `std` or `std::experimental` namespace (or neither, if coroutines aren't supported at all). This is particularly tricky because different versions of C++/WinRT leave different traces behind. Ultimately, I opted not to rely on C++/WinRT at all and instead use the compiler symbols.

* `_RESUMABLE_FUNCTIONS_SUPPORTED`: `/await` mode (pre-standard)
* `__cpp_lib_coroutine`: C++20 standard mode

### DIFF
--- a/include/wil/cppwinrt.h
+++ b/include/wil/cppwinrt.h
@@ -337,6 +337,7 @@ namespace wil
 
 #if (defined(__cpp_lib_coroutine) && (__cpp_lib_coroutine >= 201902L)) || defined(_RESUMABLE_FUNCTIONS_SUPPORTED)
 
+/// @cond
 #if defined(_RESUMABLE_FUNCTIONS_SUPPORTED)
 #include <experimental/coroutine>
 namespace wil::details
@@ -350,8 +351,6 @@ namespace wil::details
     template<typename T = void> using coroutine_handle = std::coroutine_handle<T>;
 }
 #endif
-
-/// @cond
 
 namespace winrt::Windows::UI::Core
 {
@@ -390,7 +389,7 @@ namespace wil
         struct dispatcher_RunAsync
         {
             template<typename Dispatcher, typename... Args>
-            static void Queue(Dispatcher const& dispatcher, Args&&... args)
+            static void Schedule(Dispatcher const& dispatcher, Args&&... args)
             {
                 dispatcher.RunAsync(std::forward<Args>(args)...);
             }
@@ -399,7 +398,7 @@ namespace wil
         struct dispatcher_TryEnqueue
         {
             template<typename Dispatcher,typename... Args>
-            static void Queue(Dispatcher const& dispatcher, Args&&... args)
+            static void Schedule(Dispatcher const& dispatcher, Args&&... args)
             {
                 dispatcher.TryEnqueue(std::forward<Args>(args)...);
             }
@@ -412,7 +411,7 @@ namespace wil
         {
             using Priority = winrt::Windows::UI::Core::CoreDispatcherPriority;
             using Handler = winrt::Windows::UI::Core::DispatchedHandler;
-            using Queuer = dispatcher_RunAsync;
+            using Scheduler = dispatcher_RunAsync;
         };
 
         template<>
@@ -420,7 +419,7 @@ namespace wil
         {
             using Priority = winrt::Windows::System::DispatcherQueuePriority;
             using Handler = winrt::Windows::System::DispatcherQueueHandler;
-            using Queuer = dispatcher_TryEnqueue;
+            using Scheduler = dispatcher_TryEnqueue;
         };
 
         template<>
@@ -428,7 +427,7 @@ namespace wil
         {
             using Priority = winrt::Microsoft::System::DispatcherQueuePriority;
             using Handler = winrt::Microsoft::System::DispatcherQueueHandler;
-            using Queuer = dispatcher_TryEnqueue;
+            using Scheduler = dispatcher_TryEnqueue;
         };
 
         template<>
@@ -436,7 +435,7 @@ namespace wil
         {
             using Priority = winrt::Microsoft::UI::Dispatching::DispatcherQueuePriority;
             using Handler = winrt::Microsoft::UI::Dispatching::DispatcherQueueHandler;
-            using Queuer = dispatcher_TryEnqueue;
+            using Scheduler = dispatcher_TryEnqueue;
         };
 
         struct dispatched_handler_state
@@ -501,7 +500,9 @@ namespace wil
                 Handler handler{ details::dispatcher_handler(&m_state) };
                 try
                 {
-                    Traits::Queuer::Queue(m_dispatcher, m_priority, handler);
+                    // The return value of Schedule is not reliable. Use the dispatcher_handler destructor
+                    // to detect whether the work item failed to run.
+                    Traits::Scheduler::Schedule(m_dispatcher, m_priority, handler);
                 }
                 catch (...)
                 {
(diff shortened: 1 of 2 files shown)


include/wil/cppwinrt.h
```diff
@@ -335,4 +335,210 @@ namespace wil
...
+                dispatcher.RunAsync(std::forward<Args>(args)...);
+            }
+        };
+
+        struct dispatcher_TryEnqueue
+        {
+            template<typename Dispatcher,typename... Args>
+            static void Queue(Dispatcher const& dispatcher, Args&&... args)
+            {
+                dispatcher.TryEnqueue(std::forward<Args>(args)...);
```
benwestprog, line 404: > dispatcher.TryEnqueue(std::forward<Args>(args)...); [](http://example.com/codeflow?start=15&length=52)

Should we be checking the return value here? Or does the wrapper struct convert the bool into an exception somehow?

jonwis, line 404: Does this need to check the return of `TryEnqueue` and throw if it returns false?

tests/CppWinRTTests.cpp
```diff
@@ -142,3 +143,120 @@ TEST_CASE("CppWinRTTests::CppWinRTConsistencyTest", "[cppwinrt]")
...
+
+    struct TestDispatcher
+    {
+        TestDispatcher() = default;
+        TestDispatcher(TestDispatcher const&) = delete;
+
+        TestDispatcherMode mode = TestDispatcherMode::Dispatch;
+        TestDispatcherPriority expected_priority = TestDispatcherPriority::Normal;
+
+        void TryEnqueue(TestDispatcherPriority priority, TestDispatcherHandler const& handler) const
```
jonwis, line 178: https://docs.microsoft.com/en-us/uwp/api/windows.system.dispatcherqueue.tryenqueue?view=winrt-20348

Probably need a mode where `TryEnqueue` says it's displeased by returning 'false'. Docs say that it returns false when the dispatcher has been shut down, so that's presumably under app control (or can be known by the app.)

jonwis (changes requested):

include/wil/cppwinrt.h
oldnewthing: We already designed for the case where `DispatcherQueue.TryEnqueue` and `CoreDispatcher.RunAsync` might accept the work item and then later change their mind and release it without ever invoking it. If `TryEnqueue` returns false, that same "changed my mind" logic will kick in: The `DispatcherQueueHandler` is destructed without ever being invoked, meaning that the work item failed to run, and the coroutine should resume with an exception.

oldnewthing: We already designed for the case where `DispatcherQueue.TryEnqueue` and `CoreDispatcher.RunAsync` might accept the work item and then later change their mind and release it without ever invoking it. If `TryEnqueue` returns false, that same "changed my mind" logic will kick in: The `DispatcherQueueHandler` is destructed without ever being invoked, meaning that the work item failed to run, and the coroutine should resume with an exception.

tests/CppWinRTTests.cpp
oldnewthing: DispatcherQueue might return `true` but still end up never invoking the handler. The return value is useless to us. (Returning `false` means "Your handle is definitely not going to be invoked', but `true` means "Your handler is probably going to be invoked, but maybe not.") No need to add extra logic for the `false` case since it's already covered by the `true` case.

jonwis (approved):

include/wil/cppwinrt.h
benwestprog: Since Jon and I both missed that, worth adding a brief comment explaining why we ignore the return value? This code is tricky!

benwestprog (approved):
:shipit:

ChrisGuzak (approved):

sylveon:
Why not just explicitly refer to the global namespace? `::winrt::Microsoft`

oldnewthing:
@jonwis @dunhor Rats, namespace name collisions. Again. Ideas:

1. Make cppwinrt.h a "multi-include" style header, where you can `#include` it multiple times, and each time you include it, it lights up whatever new things showed up since the previous `#include`.

```cpp
#include <wil/cppwinrt.h> // No wil::resume_foreground support yet
#include <winrt/Microsoft.System.h>
#include <wil/cppwinrt.h> // turns on wil::resume_foreground(winrt::Microsoft::System::DispatcherQueue)
#include <winrt/Windows.UI.Core.h>
#include <wil/cppwinrt.h> // turns on wil::resume_foreground(winrt::Windows::UI::Core::CoreDispatcher)
```

2. Move the `resume_foreground` stuff into a separate `cppwinrt_helpers.h` which is multi-includable.

```cpp
#include <wil/cppwinrt.h> // No wil::resume_foreground support yet
#include <winrt/Microsoft.System.h>
#include <wil/cppwinrt_helpers.h> // turns on wil::resume_foreground(winrt::Microsoft::System::DispatcherQueue)
#include <winrt/Windows.UI.Core.h>
#include <wil/cppwinrt_helpers.h> // turns on wil::resume_foreground(winrt::Windows::UI::Core::CoreDispatcher)
```

PR #201, opened 2021-08-16
Add wil\coroutine.h for wil::task and related classes

oldnewthing:
A `wil::task<T>` / `com_task<T>` is a coroutine with the following characteristics:

- `T` must be a copyable object, movable object, reference, or void.
- The coroutine may be awaited at most once. The second await will crash.
- The coroutine may be abandoned (allowed to destruct without `co_await`), in which case unobserved exceptions are fatal.
- The awaiting coroutine may be destroyed while suspended. (Don't worry if you don't know what this means.)
- By default, `task` resumes in an arbitrary COM context.
- By default, `com_task` resumes in the same COM context.
- `task.resume_any_thread()` forces agile behavior.
- `task.resume_same_context()` forces non-agile behavior.

The `com_task` and `task` are intended to supplement PPL and C++/WinRT, not to replace them. It provides coroutine implementations for scenarios that PPL and C++/WinRT do not support.

The implementation is optimized on the assumption that the coroutine is awaited only once, and that the coroutine is discarded after completion. To ensure proper usage, the task object is move-only, and `co_await` takes ownership of the task. See further discussion below.

Comparison with PPL and C++/WinRT:

|                                                     | PPL       | C++/WinRT | simple_*_task |
|-----------------------------------------------------|-----------|-----------|---------------|
| `T` can be non-constructible                        | No        | Yes       | Yes           |
| `T` can be void                                     | Yes       | Yes       | Yes           |
| `T` can be reference                                | No        | No        | Yes           |
| `T` can be WinRT object                             | Yes       | Yes       | Yes           |
| `T` can be non-WinRT object                         | Yes       | No        | Yes           |
| `T` can be move-only                                | No        | No        | Yes           |
| Coroutine can be cancelled                          | Yes       | Yes       | No            |
| Awaiting coroutine may be destroyed while suspended | No        | No        | Yes           |
| Coroutine can throw arbitrary exceptions            | Yes       | No        | Yes           |
| Can `co_await` more than once                       | Yes       | No        | No            |
| `co_await` resumes in same COM context              | Sometimes | Yes       | You choose [1]|
| Can force `co_await` to resume in same context      | Yes       | N/A       | Yes [1]       |
| Can force `co_await` to resume in any context       | Yes       | No        | Yes           |
| Can change coroutine's resumption model             | No        | No        | Yes           |
| Can wait synchronously                              | Yes       | Yes       | Yes [2]       |
| Can be consumed by non-C++ languages                | No        | Yes       | No            |
| Implementation is small and efficient               | No        | Yes       | Yes           |
| Can abandon coroutine (fail to `co_await`)          | Yes       | Yes       | Yes           |
| Exception in abandoned coroutine                    | Crash     | Ignored   | Crash         |
| Coroutine starts automatically                      | Yes       | Yes       | Yes           |
| Coroutine starts synchronously                      | No        | Yes       | Yes           |
| Integrates with C++/WinRT coroutine callouts        | No        | Yes       | No            |

[1] Resumption in the same COM context requires that you include COM headers.
[2] Synchronous waiting requires that you include synchapi.h (usually via windows.h).

You can include the COM headers and/or synchapi.h headers, and then re-include this header file to activate the features dependent upon those headers.

[**Background reading**](https://devblogs.microsoft.com/oldnewthing/20210504-01/?p=105178) dives deep into the design and development of this header file.

Open issues
===========

* What should we call these things? Possible names (or suggest your own):

| COM-aware     | Agile               | Notes                                                                |
|---------------|---------------------|----------------------------------------------------------------------|
| `simple_task` | `simple_agile_task` | Aligns with C++/WinRT where COM-awareness is default. (But not PPL.) |
| `task`        | `agile_task`        | Get rid of "simple" prefix.                                          |
| `com_task`    | `agile_task`        | Name the COM-aware one explicity "com", though may overpromise COM support |
| `com_aware_task` | `agile_task`     | `com_aware` more clearly captures the behavior (it knows about COM but doesn't go out of its way to do anything fancy.) |
| `com_aware_task` | `task`           | Give the simple name to the agile version.                           |

Conclusion: Give the simple name to the agile version, consistent with the C++ standard library where agility is the default. The COM-aware one is called `com_task`.

* What should happen if you abandon a task (fail to `co_await` it)?
   1. Illegal (fatal exception).
   2. Legal as long as task completes without an exception, but throws fatal exception if task encountered an exception. (Current implementation.)
   3. Legal, and any exceptions in the task are ignored.

* Do we need cold-start tasks? (Can be added, but no point if there is no demand.)
* Should we continue to use wistd or just use the standard versions?
* Should it support the C++/WinRT coroutine hooks?

Examples
========

Implement a coroutine that returns a move-only non-WinRT type and which resumes in the same context in which it was called.

```cpp
wil::com_task<wil::unique_cotaskmem_string> GetNameAsync()
{
    co_await resume_background(); // do work on BG thread
    wil::unique_cotaskmem_string name;
    THROW_IF_FAILED(GetNameSlow(&name));
    co_return name; // awaiter will switch back to original context
}
```

Consumers:
```cpp
winrt::IAsyncAction UpdateNameAsync()
{
    // com_task resumes on the same COM context
    auto name = co_await GetNameAsync();
    // so we are still on the UI thread
    NameElement().Text(winrt::hstring(name.get()));
}

winrt::IAsyncAction UpdateNameAsync()
{
    // override default behavior of com_task and
    // allow it to resume on any thread.
    auto name = co_await GetNameAsync().resume_any_thread();
    // could be on any thread now
    co_await SendNameAsync(name.get());
}
```

Conversely, you can define a coroutine that returns a `wil::task<T>` and it will default to resuming on any thread, but you can force it to resume in the same COM context by doing `co_await GetNameAsync().resume_same_context()`.

There is no harm in doing `resume_same_context()` / `resume_any_thread()` for a task that already defaults to resuming in that context. In fact, awaiting the task directly is just a shorthand for awaiting the corresponding `resume_whatever_context()` method.

Alternatively, you can convert between `task<T>` and `com_task<T>` to change the default resumption context.

```cpp
co_await com_task(GetNameAsync()); // now defaults to resume_same_context();
```

You can store the task in a variable, but since it is a move-only object, you will have to use `std::move` in order to transfer ownership out of an lvalue.

```cpp
winrt::IAsyncAction SomethingAsync()
{
    com_task<int> task;
    switch (source)
    {
    // Some of these might return task<int>,
    // but assigning to a com_task<int> will make
    // the task non-agile.
    case widget: task = GetValueFromWidgetAsync(); break;
    case gadget: task = GetValueFromGadgetAsync(); break;
    case doodad: task = GetValueFromDoodadAsync(); break;
    default:     FAIL_FAST(); // unknown source
    }
    auto value = co_await std::move(task); // **** need std::move
    DoSomethingWith(value);
}
```

You can wait synchronously by calling `get()`. The usual caveats about synchronous waits on STA threads apply.


mikrause79 (approved):

ChrisGuzak:
On the names, I vote for `wil::task` and `wil::com_task` as the names for these things.

ChrisGuzak (changes requested):
Per our discussion today, rename to `wil::task` and `wil::com_task`

ChrisGuzak (approved):

include/wil/coroutine.h
```diff
@@ -0,0 +1,791 @@
...
+                    {
+                        WI_ASSERT(waiting != completed_ptr());
+                        self.resume_waiting_coroutine(waiting);
+                    }
+                };
+            };
+            return awaiter{ {}, *this };
+        }
+
+        // The remaining methods are used by the awaiters.
```
TheJCAB, line 414: These "client" awaiter functions are a great facility for creating custom ways for the task to produce its value. We needed a different implementation of `get()` in TTD (we still support Windows 7), for which we needed direct access to these.

We request the addition of public thin thunks to all three functions in the task_base class. Details of our findings are provided offline.

include/wil/coroutine.h
```diff
@@ -0,0 +1,791 @@
...
+
+        template<typename D>
+        D& assign(D* self, task_base&& other) noexcept
+        {
+            static_cast<task_base&>(*this) = wistd::move(other);
+            return *self;
+        }
+
+    private:
+        promise_ptr<T> promise;
```
TheJCAB, line 541: As part o trying this on TTD we found a need to query a task for whether it's `empty()` (whether the `promise` pointer is `nullptr`. In a way, a task object has two states: it either contains a coroutine that will produce a value or it doesn't. Default-constructed task objects, and task objects that have been awaited (via `co_await` or via `get()`) don't contain a coroutine, so in STL container parlance they are `empty()`.

An alternative is to implement the explicit `operator bool()` like the STL does for smart pointers and for `optional`, but for a task object there are two possible meanings of the bool conversion. A boolean of `true` could mean `!empty()` or it could mean that the coroutine has already produced a value. This semantic ambiguity makes me prefer an explicitly-named function.

Aside: providing another function to check for coroutine completion would also be simple and might help in some scenarios.

TheJCAB (changes requested):
We tested this as a way to replace PPL in the Time Travel Debugging implementation, and it works very well. Leaner, and with better object lifetime and ownership considerations than PPL.

TTD's use of wil::task did require a couple of small additions that we performed locally and list in this review. We don't believe the additions represent bloat, but they would require some amount of documentation on their correct use.

include/wil/coroutine.h
```diff
@@ -0,0 +1,791 @@
...
+    template<typename T>
+    struct task_promise;
+
+    // Unions may not contain references, C++/CX types, or void.
+    // To work around that, we put everything inside a result_wrapper
+    // struct, and put the struct in the union. For void,
+    // we create a special empty structure.
+    //
+    // get_value returns rvalue reference to T for object
+    // category, or just T itself for void and reference
```
TheJCAB, line 187: "get_value returns rvalue reference to T for object category" is incorrect (stale comment?)

In fact, `get_value()` is clearly destructive for move-only objects. Would it be reasonable to make it an rvalue method, along with `client_await_resume()`?

include/wil/coroutine.h
```diff
@@ -0,0 +1,791 @@
...
+                    // can read the results.
+                    auto waiting = self.m_waiting.exchange(completed_ptr(), std::memory_order_acq_rel);
+                    if (waiting == abandoned_ptr())
+                    {
+                        self.destroy();
+                    }
+                    else if (waiting != running_ptr())
+                    {
+                        WI_ASSERT(waiting != completed_ptr());
+                        self.resume_waiting_coroutine(waiting);
```
davidmatson, line 407: I don't think task<T> currently supports symmetric transfer. Could it? That would be be a nice stack depth/perf improvement here.

Specifically, if what's waiting is a std::coroutine_handle<>, returning a std::coroutine_handle<> from await_suspend, rather than void, would mean the waiting coroutine could be executed directly without the extra stack depth of the current coroutine. See:
https://en.cppreference.com/w/cpp/coroutine/noop_coroutine
and
https://lewissbaker.github.io/2020/05/11/understanding_symmetric_transfer

oldnewthing: At the time this was originally written, MSVC didn't support symmetric transfer. Even after MSVC added support for symmetric transfer, the implementation was buggy. It was [finally fixed](https://devblogs.microsoft.com/cppblog/cpp20-coroutine-improvements-in-visual-studio-2019-version-16-11/) two weeks after the PR was created, but it will take time for everybody to migrate off of the older version of the compiler.

sylveon: Sounds like something that could be guarded under `#if _MSC_VER > some_version`?

oldnewthing: Yes, but it also creates interop issues, where you cannot reliably link together libraries built with different versions of MSVC (ODR violations).

sylveon: Yeah, an opt-in that's automatically enabled in C++20 mode (or with `/await:strict`) sounds good.

AtariDreams:
Any update on this?

jonwis:
I'm in need of this so I can add wil::async_scope_lock, which would be this:

```c++
struct async_scope_lock
{
    wil::unique_semaphore lock{1, 1};
    wil::task<semaphore_release_scope_exit> enter() {
        co_await resume_on_signal(lock.get());
        return lock.ReleaseSemaphore_scope_exit()
    }
};

winrt::fire_and_forget Foo::OnClickaroo(...)
{
    auto strong = get_strong();
    auto block = co_await m_onclick_scopelock.enter(); // the lock is released when the task result is destroyed
}
```

... I'm finding people who are trying to emulate this with a critical_section or a boolean and not succeeding.

oldnewthing:
I'll merge this as-is and deal with @TheJCAB's revisions separately.

include/wil/coroutine.h
```diff
@@ -0,0 +1,810 @@
...
+        //
+        // * no parameters (void category)
+        // * The reference type T (reference category)
+        // * Some kind of reference to T (object category)
+        //
+        // Set the status after constructing the object.
+        // That way, if object construction throws an exception,
+        // the holder remains empty.
+        template<typename...Args>
+        void set_value(Args&&... args)
```
dunhor, line 242: Consider renaming these to something like `emplace` or `emplace_value` for consistency with the language

PR #202, opened 2021-08-16
`lock_winrt_module()` to prevent C++/WinRT host module from unloading

oldnewthing:
Active coroutines normally prevent the C++/WinRT host module from unloading because they employ one or more of these techniques:

* They are part of an IAsyncAction or IAsyncOperation, which retains a reference to the host module until they are destructed.
* They explicitly take a reference to their parent object via `get_strong()`.

The missing case is a `fire_and_forget` coroutine that is not an instance member function. (Either a free function or a static member function.) These things are not attached to COM objects and therefore cannot prevent the host DLL from unloading.

The new `wil::lock_winrt_module()` method acquires a reference count on the C++/WinRT host object's module lock (`winrt::get_module_lock()`) and returns an RAII type that releases the reference upon destruction.

Example usage
=============

```cpp
winrt::fire_and_forget DoSomethingInBackground()
{
    // take a strong reference to the host module
    // to prevent premature unload.
    auto module_strong = wil::lock_winrt_module();

    co_await DoStuffThatUpdatesGlobals();

    // module_strong destructor release the reference,
    // allowing the module to be unloaded.
}
```

### DIFF
--- a/include/wil/cppwinrt.h
+++ b/include/wil/cppwinrt.h
@@ -16,6 +16,7 @@
 #include <unknwn.h>
 #include <inspectable.h>
 #include <hstring.h>
+#include "resource.h"
 
 // WIL and C++/WinRT use two different exception types for communicating HRESULT failures. Thus, both libraries need to
 // understand how to translate these exception types into the correct HRESULT values at the ABI boundary. Prior to
@@ -332,6 +333,22 @@ namespace wil
     auto capture_interop(winrt::Windows::Foundation::IUnknown const& o, HRESULT(__stdcall Interface::* method)(InterfaceArgs...), Args&&... args)
     {
         return winrt::capture<WinRTResult>(o.as<Interface>(), method, std::forward<Args>(args)...);
+    }
+
+    namespace details
+    {
+        inline void decrement_winrt_module_ref() noexcept
+        {
+            --winrt::get_module_lock();
+        }
+    }
+
+    using unique_winrt_module_ref = unique_call<decltype(details::decrement_winrt_module_ref), &details::decrement_winrt_module_ref>;
+
+    [[nodiscard]] unique_winrt_module_ref lock_winrt_module() noexcept
+    {
+        ++winrt::get_module_lock();
+        return unique_winrt_module_ref();
     }
 }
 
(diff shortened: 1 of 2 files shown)


jonwis:
This is a common problem even for non-COM/WinRT DLLs as well.  A dll that uses coroutines (or threadpool, or createthread) needs to acquire and pin a module handle and then free it when the work item terminates.

Should the coroutine mechanics also bump the loader's refcount with `GetModuleHandleEx` using the handler function's address to map to a module handle?

ChrisGuzak:
@Jonwis, my understanding is you must call FreeLibraryAndExitThread for that to work. is that possible in the cases you are thinking of?

ChrisGuzak (approved):

sylveon:
Why not add this to C++/WinRT? Or maybe even make it built-in to `fire_and_forget`?

oldnewthing:
> Why not add this to C++/WinRT? Or maybe even make it built-in to `fire_and_forget`?

C++/WinRT should be focused on the projection. It was not intended to be a general-purpose coroutine library.

Helper functions like this can be provided by outside libraries, just like how System.Threading.Tasks is not part of the Windows Runtime C# projection, but is rather a separate library in the BCL.

sylveon:
Being part of wil certainly makes the discoverability for non-MS employees harder, as wil being public is still relatively recent, and the documentation for C++/WinRT never mentions it.

dmachaj:
This seems genuinely useful and solves a problem.  As-is it is not a lot of code and gets the job done.  It does rely on cppwinrt to do the heavy lifting so I wonder if it would make sense to have an open backlog item to create a version of this that can be used by code bases that don't otherwise rely on cppwinrt.  If that is not widely needed then it can go un-implemented.  If others do need it then someone can add it in the future.

oldnewthing:
> I wonder if it would make sense to have an open backlog item to create a version of this that can be used by code bases that don't otherwise rely on cppwinrt.

Good point. I'll add a WRL version to `cppwinrt_wrl.h`. (Could in theory also add an ATL version, but it looks like WIL doesn't bother supporting ATL, so why start now.)

oldnewthing:
Gonna solve this a different way when I add WRL support. Abandoning this one.

PR #204, opened 2021-08-18
Update the cpplatest test to use C++20

dunhor:
Currently it's targeting C++17, which is a little bit out of date. Seems like CMake & co. have been updated to allow C++20 as a target. This should help avoid build issues with cppwinrt that we're seeing.

I also explicitly tested adding an include of `winrt/Windows.Foundation.h` to make sure that the compilation errors related to that were gone.

### DIFF
--- a/include/wil/result.h
+++ b/include/wil/result.h
@@ -252,7 +252,8 @@ namespace wil
                 if (ProcessShutdownInProgress())
                 {
                     // There are no other threads to contend with.
-                    if (--m_refCount == 0)
+                    m_refCount = m_refCount - 1;
+                    if (m_refCount == 0)
                     {
                         m_data.ProcessShutdown();
                     }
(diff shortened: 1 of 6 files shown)


include/wil/result.h
```diff
@@ -252,15 +252,17 @@ namespace wil
                 if (ProcessShutdownInProgress())
                 {
                     // There are no other threads to contend with.
-                    if (--m_refCount == 0)
+                    m_refCount = m_refCount - 1;
```
dunhor, line 255: This almost certainly does not need to be `volatile`, but I've opted to keep functionality as-is for now and just address the warnings

tests/cpplatest/CMakeLists.txt
```diff
@@ -12,11 +12,6 @@ if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
     target_compile_definitions(witest.cpplatest PRIVATE _SILENCE_CLANG_COROUTINE_MESSAGE)
 endif()
 
-# Semi-arbitrary insiders SDK version selected that uses C++/WinRT "2.0"
```
dunhor, line 15: No longer necessary now that we're manually running cppwinrt. Just forgot to remove this as a part of the previous change.

oldnewthing (approved):

PR #206, opened 2021-08-25
Extend hstring_compare to support std::wstring_view

tusharpm:
This PR expands on the existing support for `std::wstring` parameters in `hstring_compare`:
- Change `.c_str()` to `.data()` to allow `std::wstring_view`.
- Change `.length()` to `.size()` to allow `winrt::hstring` ~~and other ContiguousContainers such as `std::array<wchar_t, N>` or `std::vector<wchar_t>`~~.

**Testing**:
- Add unit-test for `std::wstring_view`.
- Verified no regression for `std::wstring` tests.


dunhor:
Also, I'm not quite convinced that enabling this for `std::array` is such a great idea... I'd expect most declarations of `std::array<wchar_t, N>` to be used as a buffer with an actual string size possibly less than `N`

tusharpm:
I did a quick survey of the usage of hstring_compare, hstring_less and hstring_insensitive_less and didn't find any types defining `c_str` but not `data`. Regardless, I agree that the first option is safer - I will test the change on my feature branch and run a full build for more confidence.

For `std::array`, we have two options:
1. Document the expectation for `size`, and ask callers to construct an appropriate `std::wstring_view` from the buffer if required.
2. Don't document support for it (or `std::vector<wchar_t>`) - the caller can use it as unspecified behavior at their risk.

However, using `size` also enables `winrt::hstring` parameters which otherwise require a `std::wstring_view` conversion in the caller.

dunhor:
> I did a quick survey of the usage of hstring_compare, hstring_less and hstring_insensitive_less and didn't find any types defining `c_str` but not `data`

Keep in mind that this file provides an explicit specialization of `std::less`, so uses of this type may not be explicit.

dunhor:
I had a random thought this morning. We could SFINAE on `is_constructible<T, wchar_t*>` (note the lack of `const`), which would solve the `array`/`vector` problem. This has the possibility of further breaking existing code, but I'd expect most string types to be constructable from a null terminated string.

tusharpm:
Thanks @dunhor. I have added this change and verified the tests run locally.

ChrisGuzak (approved):

raghavthind2005 (approved):

AtariDreams:
Any update on this?

tusharpm:

dunhor (approved):

PR #208, opened 2021-09-15
Add unique_boundary_descriptor and unique_private_namespace

m417z:


include/wil/resource.h
```diff
@@ -2528,6 +2528,23 @@ namespace wil
 
 #if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM)
     typedef unique_any<PSID, decltype(&::FreeSid), ::FreeSid> unique_sid;
+    typedef unique_any_handle_null_only<decltype(&::DeleteBoundaryDescriptor), ::DeleteBoundaryDescriptor> unique_boundary_descriptor;
+
+    namespace details
+    {
+        // Define a templated type in order to satisfy overload resolution below
+        template <ULONG flags>
+        struct private_namespace_helpers
```
sylveon, line 2537: I don't think this needs a templated type. Can't this work?

```cpp
template<ULONG flags>
static void ClosePrivateNamespaceHelper(HANDLE h)
{
    ::ClosePrivateNamespace(h, flags);
}
```

m417z: I based my code on the `FreePoolWithTag` wrapper. I thought that maybe there's a reason for this kind of definition. I can change it as you suggest, it seems to be working too.
https://github.com/microsoft/wil/blob/209ff9c7e30e71bf6d1405557bf78544920f8157/include/wil/resource.h#L6564-L6577

m417z: Done. I also added `WI_NOEXCEPT`. Not sure about other properties such as `inline`, the headers are not very consistent in this regard.

include/wil/resource.h
```diff
@@ -2528,6 +2528,19 @@ namespace wil
...
+    {
+        template<ULONG flags>
+        inline void __stdcall ClosePrivateNamespaceHelper(HANDLE h) WI_NOEXCEPT
+        {
+            ::ClosePrivateNamespace(h, flags);
+        }
+    }
+
+    template <ULONG flags = 0>
+    using unique_private_namespace = unique_any_handle_null_only<decltype(details::ClosePrivateNamespaceHelper<flags>), &details::ClosePrivateNamespaceHelper<flags>>;
```
dunhor, line 2543: The usage of this is a bit odd. If you want the "default", you'd need to declare a variable of the type `unique_private_namespace<>`. Since it appears as though there's only two "useful" calls, it might make sense to have two type alias - e.g. `unique_private_namespace_close` and `unique_private_namespace_destroy` - and then get rid of the template argument.

m417z: I thought about it, and decided to make it parametrized in case more flags are added in the future. I can change it to `unique_private_namespace_close` and `unique_private_namespace_destroy` if you prefer.

dunhor: Idk how likely that would be. Future proofing at the expense of usability is generally not the greatest option. It's probably fine to keep the parameterized alias, but it still seems beneficial to have the two other well-named aliases.

m417z: Added.

dunhor (approved):

PR #221, opened 2021-11-29
Get build working again

dunhor:
A few things broke Clang build. Fixing those


cmake/common_build_flags.cmake
```diff
@@ -42,6 +42,9 @@ if (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang")
     append_cxx_flag("-Wno-self-assign-overloaded")
     append_cxx_flag("-Wno-self-move")
 
+    # C++/WinRT does not declare 'override' in a number of places
+    append_cxx_flag("-Wno-inconsistent-missing-override")
```
sylveon, line 46: Sounds like a cppwinrt issue could be filled.

PR #223, opened 2022-01-02
Make forwarding `unique_any_t` constructor conditionally noexcept

### LINKED ISSUE #191: Should `wil::unique_any_t` constructor be conditionally `noexcept`?
I noticed how the `unique_any_t` constructor is not `noexcept` even when the storage constructor is `noexcept`. Is this intentional, and if so, why? A comment does note "`should not be WI_NOEXCEPT (may forward to a throwing constructor)`", but couldn't the `unique_any_t` constructor be declared as:
```cpp
template <typename arg1, typename... args_t>
explicit unique_any_t(arg1 && first, args_t&&... args) noexcept(noexcept(storage_t(wistd::forward<arg1>(first), wistd::forward<args_t>(args)...))) :
	storage_t(wistd::forward<arg1>(first), wistd::forward<args_t>(args)...)
{ ... }
```

Just one example where it could be `constexpr` is `wil::unique_hmodule`, but I'm not even sure if WIL currently has any throwing constructors - does it?

stevenwdv:
Closes #191

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -273,7 +273,7 @@ namespace wil
 
         // forwarding constructor: forwards all 'explicit' and multi-arg constructors to the base class
         template <typename arg1, typename... args_t>
-        explicit unique_any_t(arg1 && first, args_t&&... args) :  // should not be WI_NOEXCEPT (may forward to a throwing constructor)
+        explicit unique_any_t(arg1 && first, args_t&&... args) noexcept(noexcept(storage_t(wistd::forward<arg1>(first), wistd::forward<args_t>(args)...))) :
             storage_t(wistd::forward<arg1>(first), wistd::forward<args_t>(args)...)
         {
             static_assert(wistd::is_same<typename policy::pointer_access, details::pointer_access_none>::value ||
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -12,6 +12,7 @@
 #include "result_macros.h"
 #include "wistd_functional.h"
 #include "wistd_memory.h"
+#include "wistd_type_traits.h"
 
 #pragma warning(push)
 #pragma warning(disable:26135 26110)    // Missing locking annotation, Caller failing to hold lock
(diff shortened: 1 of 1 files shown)


ChrisGuzak (approved):
looks good to me

dunhor:
This approach does not work if the derived type's constructor is protected. Something like `is_nothrow_constructible` should probably work. E.g. see https://godbolt.org/z/oYxaneecY

stevenwdv:
@dunhor Yup (see #191), but I think you found the solution!

stevenwdv:
I think this should work!

lhecker:
Oh damn, I'm blind. 😅
I just opened a PR with the same goal: https://github.com/microsoft/wil/pull/248

Unfortunately `wistd::is_nothrow_constructible_v` has the same problem as a `noexcept(...)`: It doesn't work if the constructor isn't publicly accessible. My PR adds a few unit tests which fail with your PR.
I solved the problem by simply making the constructors public.

stevenwdv:
@lhecker Hm I guess that is a problem with `wistd::is_nothrow_constructible` as it currently seems to use `noexcept()` internally. It works fine with `std::is_nothrow_constructible`: https://godbolt.org/z/nGq53zsoW
Maybe that underlying problem can be fixed? Possibly hard to make that compatible with all compilers.

lhecker:
> @lhecker Hm I guess that is a problem with `wistd::is_nothrow_constructible` as it currently seems to use `noexcept()` internally. It works fine with `std::is_nothrow_constructible`: [godbolt.org/z/nGq53zsoW](https://godbolt.org/z/nGq53zsoW) Maybe that underlying problem can be fixed? Possibly hard to make that compatible with all compilers.

It _compiles_ fine, but does it determine the correct `noexcept`? https://godbolt.org/z/j86srbdnz

stevenwdv:
@lhecker Oh... (You forgot to make `Base()` actually `noexcept` but even if you do, `std::is_nothrow_constructible_v` returns `false`.) The standard says this about access checking for `std::is_*constructible_v`:

> Access checking is performed as if in a context unrelated to T and any of the Args[.](https://eel.is/c++draft/type.traits#meta.unary.prop-9.sentence-3)
Only the validity of the immediate context of the variable initialization is considered[.](https://eel.is/c++draft/type.traits#meta.unary.prop-9.sentence-4)
> 
> [Note [6](https://eel.is/c++draft/type.traits#meta.unary.prop-note-6): The evaluation of the initialization can result in side effects such as the instantiation of class template specializations and function template specializations, the generation of implicitly-defined functions, and so on[.](https://eel.is/c++draft/type.traits#meta.unary.prop-9.sentence-5)
Such side effects are not in the “immediate context” and can result in the program being ill-formed[.](https://eel.is/c++draft/type.traits#meta.unary.prop-9.sentence-6)
— end note]

https://eel.is/c++draft/type.traits#meta.unary.prop-9

But I'm not sure what it means by 'immediate context', I can't find a definition

PR #225, opened 2022-01-31
Sync with OS

dunhor:


include/wil/com.h
```diff
@@ -1984,6 +1988,140 @@ namespace wil
...
+            auto hr = obj->QueryInterface(IID_PPV_ARGS(&multiQi));
+            if (SUCCEEDED(hr))
+            {
+                hr = multiQi->QueryMultipleInterfaces(ARRAYSIZE(multiQis), multiQis);
+                std::apply([i = 0, &multiQis](auto&... a) mutable
+                {
+                    (a.attach(reinterpret_cast<typename std::remove_reference_t<decltype(a)>::pointer>(multiQis[i++].pItf)), ...);
+                }, resultTuple);
+            }
+            return std::tuple<HRESULT, decltype(resultTuple)>{hr, std::move(resultTuple)};
```
sylveon, line 2045: Why not use std::tuple_cat?

dunhor: I'll assume that the idea was to return a tuple of size 2. Personally, I'd prefer a flat tuple so that it works better with structured bindings, but what's done is done and I'm not going to change the function signature here.

dunhor: I suppose it also makes it easier to implement the throwing/failfast variants this way

dunhor: I'm also being told that @oldnewthing is making changes to this code, but I'm unsure of what those are right now

Issue #232, opened 2022-03-29
get_module_reference_for_thread is undocumented on Wiki


CookiePLMonster:
This useful function seems to be [a part of `win32_helpers.h`](https://github.com/microsoft/wil/blob/23fca640eae099f9120545483ae7d5d0a98a6fa5/include/wil/win32_helpers.h#L525), but lacks any documentation on Wiki:
https://github.com/microsoft/wil/wiki/Win32-helpers#other-helper-functions

It lives just next to `GetModuleInstanceHandle` so it probably should be documented nearby too, although admittedly its naming convention is not too consistent with what is "near" that function. Perhaps `GetModuleReferenceForThread` would've been more fortunate?

Curiously, it **does** have a test:
https://github.com/microsoft/wil/blob/23fca640eae099f9120545483ae7d5d0a98a6fa5/tests/wiTest.cpp#L3554

jonwis:
We're definitely inconsistent about `snake_case` or `PascalCase` for things. I'll take this to update the wiki. Thanks for the reminder!

CookiePLMonster:
Something I just noticed right now - it should probably also be marked with `WI_NOEXCEPT` and not `noexcept` to be more consistent.

jonwis:
Ok, updated the wiki page. This type appears to have a Sharp Edge, in that you can't use it to hold a reference on the DLL while queuing work to start, and you really shouldn't pass it from the initiating thread to the handling thread (because the type might either get leaked on errors to start thread or destroyed and exiting the initiating thread on error.)

Our motivation for the type was like this:

```c++
wil::unique_handle myThread(::CreateThread(, [](auto&&...)
{
    auto dllRef = wil::get_module_reference_for_thread();
    // Do work here
    // On destruction, dllRef calls FreeLibraryAndExitThread for you
}));
```

Where we are in control of the call to CreateThread and the threadproc.  However, there's a timing gap between when CreateThread is called and when the threadproc is called when the hosting DLL could get unloaded. So what's _really_ needed is a movable and armable object, but passing context to `CreateThread` is super hard. And this type probably can't be used with `std::thread` because calling `FreeLibrary` before returning control to the STL would be Bad News.

So we'd need some support for WIL to create threads, that would encapsulate this:

```c++
auto t = wil::create_thread([foo, bar] { /* my proc */ });
```

where the "hidden threadproc" for `create_thread` uses this module reference wrapper.

CookiePLMonster:
[...]

PR #235, opened 2022-04-27
Free-and-exit should be transferable

jonwis:
See #232 for context. Adds a type that takes a module ref from the current thread and holds it while starting up another win32 thread. While `get_module_reference_for_thread` is great if you're already on the new thread, there is a potential for a race between `CreateThread` returning and the new thread being scheduled to execute its first instructions in which the DLL backing the instructions could be unloaded.

### DIFF
--- a/include/wil/win32_helpers.h
+++ b/include/wil/win32_helpers.h
@@ -531,6 +531,105 @@ namespace wil
             FreeLibraryAndExitThread(thisModule, 0);
         });
     }
+
+    struct module_reference_for_new_thread
+    {
+        bool free_on_exit = false;
+        HMODULE module_ref = nullptr;
+
+        module_reference_for_new_thread()
+        {
+            FAIL_FAST_IF(!::GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS, L"", &module_ref));
+        }
+
+        ~module_reference_for_new_thread()
+        {
+            reset();
+        }
+
+        module_reference_for_new_thread(module_reference_for_new_thread&& other)
+        {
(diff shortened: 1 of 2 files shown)


include/wil/win32_helpers.h
```diff
@@ -531,6 +531,105 @@ namespace wil
...
+        }
+
+        ~module_reference_for_new_thread()
+        {
+            reset();
+        }
+
+        module_reference_for_new_thread(module_reference_for_new_thread&& other)
+        {
+            wistd::exchange(free_on_exit, other.free_on_exit);
```
dunhor, line 552: Unless I'm going crazy, you want these to be `swap`. `exchange` just assigns the new value without setting/clearing the assigned-from value. I.e. `other` will still have its destructor run and you'll have a double free of `module_ref`

include/wil/win32_helpers.h
```diff
@@ -531,6 +531,105 @@ namespace wil
...
+            wistd::exchange(module_ref, other.module_ref);
+        }
+
+        module_reference_for_new_thread& operator=(module_reference_for_new_thread&& other)
+        {
+            if (this != wistd::addressof(other))
+            {
+                reset();
```
dunhor, line 560: nit: Assuming you change to `swap`, calling `reset` here is not necessary. Moved-from objects are specified as being in a "valid but unspecified state", so giving the moved-from object the current object's resources to free in its destructor is a relatively common pattern

include/wil/win32_helpers.h
dunhor: Or perhaps you meant `free_on_exit = wistd::exchange(other.free_on_exit, false)` etc?

PR #239, opened 2022-05-05
Hide QueryUnbiasedInterruptTimeAs100ns and QueryUnbiasedInterruptTimeAsMSec when not targetting Win7

CookiePLMonster:
When targetting `_WIN32_WINNT = 0x0600`, an unconditional reliance on `QueryUnbiasedInterruptTime` prevented `wil/win32_helpers.h` from compiling. This change modifies the ifdef to mirror `realtimeapiset.h`.

Since pre-Win7 targets are technically unsupported, no tests or documentation have been updated.

### DIFF
--- a/include/wil/win32_helpers.h
+++ b/include/wil/win32_helpers.h
@@ -120,7 +120,7 @@ namespace wil
             return timeMsec * filetime_duration::one_millisecond;
         }
 
-#if defined(_APISETREALTIME_)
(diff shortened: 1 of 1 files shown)


jonwis:
Hmm, the `tests\win7` subdirectory should not be compiling, then.  Investigating that as well...

jonwis:
> Hmm, the `tests\win7` subdirectory should not be compiling, then. Investigating that as well...

Drat, I misread. You're targeting _pre-win7_ ... the fix is fine.

jonwis (approved):

PR #240, opened 2022-05-14
wil::to_array for bulk fetching of vector and iterable types

jonwis:
Provides a "bulk convert collection to vector" to reduce calls when the caller intends to process the entire content of the collection, such as in a loop.  For example:

```c++
winrt::IVector<winrt::IPropertyValue> values = /* ... */
uint32_t total = 0;
for (auto&& i : values) {
    total += i.GetUInt32();
}
```

The range-for initialization uses `Size()` once.  Each step of the range-for calls `GetAt(idx)` to retrieve the next type. Even for small collections, the cost of cross-apartment calls can be significant.

The `wil::to_vector` method instead uses the `GetMany()` method available on `IVector[View]` and other collection types to retrieve the full content of the container.  When used with `IMap` or `IIterable`, it uses `IIterator::GetMany` in blocks of 64 elements to retrieve the full content.

### DIFF
--- a/include/wil/cppwinrt_helpers.h
+++ b/include/wil/cppwinrt_helpers.h
@@ -212,12 +212,14 @@ namespace wil::details
     };
 }
 #endif // __WIL_CPPWINRT_MICROSOFT_UI_DISPATCHING_HELPERS
+/// @endcond
 
 #if defined(WINRT_Windows_Foundation_Collections_H) && !defined(__WIL_CPPWINRT_WINDOWS_FOUNDATION_COLLECTION_HELPERS)
 #define __WIL_CPPWINRT_WINDOWS_FOUNDATION_COLLECTION_HELPERS
 
 namespace wil
 {
+    /// @cond
     namespace details
     {
         template<typename TData, typename TSrc> std::vector<TData> to_vector_impl(TSrc&& fetcher)
@@ -226,12 +228,11 @@ namespace wil
             const uint32_t chunkSize = 64;
             while (true)
             {
-                uint32_t offset = output.size();
-                output.resize(output.size() + chunkSize, winrt::impl::empty_value<TData>());
-                auto fetched = fetcher(offset, winrt::array_view<TData>{ output.data() + offset, output.data() + output.size() });
+                TData nextChunk[chunkSize]{ winrt::impl::empty_value<TData>() };
+                auto fetched = fetcher(static_cast<uint32_t>(output.size()), nextChunk);
+                output.insert(output.end(), std::make_move_iterator(std::begin(nextChunk)), std::make_move_iterator(std::begin(nextChunk) + fetched));
                 if (fetched < chunkSize)
                 {
-                    output.resize(offset + fetched);
                     break;
                 }
             }
@@ -244,7 +245,22 @@ namespace wil
         template<typename T> struct is_iterator : std::bool_constant<false> {};
         template<typename Q> struct is_iterator<winrt::Windows::Foundation::Collections::IIterator<Q>> : std::bool_constant<true> {};
     }
+    /// @endcond
 
+    /** Converts C++ / WinRT vectors, iterators, and iterables to std::vector by requesting
+    the collection's data in chunks. This can be more efficient in terms of IPC cost than
+    iteratively processing the collection.
+    ~~~
+    winrt::IVector<winrt::hstring> collection = GetCollection();
+    std::vector<winrt::hstring> allData = wil::to_vector(collection); // read all data from collection
+    for (winrt::hstring const& item : allData)
+    {
+        // use item
+    }
+    ~~~
+    Can be used for IVector<T>, IVectorView<T>, IIterable<T>, IIterator<T>, and any type or
+    interface that C++/WinRT projects those interfaces for (PropertySet, IMap<T,K>, etc.)
+    */ 
     template<typename TSrc> auto to_vector(TSrc const& src)
     {
         if constexpr (details::is_vector<TSrc>())
@@ -264,4 +280,3 @@ namespace wil
 
 #endif
 
-/// @endcond
--- a/include/wil/cppwinrt_helpers.h
+++ b/include/wil/cppwinrt_helpers.h
@@ -276,7 +276,7 @@ namespace wil
         {
             using T = decltype(src.Current());
             std::vector<T> result;
-            const uint32_t chunkSize = 5;
+            const uint32_t chunkSize = 64;
             while (true)
             {
                 auto const lastSize = result.size();
--- a/include/wil/cppwinrt_helpers.h
+++ b/include/wil/cppwinrt_helpers.h
@@ -240,8 +240,8 @@ namespace wil
     }
     /// @endcond
 
-    /** Converts C++ / WinRT vectors, iterators, and iterables to std::vector by requesting
-    the collection's data in chunks. This can be more efficient in terms of IPC cost than
+    /** Converts C++ / WinRT vectors, iterators, and iterables to std::vector by requesting the
+    collection's data a single request. This can be more efficient in terms of IPC cost than
     iteratively processing the collection.
     ~~~
     winrt::IVector<winrt::hstring> collection = GetCollection();
@@ -253,6 +253,8 @@ namespace wil
     ~~~
     Can be used for IVector<T>, IVectorView<T>, IIterable<T>, IIterator<T>, and any type or
     interface that C++/WinRT projects those interfaces for (PropertySet, IMap<T,K>, etc.)
+    Iterable-only types fetch content in units of 64. When used with an iterator, the returned
+    vector contains the iterator's current position and any others after it.
     */ 
     template<typename TSrc> auto to_vector(TSrc const& src)
     {
--- a/include/wil/cppwinrt_helpers.h
+++ b/include/wil/cppwinrt_helpers.h
@@ -220,15 +220,15 @@ namespace wil
 {
     namespace details
     {
-        template<typename TData, typename TSrc, uint32_t chunkSize = 64> std::vector<TData> to_vector_impl(TSrc const& src)
+        template<typename TData, typename TSrc> std::vector<TData> to_vector_impl(TSrc&& fetcher)
         {
-            static_assert(chunkSize > 0);
             std::vector<TData> output;
             uint32_t offset = 0;
+            const uint32_t chunkSize = 64;
             while (true)
             {
                 output.resize(output.size() + chunkSize, winrt::impl::empty_value<TData>());
-                auto fetched = src.GetMany(offset, { output.data() + offset, output.data() + output.size() });
+                auto fetched = fetcher(offset, winrt::array_view<TData>{ output.data() + offset, output.data() + output.size() });
                 if (fetched < chunkSize)
                 {
                     output.resize(offset + fetched);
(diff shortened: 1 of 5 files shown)


include/wil/cppwinrt_helpers.h
```diff
@@ -212,5 +212,71 @@ namespace wil::details
...
+    {
+        // use item
+    }
+    ~~~
+    Can be used for IVector<T>, IVectorView<T>, IIterable<T>, IIterator<T>, and any type or
+    interface that C++/WinRT projects those interfaces for (PropertySet, IMap<T,K>, etc.)
+    */ 
+    template<typename TSrc> auto to_vector(TSrc const& src)
+    {
+        if constexpr (details::is_vector<TSrc>())
```
jonwis, line 266: Feels like this could be more elegant with an sfinae doodad.  There is winrt::impl::has_GetAt or similar that the projections use to generate range iterators.  Feels bad enough that I'm reaching into winrt::impl above for the default-value construct.

include/wil/winrt.h
```diff
@@ -46,6 +46,12 @@ namespace std
 #pragma warning(pop)
 #endif
 #endif
+#if defined(WIL_ENABLE_EXCEPTIONS) && defined(__has_include)
+#if __has_include(<vector>)
+#define __WI_HAS_STD_VECTOR 1
```
jonwis, line 51: Is this good? Is it OK to assume that if you have vector and exceptions turned on, you're OK with bringing in here?

include/wil/cppwinrt_helpers.h
```diff
@@ -212,5 +212,71 @@ namespace wil::details
...
+    /// @cond
+    namespace details
+    {
+        template<typename TData, typename TSrc> std::vector<TData> to_vector_impl(TSrc&& fetcher)
+        {
+            std::vector<TData> output;
+            const uint32_t chunkSize = 64;
+            while (true)
+            {
+                TData nextChunk[chunkSize]{ winrt::impl::empty_value<TData>() };
```
jonwis, line 231: I had a previous model that fetched data right into the vector, but that required an initial allocation on the storage even if the underlying IVector was zero-length.

This model causes a move from the local storage into an allocator, but since these are all scalar or pointer-sized, maybe that's OK?  I'm pretending that `vector<scalar>.insert` could have some optimization that blits from the input array to the output.

Not sure there's a good way around this.  (You can see the previous model in an earlier commit...)

include/wil/winrt.h
```diff
@@ -994,6 +1000,42 @@ namespace wil
...
+    }
+    Can be used for ABI::Windows::Foundation::Collections::IVector<T> and
+    ABI::Windows::Foundation::Collections::IVectorView<T>
+    */
+    template<typename VectorType> auto to_vector(VectorType* src)
+    {
+        using TResult = typename details::MapVectorResultType<VectorType>::type;
+        using TSmart = typename details::MapToSmartType<TResult>::type;
+        static_assert(sizeof(TResult) == sizeof(TSmart), "result and smart sizes are different");
+        std::vector<TSmart> output;
```
jonwis, line 1022: Sadly there's (not yet) a vector-like wrapper for the wil::unique_foo_ptr<T[]> types that supports resize/enlarge/append. Making the nothrow non-stl version of this was going to be a lot of code.  I'm open to writing that, though, if folks think it's useful.

jonwis:
Some motivation on this - using WPA, I noticed some OS components making many cross-apartment/cross-process calls to iterate vectors:

```c++
UINT32 size = 0;
ComPtr<IVector<IFoo*>> vec;
RETURN_IF_FAILED(remoteType->GetFoos(&vec));
RETURN_IF_FAILED(vec->get_Size(&size));
for (UINT32 i = 0; i < size; ++i)
{
    ComPtr<IFoo> nextFoo;
    RETURN_IF_FAILED(vec->GetAt(i, &nextFoo));
    RETURN_IF_FAILED(nextFoo->Bar());
}
```

There's a fixed overhead to each IPC call as the flow transits from the app->COM->RPC->ALPC->RPC->COM->implementation. For a vector with 10-15 HSTRING members, that overhead dominates the time.

We now have the technology to avoid this somewhat by returning conformant arrays for things, but it's unevenly distributed in the OS codebase.  API design rules now permit this:

```idl
runtimeclass RemoteType {
    Foo[] GetFoos();
}
```

And then consumption wise, in C++/WinRT:

```c++
auto foos = remoteType.GetFoos(); // winrt::com_array<Foo>
for (auto&& foo : foos) {
    foo.Bar();
}
```

oldnewthing:
This may succeed but return erroneous results if the collection changes between successive chunks.

Here's a fixed version that is also non-chunky. Costs an extra RPC in the case where the vector has fewer than 64 elements.
```cpp
std::vector<T> to_vector(TSrc const& source)
{
    auto expected = source.Size();
    std::vector<T> result(expected + 1, empty<T>());
    if (source.GetMany(0, result) <= expected) return result;
    throw hresult_changed_state();
}
```

I blogged [a simplified version of this](https://devblogs.microsoft.com/oldnewthing/20191122-00/?p=103123/) and left this full version as an exercise.

include/wil/cppwinrt_helpers.h
jonwis: Using the "get everything in at most two calls" pattern now.

jonwis:
> I blogged [a simplified version of this](https://devblogs.microsoft.com/oldnewthing/20191122-00/?p=103123/) and left this full version as an exercise.

Picked that one up instead. A little squicky about the `.data() + base + newsize` math for the iterator, but tests pass.

include/wil/winrt.h
jonwis: Hmm, see also #242 that is probably a prerequisite to making this one easy. Neither `wistd::unique_ptr<T[]>` (no length member) nor `wil::unique_cotaskmem_array_ptr<T>` (no call to dtor) are directly suitable.

include/wil/cppwinrt_helpers.h
```diff
@@ -212,5 +212,88 @@ namespace wil::details
...
+            else
+            {
+                return {};
+            }
+        }
+    }
+    /// @endcond
+
+    /** Converts C++ / WinRT vectors, iterators, and iterables to std::vector by requesting
+    the collection's data in chunks. This can be more efficient in terms of IPC cost than
```
jonwis, line 244: No longer in chunks, though. "In a single request."

ChrisGuzak (approved):

include/wil/winrt.h
ChrisGuzak: I think so

include/wil/cppwinrt_helpers.h
```diff
@@ -212,5 +212,90 @@ namespace wil::details
...
+namespace wil
+{
+    /// @cond
+    namespace details
+    {
+        template<typename T> struct is_vector : std::bool_constant<false> {};
+        template<typename Q> struct is_vector<winrt::Windows::Foundation::Collections::IVector<Q>> : std::bool_constant<true> {};
+        template<typename Q> struct is_vector<winrt::Windows::Foundation::Collections::IVectorView<Q>> : std::bool_constant<true> {};
+        template<typename T> struct is_iterator : std::bool_constant<false> {};
+        template<typename Q> struct is_iterator<winrt::Windows::Foundation::Collections::IIterator<Q>> : std::bool_constant<true> {};
```
oldnewthing, line 228: Should use duck typing so we will work with runtime classes which "quack like" `IVector` or `IVectorView` (`AccessListEntryView`, `DeviceInformationCollection`, `JsonArray`, etc.) without having to fall back to `IIterator`.

Also use duck typing to detect `IIterator` support, so we can support `ResourceMapIterator` and other runtime classes which implement `IIterator`.

```cpp
namespace details
{
    template<typename T>
    struct is_winrt_vector_like {
    private:
        template <typename U,
            typename = decltype(std::declval<U>().GetMany(std::declval<U>().Size(),
                winrt::array_view<decltype(std::declval<U>().GetAt(0))>{}))>
        static constexpr bool get_value(int) { return true; }
        template <typename> static constexpr bool get_value(...) { return false; }  
    public:
        static constexpr bool value = get_value<T>(0);                                                                                                                         
    };

    template<typename T>
    struct is_winrt_iterator_like {
    private:
        template <typename U,
            typename = decltype(std::declval<U>().GetMany(0,
                winrt::array_view<decltype(std::declval<U>().Current())>{}))>
        static constexpr bool get_value(int) { return true; }
        template <typename> static constexpr bool get_value(...) { return false; }  
    public:
        static constexpr bool value = get_value<T>(0);                                                                                                                         
    };
}
```

For `is_winrt_vector_like`, we need `Size()` and `GetMany()`, so I use the expression `o.GetMany(o.Size(), array_view<Elem>{})` where `Elem` is the element type returned by `GetAt()`. Couldn't see an easy way to infer the element type from GetMany, so we end up requiring `GetAt()` too, even though we don't call it.

The `is_winrt_iterator_like` looks just for `GetMany()`, and this time we use `Current()` to infer the element type.

One oddity is that calling a method only from SFINAE sometimes triggers a "Method is declared but never called" warning from MSVC. I think there's an open MSVC bug on it.

include/wil/cppwinrt_helpers.h
```diff
@@ -212,5 +212,90 @@ namespace wil::details
...
+            else
+            {
+                return {};
+            }
+        }
+    }
+    /// @endcond
+
+    /** Converts C++ / WinRT vectors, iterators, and iterables to std::vector by requesting the
+    collection's data a single request. This can be more efficient in terms of IPC cost than
```
oldnewthing, line 244: TYPO: "the collections' data *in* a single request."

Except that the IVector case makes two requests, and IIterator can make many. So maybe say "the collection's data in bulk."

include/wil/cppwinrt_helpers.h
```diff
@@ -212,5 +212,90 @@ namespace wil::details
...
+    {
+        if constexpr (details::is_vector<TSrc>())
+        {
+            using T = decltype(src.GetAt(0));
+            std::vector<T> result;
+            if (auto expected = src.Size())
+            {
+                result.resize(expected + 1, details::empty<T>());
+                auto actual = src.GetMany(0, result);
+                if (expected > actual)
```
oldnewthing, line 269: Test is backward. Should be
```cpp
if (actual > expected)
```

The idea is that we ask how many elements there are and ask for that many *plus one*.

If we get too many (`actual > expected`), then that means that the vector grew while we were capturing it, and we weren't able to capture all of them atomically. Bad. Throw the concurrency exception. (Alternatively, loop back and try again.)

If we get exactly right or too few, then we know that we got the whole thing atomically, and all is good.

We have to ask for "plus one" to detect the "vector grew while we were capturing it" case.

include/wil/cppwinrt_helpers.h
```diff
@@ -212,5 +212,90 @@ namespace wil::details
...
+                }
+                result.resize(actual);
+            }
+            return result;
+        }
+        else if constexpr (details::is_iterator<TSrc>())
+        {
+            using T = decltype(src.Current());
+            std::vector<T> result;
+            const uint32_t chunkSize = 64;
```
oldnewthing, line 281:
```cpp
constexpr uint32_t chunkSize = 64;
```

tests/CppWinRTTests.cpp
```diff
@@ -26,6 +27,79 @@ static const HRESULT cppwinrt_mapped_hresults[] =
...
+template<typename T, typename K> 
+void CheckMapVector(std::vector<winrt::Windows::Foundation::Collections::IKeyValuePair<T, K>> const& test, std::map<T, K> const& src)
+{
+    REQUIRE(test.size() == src.size());
+    for (auto&& i : test)
+    {
+        REQUIRE(i.Value() == src.at(i.Key()));
+    }
+}
+
```
oldnewthing, line 44: Maybe add cases for things like `DeviceInformationCollection` and `ResourceMapIterator`.

```cpp
struct vectorLike : implements<vectorLike, IVectorView<int>>
{
    auto Size() { return 0; }
    int GetAt(uint32_t) { throw hresult_out_of_bounds() }
    uint32_t GetMany(uint32_t start, array_view<int>) { if (start > 0) throw hresult_out_of_bounds(); return 0; }
    bool IndexOf(int, uint32_t) { throw hresult_not_implemented(); }
};
```

PR #244, opened 2022-05-19
Apply PR feedback for duck-typing of to_vector

jonwis:
@oldnewthing had a great suggestion to duck-type the to_vector helper from #240 but I was too hasty.

Along the way, discovered that the test crash was due to an interaction between the malloc spy and leaking an apartment init, so added explicit init/shutdown that uses it.

And suddenly, LLVM is complaining that cppwinrt's base.h uses `_InterlockedCompareExchange128` which needs their `-mcx16` option, but which wasn't included in our common build flags.

### DIFF
--- a/include/wil/cppwinrt_helpers.h
+++ b/include/wil/cppwinrt_helpers.h
@@ -221,11 +221,27 @@ namespace wil
     /// @cond
     namespace details
     {
-        template<typename T> struct is_vector : std::bool_constant<false> {};
-        template<typename Q> struct is_vector<winrt::Windows::Foundation::Collections::IVector<Q>> : std::bool_constant<true> {};
-        template<typename Q> struct is_vector<winrt::Windows::Foundation::Collections::IVectorView<Q>> : std::bool_constant<true> {};
-        template<typename T> struct is_iterator : std::bool_constant<false> {};
-        template<typename Q> struct is_iterator<winrt::Windows::Foundation::Collections::IIterator<Q>> : std::bool_constant<true> {};
+        template<typename T> struct is_winrt_vector_like {
+        private:
+            template <typename U,
+                typename = decltype(std::declval<U>().GetMany(std::declval<U>().Size(),
+                    winrt::array_view<decltype(std::declval<U>().GetAt(0))>{}))>
+            static constexpr bool get_value(int) { return true; }
+            template <typename> static constexpr bool get_value(...) { return false; }  
+        public:
+            static constexpr bool value = get_value<T>(0);                                                                                                                         
+        };
+
+        template<typename T> struct is_winrt_iterator_like {
+        private:
+            template <typename U,
+                typename = decltype(std::declval<U>().GetMany(winrt::array_view<decltype(std::declval<U>().Current())>{}))>
+            static constexpr bool get_value(int) { return true; }
+            template <typename> static constexpr bool get_value(...) { return false; }  
+        public:
+            static constexpr bool value = get_value<T>(0);                                                                                                                         
+        };
+
         template<typename T> constexpr T empty() noexcept
         {
             if constexpr (std::is_base_of_v<winrt::Windows::Foundation::IUnknown, T>)
@@ -241,8 +257,8 @@ namespace wil
     /// @endcond
 
     /** Converts C++ / WinRT vectors, iterators, and iterables to std::vector by requesting the
-    collection's data a single request. This can be more efficient in terms of IPC cost than
-    iteratively processing the collection.
+    collection's data in bulk. This can be more efficient in terms of IPC cost than iteratively
+    processing the collection.
     ~~~
     winrt::IVector<winrt::hstring> collection = GetCollection();
     std::vector<winrt::hstring> allData = wil::to_vector(collection); // read all data from collection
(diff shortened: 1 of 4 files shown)


tests/CppWinRTTests.cpp
```diff
@@ -42,8 +42,52 @@ void CheckMapVector(std::vector<winrt::Windows::Foundation::Collections::IKeyVal
     }
 }
 
+struct vector_like
+{
+    uint32_t Size() const { return 100; }
+    int GetAt(uint32_t) const { return 15; }
+    uint32_t GetMany(uint32_t start, winrt::array_view<int> items) const
+    { 
+        if (start > 0) throw winrt::hresult_out_of_bounds();
```
DrusTheAxe, line 51: if 1-liners?
if without `{}`?

Shouldn't that be

```
if (start > 0)
{
    throw winrt::hresult_out_of_bounds();
}
```

?

tests/CppWinRTTests.cpp
```diff
@@ -42,8 +42,52 @@ void CheckMapVector(std::vector<winrt::Windows::Foundation::Collections::IKeyVal
...
+struct iterable_like
+{
+    auto First() const { return iterator_like{}; }
+};
+
+struct unstable_vector : winrt::implements<unstable_vector, winrt::Windows::Foundation::Collections::IVectorView<int>>
+{
+    auto Size() { return 4; }
+    int GetAt(uint32_t) { return 7; }
+    uint32_t GetMany(uint32_t, winrt::array_view<int> items) {
```
DrusTheAxe, line 81: NIT: Open brace on same line, not next line?

DrusTheAxe, line 81: Size, GetAt, GetMany and IndexOf aren't `const` methods?

Because their inherited and not const in the parent (unstable_vector?)

DrusTheAxe (approved):

tests/CppWinRTTests.cpp
jonwis: This is the implementation of IVector for C++/WinRT. _Inside_ the implementation, C++/WinRT calls the non-const version of the type.  The _projection_ of IVector<T> is all const methods.

include/wil/cppwinrt_helpers.h
```diff
@@ -221,11 +221,27 @@ namespace wil
...
+        template<typename T> struct is_winrt_vector_like {
+        private:
+            template <typename U,
+                typename = decltype(std::declval<U>().GetMany(std::declval<U>().Size(),
+                    winrt::array_view<decltype(std::declval<U>().GetAt(0))>{}))>
+            static constexpr bool get_value(int) { return true; }
+            template <typename> static constexpr bool get_value(...) { return false; }  
+        public:
+            static constexpr bool value = get_value<T>(0);                                                                                                                         
+        };
```
oldnewthing, line 233: For future reference, here's a version that supports customizing the type of the vector and the allocator. We probably don't need it, but just saving here for posterity.

```cpp
        template<typename T, typename ElementType>
        class is_winrt_vector_like
        {
            template <typename U, typename = decltype(std::declval<U>().GetMany(0, winrt::array_view<ElementType>{})) >
            static constexpr bool get_value(int) { return true; }
            template <typename>
            static constexpr bool get_value(...) { return false; }

        public:
            static constexpr bool value = get_value<T>(0);
        };
    }

    template<typename ElementType = void, typename Allocator = void, typename T,
        typename ActualElementType = std::conditional_t<std::is_same_v<ElementType, void>, std::decay_t<decltype(std::declval<T>().GetAt(0))>, ElementType>,
        typename ActualAllocator = std::conditional_t<std::is_same_v<Allocator, void>, std::allocator<ActualElementType>, Allocator>,
        typename std::enable_if_t<details::is_winrt_vector_like<T, ActualElementType>::value, int> = 0>
        auto to_std_vector(T const& v, ActualAllocator const& alloc = ActualAllocator())
    {
        std::vector<ActualElementType, ActualAllocator> result(alloc);
        auto size = v.Size();
        if (size > 0)
        {
            result.resize(size + 1, winrt_empty_value<ActualElementType>());
            auto actual = v.GetMany(0, result);
            if (actual > size)
            {
                throw winrt::hresult_changed_state();
            }
            result.erase(result.begin() + actual, result.end());
        }
        return result;
    }
```

The weirdness with the template types is to allow you to write `to_std_vector<MyType, MyAllocator>(winrtVector)`.

oldnewthing (approved):

ChrisGuzak (approved):

include/wil/cppwinrt_helpers.h
```diff
@@ -258,27 +274,27 @@ namespace wil
...
+        if constexpr (details::is_winrt_vector_like<TSrc>::value)
         {
             using T = decltype(src.GetAt(0));
             std::vector<T> result;
             if (auto expected = src.Size())
             {
                 result.resize(expected + 1, details::empty<T>());
                 auto actual = src.GetMany(0, result);
-                if (expected > actual)
+                if (actual > expected)
```
dunhor, line 285: Why not `actual != expected`? As-is, it allows fewer values to be returned

oldnewthing: If `actual < expected`, then it means that the vector shrunk while we were trying to capture it, but the capture was still successful: We got all the elements atomically.

PR #248, opened 2022-06-11
Allow unique/shared_any_t constructor to be noexcept

### LINKED ISSUE #191: Should `wil::unique_any_t` constructor be conditionally `noexcept`?

lhecker:
This commit allows the use of most `wil::unique_*` types in `noexcept` contexts.
Unfortunately the use of inheritance made the modification slightly more
difficult and required the `unique_storage` constructor to be made public.
But thanks to inheritance those constructors are hidden from public use
as the explicit `unique_any_t` constructors "override" the base class ones.

The `shared_any_t` (and `shared_storage`) classes can't be `noexcept`
at the time of writing as they use a `std::shared_ptr` internally
which can throw during construction. But the classes were modified anyways
just in case other kinds of (non-throwing) sharing are added in the future.

Closes #191

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -105,6 +105,7 @@ namespace wil
             typedef typename policy::pointer pointer;
             typedef unique_storage<policy> base_storage;
 
+        public:
             unique_storage() WI_NOEXCEPT :
             m_ptr(policy::invalid_value())
             {
@@ -129,13 +130,6 @@ namespace wil
                 }
             }
 
-            void replace(unique_storage &&other) WI_NOEXCEPT
-            {
-                reset(other.m_ptr);
-                other.m_ptr = policy::invalid_value();
-            }
-
-        public:
             bool is_valid() const WI_NOEXCEPT
             {
                 return policy::is_valid(m_ptr);
@@ -175,6 +169,13 @@ namespace wil
                 return &m_ptr;
             }
 
+        protected:
+            void replace(unique_storage &&other) WI_NOEXCEPT
+            {
+                reset(other.m_ptr);
+                other.m_ptr = policy::invalid_value();
+            }
+
         private:
             pointer_storage m_ptr;
         };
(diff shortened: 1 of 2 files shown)


tests/ResourceTests.cpp
```diff
@@ -821,10 +851,10 @@ TEST_CASE("ComTokenDirectCloser", "[resource]")
 
 TEST_CASE("UniqueCloseClipboardCall", "[resource]")
 {
-#if defined(__WIL__WINUSER_) && !defined(NOCLIPBOARD)    
+#if defined(__WIL__WINUSER_) && !defined(NOCLIPBOARD)
```
DHowett, line 854: unrelated -- may want to back this out?

include/wil/resource.h
```diff
@@ -105,6 +105,7 @@ namespace wil
             typedef typename policy::pointer pointer;
             typedef unique_storage<policy> base_storage;
 
+        public:
```
jonwis, line 108: Chat indicates that this is required bc the "is nothrow constructable" check requires this type to be, well, constructable. Maybe to retain the whole "don't use this directly, use something else" retain the dtor as protected?

tests/ResourceTests.cpp
```diff
@@ -276,6 +276,36 @@ void UniqueProcessInfo()
...
+
+    struct ProtectedConstructor : BaseStorage
+    {
+    protected:
+        explicit ProtectedConstructor(HANDLE) WI_NOEXCEPT {}
+    };
+
+    // wil::unique_handle is one of the many types which are expected to be noexcept
+    // constructible since they don't perform any "advanced" initialization.
+    static_assert(wistd::is_nothrow_constructible_v<wil::unique_handle, HANDLE>, "wil::unique_any_t should be noexcept if the storage is");
```
jonwis, line 298: Maybe also assert that it's nothrow constructable with no parameters? Just for completeness.

jonwis (approved):

tests/ResourceTests.cpp
lhecker: Yeah I wasn't sure if I should keep it. Other files all have a trailing newline and stripped whitespace, so I kept it for my initial submission, but I can definitely remove it if anyone would like to. (Unless you meant just that.)

lhecker: We talked outside of GitHub, and I've decided to keep this change at the moment until anyone else chimes in.

include/wil/resource.h
lhecker: As it turns out, if you make the destructor protected, `is_nothrow_constructible_v` doesn't work either. Both constructor and destructor have to be public, so that they can be used in the "forein", non-friend `is_nothrow_constructible` template.

PR #250, opened 2022-06-23
fix SAL annotation

NoelCross:
The static analysis engine in VS2022 is flagging the _In_opt_ annotation for a DWORD as being invalid.

private\external\wil\include\wil\resource.h | 2429 | 6553 |   | The annotation for function 'SetThreadpoolTimer' on _Param_(4) does not apply to a value type.

Switching this annotation to _In_ instead.

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -2361,7 +2361,7 @@ namespace wil
         // Non-RTL implementation for threadpool_t parameter of DestroyThreadPoolTimer<>
         struct SystemThreadPoolMethods
         {
-            static void WINAPI SetThreadpoolTimer(_Inout_ PTP_TIMER Timer, _In_opt_ PFILETIME DueTime, _In_ DWORD Period, _In_opt_ DWORD WindowLength) WI_NOEXCEPT
+            static void WINAPI SetThreadpoolTimer(_Inout_ PTP_TIMER Timer, _In_opt_ PFILETIME DueTime, _In_ DWORD Period, _In_ DWORD WindowLength) WI_NOEXCEPT
             {
                 ::SetThreadpoolTimer(Timer, DueTime, Period, WindowLength);
             }


sylveon:
FWIW: the Windows headers use the same annotation
```cpp
WINBASEAPI
VOID
WINAPI
SetThreadpoolTimer(
    _Inout_ PTP_TIMER pti,
    _In_opt_ PFILETIME pftDueTime,
    _In_ DWORD msPeriod,
    _In_opt_ DWORD msWindowLength
    );
```

Might need to figure out who's right

dmachaj:
[SetThreadpoolTimer](https://docs.microsoft.com/en-us/windows/win32/api/threadpoolapiset/nf-threadpoolapiset-setthreadpooltimer)'s fourth parameter is publicly documented as "optional" on docs.microsoft.com.  However, that seems like a strange annotation for a DWORD.  [`_In_opt_`](https://docs.microsoft.com/en-us/cpp/code-quality/understanding-sal?view=msvc-170#example-the-_in_opt_-annotation) is documented such that "optional" means "can be NULL".  `NULL` is a pointer concept so this is nonsensical for a DWORD.  

Perhaps this is a legacy of `NULL` being defined as `0` for a long time, and that really the annotation is trying to say "can be 0".  Which is of course always true for a DWORD anyway.

jonwis:
> @oldnewthing @jonwis I'm not sure how to add reviewers for this PR via the github UX so attempting to get your attention via PR comments.

Over here:

image

It's on the right side of the PR creation thingy, click the "Reviewers" heading and add folks.

jonwis:
Thanks @sylveon.  @ChrisGuzak we were having a similar conversation about SAL. In Windows, our position is that SAL isn't necessary for things that the compiler or definition already expresses.  SAL on scalars is allowed but not required - see https://docs.microsoft.com/cpp/code-quality/annotating-function-parameters-and-return-values where it's "explicitly allowed" but not "required."

The wording on https://docs.microsoft.com/windows/win32/api/threadpoolapiset/nf-threadpoolapiset-setthreadpooltimer is a little confusing; while it's "optional" to add a window duration, "no window duration" is the same as saying "zero length window duration" so it's not _really_ optional?

The change looks fine, and we'll take a note to look at the SAL on the PSDK types in the future for clarity.

include/wil/resource.h
```diff
@@ -2361,7 +2361,7 @@ namespace wil
         // Non-RTL implementation for threadpool_t parameter of DestroyThreadPoolTimer<>
         struct SystemThreadPoolMethods
         {
-            static void WINAPI SetThreadpoolTimer(_Inout_ PTP_TIMER Timer, _In_opt_ PFILETIME DueTime, _In_ DWORD Period, _In_opt_ DWORD WindowLength) WI_NOEXCEPT
+            static void WINAPI SetThreadpoolTimer(_Inout_ PTP_TIMER Timer, _In_opt_ PFILETIME DueTime, _In_ DWORD Period, _In_ DWORD WindowLength) WI_NOEXCEPT
```
jonwis, line 2364: Oh drat, the casing on these parameters is unfortunate. But maybe that's for a later change. We prefer `camelCase` for variable and parameter names.

PR #252, opened 2022-06-30
Allow basic_zstring_view to be constructed from noncopyable custom string types

talagrand:
basic_zstring_view's constructor should use perfect forwarding, otherwise it copies its argument. This fails on non-copyable string types. Ask me how I know.


jonwis:
> basic_zstring_view's constructor should use perfect forwarding, otherwise it copies its argument. This fails on non-copyable string types. Ask me how I know.

"How do you know?" :D

jonwis (approved):

PR #254, opened 2022-08-04
Disable test under clang to unblock tests for now

dunhor:
Clang is having issues with coroutines. Need to investigate, but disabling for now to unblock


dunhor:
Filed #255 to track

PR #257, opened 2022-08-26
RtlDisown... is now publicly documented; use it if available

### LINKED ISSUE #256: Enable use of RtlDisownModuleHeapAllocation when available
https://github.com/microsoft/wil/blob/fd91a817faeec8dee12558b40a6ee54a863d24a7/include/wil/result_macros.h#L1301-L1310

This API is now publicly documented - https://docs.microsoft.com/en-us/windows/win32/DevNotes/rtlsupportapi/nf-rtlsupportapi-rtldisownmoduleheapallocation

Move the setting of the function pointer to the global init path when available so apps using appverifier and wil no longer get a "leak" warning.

jonwis:
This resolves #256 and eliminates AppVerifier-discovered "leaks" in WIL

### DIFF
--- a/include/wil/result_macros.h
+++ b/include/wil/result_macros.h
@@ -1297,7 +1297,7 @@ namespace wil
 
         // Called to tell Appverifier to ignore a particular allocation from leak tracking
         // If AppVerifier is not enabled, this is a no-op
-        // Desktop/System Only: Automatically setup when building Windows (BUILD_WINDOWS defined)
+        __declspec(selectany) bool g_fetchedRtlDisownModuleHeapAllocation = false;
         __declspec(selectany) NTSTATUS(__stdcall *g_pfnRtlDisownModuleHeapAllocation)(_In_ HANDLE heapHandle, _In_ PVOID address) WI_PFN_NOEXCEPT = nullptr;
 
         // Allocate and disown the allocation so that Appverifier does not complain about a false leak
@@ -1308,6 +1308,16 @@ namespace wil
             if (g_pfnRtlDisownModuleHeapAllocation)
             {
                 (void)g_pfnRtlDisownModuleHeapAllocation(::GetProcessHeap(), allocation);
+            }
+            else if (!g_fetchedRtlDisownModuleHeapAllocation)
+            {
+                g_pfnRtlDisownModuleHeapAllocation = reinterpret_cast<decltype(g_pfnRtlDisownModuleHeapAllocation)>(::GetProcAddress(::GetModuleHandleW(L"ntdll.dll"), "RtlDisownModuleHeapAllocation"));
+                g_fetchedRtlDisownModuleHeapAllocation = true;
+
+                if (g_pfnRtlDisownModuleHeapAllocation)
+                {
+                    (void)g_pfnRtlDisownModuleHeapAllocation(::GetProcessHeap(), allocation);
+                }
             }
 
             return allocation;
--- a/include/wil/result_macros.h
+++ b/include/wil/result_macros.h
@@ -1295,28 +1295,28 @@ namespace wil
         // Plugin to call RoFailFastWithErrorContext (WIL use only)
         __declspec(selectany) void(__stdcall* g_pfnFailfastWithContextCallback)(wil::FailureInfo const& failure) WI_PFN_NOEXCEPT = nullptr;
 
-        // Called to tell Appverifier to ignore a particular allocation from leak tracking
-        // If AppVerifier is not enabled, this is a no-op
-        __declspec(selectany) bool g_fetchedRtlDisownModuleHeapAllocation = false;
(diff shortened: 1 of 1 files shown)


include/wil/result_macros.h
```diff
@@ -1309,6 +1309,16 @@ namespace wil
             {
                 (void)g_pfnRtlDisownModuleHeapAllocation(::GetProcessHeap(), allocation);
             }
+            else if (!g_fetchedRtlDisownModuleHeapAllocation)
+            {
+                g_pfnRtlDisownModuleHeapAllocation = reinterpret_cast<decltype(g_pfnRtlDisownModuleHeapAllocation)>(::GetProcAddress(::GetModuleHandleW(L"ntdll.dll"), "RtlDisownModuleHeapAllocation"));
+                g_fetchedRtlDisownModuleHeapAllocation = true;
+
+                if (g_pfnRtlDisownModuleHeapAllocation)
```
jonwis, line 1317: This logic does seem a little unnatural, but the _common_ case is that the pointer is set once the process starts, so I thought I'd leave that in place and put the boolean-check afterwards.

brbeec (approved):
Thanks for addressing this issue Jon!

jonwis:
> 1. In the Windows build, setting the pointer is guarded by `#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_SYSTEM)`. I'm curious if there was good reason behind that or if it was unnecessary (very well could be that we wanted to suppress private API use when building as APP, but I lack the history here)

My assumption is that's the case - WIL is used in many kinds of environments, so this was likely defensive only to prevent pulling in publicly-documented but not-available-to-WACK-in-the-kit code.  The method itself is not sensitive to app type.

jonwis:
> 2\. can we just eliminate the global and move the state inside of the `ProcessHeapAlloc` function?

The BUILD_WINDOWS code relies on access to the RTL header and import libraries to skip the runtime probing for the method. It's probably OK to move the whole thing to a function local.

When we merge this back into the OS we'll need to delete the assignment in the build-windows-specific init path.

dunhor:
> The BUILD_WINDOWS code relies on access to the RTL header and import libraries to skip the runtime probing for the method.

Perhaps I'm overlooking something obvious, but I don't see any dependency on the RTL header/import libraries in the OS code. The code in question is pasted below. The only "issue" is that `wil_details_GetNtDllModuleHandle` isn't present in the GitHub copy of the code, but all it does is stash a reference to the "ntdll.dll" `HMODULE` after a `GetModuleHandleW`.

```c++
        inline NTSTATUS __stdcall RtlDisownModuleHeapAllocation(_In_ HANDLE heapHandle, _In_ PVOID address) WI_NOEXCEPT
        {
            static decltype(RtlDisownModuleHeapAllocation) *s_pfnRtlDisownModuleHeapAllocation = nullptr;
            if (s_pfnRtlDisownModuleHeapAllocation == nullptr)
            {
                s_pfnRtlDisownModuleHeapAllocation = reinterpret_cast<decltype(RtlDisownModuleHeapAllocation)*>(GetProcAddress(wil_details_GetNtDllModuleHandle(), "RtlDisownModuleHeapAllocation"));
            }
            return s_pfnRtlDisownModuleHeapAllocation ? s_pfnRtlDisownModuleHeapAllocation(heapHandle, address) : STATUS_SUCCESS;
        }
```

jonwis:
> overlooking something obvious

Nope, I was _drastically_ misremembering the internal implementation. I'll fix it the way you suggest and we can eliminate the fake Rtl... call in the internal code when it gets merged.

include/wil/result_macros.h
```diff
@@ -1295,19 +1295,29 @@ namespace wil
...
         __declspec(selectany) void(__stdcall* g_pfnFailfastWithContextCallback)(wil::FailureInfo const& failure) WI_PFN_NOEXCEPT = nullptr;
 
-        // Called to tell Appverifier to ignore a particular allocation from leak tracking
-        // If AppVerifier is not enabled, this is a no-op
-        // Desktop/System Only: Automatically setup when building Windows (BUILD_WINDOWS defined)
-        __declspec(selectany) NTSTATUS(__stdcall *g_pfnRtlDisownModuleHeapAllocation)(_In_ HANDLE heapHandle, _In_ PVOID address) WI_PFN_NOEXCEPT = nullptr;
 
         // Allocate and disown the allocation so that Appverifier does not complain about a false leak
-        inline PVOID ProcessHeapAlloc(_In_ DWORD flags, _In_ size_t size)
+        inline PVOID ProcessHeapAlloc(_In_ DWORD flags, _In_ size_t size) WI_PFN_NOEXCEPT
```
dunhor, line 1300: Kind of a nit: this should be `WI_NOEXCEPT`. The `WI_PFN_NOEXCEPT` macro is kind of a hack workaround due to a back-compat compiler bug when MSVC added "noexcept to the type system" support.

dunhor (approved):

include/wil/result_macros.h
sudhakarvp1: @jonwis wondering as to when this change will arrive in OS2020 repo. Thanks.

PR #261, opened 2022-09-21
WI_NODISCARD is not very useful because it does not create compiler warnings/errors

### LINKED ISSUE #260: WI_NODISCARD is not very useful because it does not create compiler warnings/errors
Currently `WI_NODISCARD` does not expand to `[[nodiscard]]`.  It just has a SAL annotation that does not seem to be especially useful.  For example, the following code does not generate a compiler warning or error as expected:

```cpp
auto lock = wil::critical_section;
lock.lock(); // discarding lock lifetime object
```

What this __should__ be doing is reporting a warning (ideally treated as an error).
`warning C4834: discarding return value of function with 'nodiscard' attribute`

dmachaj:
Closes #260

I am converting some custom lock wrappers to `wil::critical_section` and was trying to demonstrate that accidentally dropping the lifetime wrapper generates an error.  Whoops, it does not.  It happily compiles with no obvious warnings and hides a severe bug.

Use the C++ standard language feature to detect `[[nodiscard]]` support and if it is present define `WI_NODISCARD` to be `[[nodiscard]]`.  If the feature is not present then the existing (mostly unhelpful) annotation is used instead.

init_all + build_all + runtests

```
Building from v:\T\wil\build\clang64debug
[85/85] Linking CXX executable tests\win7\witest.win7.exe
Building from v:\T\wil\build\clang64relwithdebinfo
[85/85] Linking CXX executable tests\win7\witest.win7.exe
Building from v:\T\wil\build\msvc64debug
[86/86] Linking CXX executable tests\win7\witest.win7.exe
Building from v:\T\wil\build\msvc64relwithdebinfo
[86/86] Linking CXX executable tests\win7\witest.win7.exe
All build completed successfully
```

There was one test failure that seems unrelated (case sensitivity error):
```
V:\T\wil\tests\FileSystemTests.cpp(613): FAILED:
  REQUIRE( procName == moduleName )
with expansion:
  "V:\T\wil\build\clang64debug\tests\cpplatest\witest.cpplatest.exe"
  ==
  "v:\T\wil\build\clang64debug\tests\cpplatest\witest.cpplatest.exe"
===============================================================================
test cases:    190 |    189 passed | 1 failed
assertions: 111716 | 111715 passed | 1 failed
```

I also overwrote the macro in the copy of WIL for an existing project and intentionally dropped the lock object.

```cpp
auto lock = wil::critical_section;
lock.lock(); // discarding lock lifetime object
```
`warning C4834: discarding return value of function with 'nodiscard' attribute`
`error C2220: the following warning is treated as an error`

### DIFF
--- a/include/wil/common.h
+++ b/include/wil/common.h
@@ -325,8 +325,12 @@ Three exception modes are available:
 #define WIL_HAS_CXX_17 0
 #endif
 
-// Until we'll have C++17 enabled in our code base, we're falling back to SAL
+#if __has_cpp_attribute(nodiscard)
+#define WI_NODISCARD [[nodiscard]]
+#else
+// If we don't have C++17 enabled in our code base then fall back to SAL
 #define WI_NODISCARD __WI_LIBCPP_NODISCARD_ATTRIBUTE
+#endif
 
 #define __R_ENABLE_IF_IS_CLASS(ptrType)                     wistd::enable_if_t<wistd::is_class<ptrType>::value, void*> = (void*)0
 #define __R_ENABLE_IF_IS_NOT_CLASS(ptrType)                 wistd::enable_if_t<!wistd::is_class<ptrType>::value, void*> = (void*)0


dmachaj:
Now I am super confused.  I dug a little deeper and [[nodiscard]] should be getting defined here: [https://github.com/microsoft/wil/blob/master/include/wil/wistd_config.h#L186](https://github.com/microsoft/wil/blob/master/include/wil/wistd_config.h#L186).

dunhor:
> The version is 14 so it does not use [[nodiscard]]

`[[nodiscard]]` was added in c++17: https://en.cppreference.com/w/cpp/language/attributes/nodiscard

dmachaj:
I think the version checking in WIL is wrong for this project.  If I apply this edit to the header used by that project then it works exactly as intended (that is where the warning output came from).

dunhor:
Then I'm confused by your comment. Were you saying that your project is using c++14 or something else?

dmachaj:
I am now super confused too.  I took a closer look at our project files and I don't see where the C++ standard is being set.  As a result it seems to be defaulting to C++14.  But that is very weird because [[nodiscard]] feature checking evaluated to true and produced the expected build break.  If it is truly C++14 then it should have broken in a different way.

dunhor:
Interesting. Looks like Clang also allows things like `[[nodiscard]]` when compiling with older C++ versions: https://godbolt.org/z/e77EW3d1r

We can probably do something like this which will "turn on" the feature, even in situations where it probably shouldn't:
```c++
#ifdef __has_cpp_attribute
#if __has_cpp_attribute(nodiscard)
#define __WI_LIBCPP_NODISCARD_ATTRIBUTE [[nodiscard]]
#endif
#endif

#ifndef __WI_LIBCPP_NODISCARD_ATTRIBUTE
#if __WI_LIBCPP_STD_VER > 14
#define __WI_LIBCPP_NODISCARD_ATTRIBUTE [[nodiscard]]
#else
#define __WI_LIBCPP_NODISCARD_ATTRIBUTE _Check_return_
#endif
#endif
```

sylveon:
MSVC and NVCC both also support [[nodiscard]] downlevel, because it helps make code safer (the STL/other libraries can add nodiscard to functions and client code will get warnings even when using C++14).

I think the `#elif __WI_LIBCPP_STD_VER > 14` branch is redundant however, simply a check for `defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard)` should be enough

dunhor:
> I think the `#elif __WI_LIBCPP_STD_VER > 14` branch is redundant

From what I can gather, `[[nodiscard]]` was added in VS 2017 15.3, but `__has_cpp_attribute` wasn't added until VS 2019 16.0: https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170. I did see something that suggests 15.8 had support for `__has_cpp_attribute`, but that's (1) from a third party and I can't confirm if that's true or not, and (1) it's also later than 15.3

> simply a check for `defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard)` should be enough

You can't do a `defined(xyz)` check _and_ use it in the same directive (else you'll get `xyz is not a function style macro` errors.

dunhor:
Arguably, the C++ version check is probably wrong (I'm not sure when the `_MSVC_LANG` version got bumped). That said, I haven't seen reports of a finer-grained `_MSC_FULL_VER` check being needed.

include/wil/common.h
```diff
@@ -325,8 +325,12 @@ Three exception modes are available:
 #define WIL_HAS_CXX_17 0
 #endif
 
-// Until we'll have C++17 enabled in our code base, we're falling back to SAL
+#if __has_cpp_attribute(nodiscard)
```
jonwis, line 328: Does this need to be guarded with a check for the presence of the `__has_cpp_attribute` doodad?  I see things like:

```c
#ifdef __has_cpp_attribute
#if __has_cpp_attribute(foo)
#else
#endif
#endif
```

AtariDreams:
Any updates on this?

dmachaj:
I haven't touched the project where I needed this for ~2 weeks.  This is on my list of things to revisit when I am there next.

dmachaj:
Closing this out because I have not yet returned to the code base where I thought I needed this and it is unlikely that I will in the future.  This is also much more complicated than I realized and I don't have the time to fully understand it.

PR #266, opened 2022-10-10
Add support for zwstring_view to natvis. Fix parse error in natvis XML

### LINKED ISSUE #265: Add natvis support for wil::zwstring_view
The visualization in windbg is not very good for this string type.  It should be visualized nicely like the other types.

dmachaj:
Closes #265

I recently began using wil::zwstring_view in a new project and when debugging in windbg the visualization for these string types is not very good.  They must be manually expanded to see the contents of the string.  The experience is worse than std::wstring_view.

This string type is a thin wrapper over `std::basic_string_view` so I just copied the natvis for that type and updated the name so it knows to use it.

There was also a preexisting warning when loading the XML because the `<?xml...>` element was not the first bytes of the file.  I'm not sure if that was breaking things or just a warning but I fixed it regardless.

I created an empty console app and added WIL to it.  I added a `wil::zwstring_view` variable and debugged the project with windbg.  The edits were made to the wil.natvis that came with the WIL NuGet package.

### Before this change
image

### After this change
image

Note that the value is visible without expanding it.  The size is also newly visible.

### DIFF
--- a/natvis/wil.natvis
+++ b/natvis/wil.natvis
@@ -1,3 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
 <!--
     Copyright (c) Microsoft. All rights reserved.
     This code is licensed under the MIT License.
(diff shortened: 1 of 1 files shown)


ChrisGuzak (approved):

ChrisGuzak:
are your before/after pictures reversed?

dmachaj:
No, they are not.  The before required expanding to see the string value.  After has it top-level.  I expanded it to also show that the underlying buffer and size are now visible.

jonwis (approved):

jonwis:
I do think the "size" is more relevant than the array of characters, though. Is there a mode to show "{size} {string-content}" or similar without getting the full dump of characters?

dmachaj:
That's fair.  Iteration 2.

image

ChrisGuzak (approved):

PR #273, opened 2022-10-21
Attempt to guard zstring_view against reading out of bounds

### LINKED ISSUE #271: Possible out-of-bounds access in zstring_view.
Hi, I'm interested in your take on 
```
        template<size_t stringArrayLength>
        constexpr basic_zstring_view(const TChar(&stringArray)[stringArrayLength]) noexcept
            : std::basic_string_view<TChar>(&stringArray[0])
        {
            if (stringArrayLength < this->size()) { WI_STL_FAIL_FAST_IF(true); }
        }
```
Whenever the user is passing in a static buffer with a known length, he pretty much declares that he doesn't own any memory following the end of this buffer. However, this implementation discards the size information and passes a pointer to it to string_view's constructor which will then walk as far as needed to discover first nul terminator. In case it was not part of the buffer, it will walk through potentially unowned memory and stop only on its occurrence, potentially causing access violation. 

Wouldn't it be safer to prevent this undefined behavior by simply testing the last character of the buffer to be nul? The current approach also contains false positives - if you have a buffer without nul but happens to be directly followed by a nul from a completely different allocation, it will pass this condition (imho should have been `<= this->size()`). But once the owner of the nul decided to deallocate it, you can get a run-time access violation on using this zstring_view since it relies on the deallocated nul to be present (or worse - upon passing to LPCSTR accepting WinAPI function, it will read garbage until next nul in memory).

CookiePLMonster:
Following the discussion in #271, this PR serves as an attempt to fixup `zstring_view` constructors against a possible out-of-bounds read:
1. `basic_zstring_view(const TChar(&stringArray)[stringArrayLength])` constructor ignored the passed array length and passed the buffer straight to an underlying `string_view` constructor, which in turn attempted to determine the length of the string. If the passed buffer contains no null terminator, it'd read out of bounds. The current tests didn't fail on this, because they pass on a crash - so even if a read access violation happened, it'd be caught and registered as a pass.
With my change, the constructor assumes that the buffer contains a null terminator, and it determines the length safely via a bounds-checked variation of `char_traits::length` (that unfortunately doesn't exist out of the box!). If no null terminator is found in the passed buffer, construction fails.
New tests have been added where a null terminator sits one character/byte past the passed buffer. Without my changes, these strings constructed with no error, while they should not have.
2. In theory, `basic_zstring_view(const TChar* pStringData, size_type stringLength)` still potentially reads past a passed buffer, but since the size parameter there is explicit, I decided against changing it - otherwise it could break some valid cases where the user ensures that the passed buffer is null terminated. I think this is a valid assumption, far more believeable one than in the first case where the user was trusted with null terminating an implicitly-sized buffer.

However, there is a downside to the current approach - as you can notice from newly added "passing" tests, `pStringData[stringLength] != 0` check disqualifies the second constructor from being truly `constexpr` since it reads out of bounds of the buffer! This should potentially be thought of as a bug too, but I'm not sure if the fix is to change the logic of this constructor, or just to make it not `constexpr`.

Resolves #271

### DIFF
--- a/include/wil/stl.h
+++ b/include/wil/stl.h
@@ -143,9 +143,8 @@ namespace wil
 
         template<size_t stringArrayLength>
         constexpr basic_zstring_view(const TChar(&stringArray)[stringArrayLength]) noexcept
-            : std::basic_string_view<TChar>(&stringArray[0])
+            : std::basic_string_view<TChar>(&stringArray[0], length_n(&stringArray[0], stringArrayLength))
         {
-            if (stringArrayLength < this->size()) { WI_STL_FAIL_FAST_IF(true); }
         }
 
         // Construct from nul-terminated char ptr. To prevent this from overshadowing array construction,
@@ -172,6 +171,15 @@ namespace wil
         }
 
     private:
+        // Bounds-checked version of char_traits::length, like strnlen. Requires that the input contains a null terminator.
+        static constexpr size_type length_n(_In_reads_opt_(buf_size) const TChar* str, size_type buf_size) noexcept
+        {
+            const std::basic_string_view<TChar> view(str, buf_size);
+            auto pos = view.find_first_of(TChar());
+            if (pos == view.npos) { WI_STL_FAIL_FAST_IF(true); }
+            return pos;
+        }
+
         // The following basic_string_view methods must not be allowed because they break the nul-termination.
         using std::basic_string_view<TChar>::swap;
         using std::basic_string_view<TChar>::remove_suffix;
--- a/tests/StlTests.cpp
+++ b/tests/StlTests.cpp
@@ -107,6 +107,13 @@ TEST_CASE("StlTests::TestZStringView", "[stl][zstring_view]")
     REQUIRE_CRASH((wil::zstring_view{ &badCharArray[0][0], _countof(badCharArray[0]) }));
     REQUIRE_CRASH((wil::zstring_view{ badCharArray[0] }));
 
+    // Test constructing with a NULL one character past the valid range, guarding against off-by-one errors
+    // Overloads taking an explicit length trust the user that they ensure valid memory follows the buffer
+    static constexpr char badCharArrayOffByOne[2][3] = {{'a', 'b', 'c' }, {}};
+    const wil::zstring_view fromTerminatedCharArray(&badCharArrayOffByOne[0][0], _countof(badCharArrayOffByOne[0]));
+    REQUIRE(fromLiteral == fromTerminatedCharArray);
+    REQUIRE_CRASH((wil::zstring_view{ badCharArrayOffByOne[0] }));
+
     // Test constructing from custom string type
     CustomNoncopyableString customString;
     wil::zstring_view fromCustomString(customString);
@@ -161,6 +168,13 @@ TEST_CASE("StlTests::TestZWStringView", "[stl][zstring_view]")
     static constexpr wchar_t badCharArray[2][3] = {{ L'a', L'b', L'c' }, { L'a', L'b', L'c' } };
     REQUIRE_CRASH((wil::zwstring_view{ &badCharArray[0][0], _countof(badCharArray[0]) }));
     REQUIRE_CRASH((wil::zwstring_view{ badCharArray[0] }));
+
+    // Test constructing with a NULL one character past the valid range, guarding against off-by-one errors
+    // Overloads taking an explicit length trust the user that they ensure valid memory follows the buffer
+    static constexpr wchar_t badCharArrayOffByOne[2][3] = {{ L'a', L'b', L'c' }, {}};
+    const wil::zwstring_view fromTerminatedCharArray(&badCharArrayOffByOne[0][0], _countof(badCharArrayOffByOne[0]));
+    REQUIRE(fromLiteral == fromTerminatedCharArray);
+    REQUIRE_CRASH((wil::zwstring_view{ badCharArrayOffByOne[0] }));
 
     // Test constructing from custom string type
     CustomNoncopyableString customString;


include/wil/stl.h
```diff
@@ -172,6 +172,18 @@ namespace wil
         }
 
     private:
+        // Bounds-checked version of char_traits::length, like strnlen.
+        static constexpr size_type length_n(_In_reads_opt_(length) const TChar* str, size_type buf_size) noexcept
```
dunhor, line 176: I'm always hesitant of custom writing things like this. Using something like `strnlen` directly would obviously be the most preferable, but then you'd need to overload on `TChar`, etc... `if constexpr` wouldn't be _that_ bad here. Otherwise, we can construct `basic_string_view` using `stringArrayLength` and then resize/fail-fast based off the result of `find_first_of`.

CookiePLMonster: `TChar` overloads are likely not to work on u16/u32strings.

dunhor: I was more suggesting specializing on types that do have a function/intrinsic available. E.g. `char` and `wchar_t` and then fall back to something like this for other types.

FWIW the `wchar_t` <-> `char16_t` cast is used pretty freely in things like C++/WinRT

CookiePLMonster: I'll check it, with one remark - I doubt I can use `if constexpr` since that'll lock the code to C++17 and newer. I'll have to look into SFINAE.

dunhor: `basic_string_view` was also added in C++17 (I suppose your edit suggests you realized that). But I also forgot that the string functions didn't get the `constexpr` update when `char_tratits` etc. did :/. I'm also unsure if `WI_STL_FAIL_FAST_IF` is `constexpr`, but that's to look at another time.

`find_first_of` seems reasonable IMO and is no more complex than what you have here.
```c++
        // Bounds-checked version of char_traits::length, like strnlen.
        static constexpr size_type length_n(_In_reads_opt_(length) const TChar* str, size_type length) noexcept
        {
            std::basic_string_view<TChar> view(str, length);
            auto pos = view.find_first_of(TChar());
            return (pos == view.npos) ? length : pos; // Perhaps maybe even move the failfast here to be simpler
        }
```
I'd probably personally prefer doing all the operations on `this`, but alas there is no way to change the length of a `basic_string_view` apart from assignment (e.g. with `.substr()`), but that likely causes a mess of other issues...

CookiePLMonster: Ah, like this, good idea! Just one note, though - fast failing inside this function discards a possibility that there is a null terminator right after the passed size, as per my footnote in the initial post. Should this be a supported scenario (that then disqualifies true `constexpr`), or should it be forbidden? If the latter, then the test checking for this particular scenario should be discarded.

dunhor: Given that this is just used for construction with arrays, I'm not sure that concerning ourselves with the data outside of the bounds of the array is a good idea, and I'd like to believe that nobody has taken a dependency on it. E.g. maybe it's a value that's initialized to zero, but later changes to something else. There's also no guarantee that the byte following the array can even be read. E.g. it's possible to get a page fault/access violation. Are you saying that there's an existing test that expects to be able to read one past the end of the array and find a null character? I only see the tests you added which verify a crash (which I'd expect).

Of course, you also have the issue that there's no guarantee of `const`-ness of the array, so someone could pass a non-`const` array that they later change breaking the precondition, but I'm not really worried about people going to such lengths to shoot themselves in the foot.

CookiePLMonster: > Are you saying that there's an existing test that expects to be able to read one past the end of the array and find a null character?

Technically no, the passing test I added does, though, hence the remark `Overloads taking an explicit length trust the user that they ensure valid memory follows the buffer`.

If you think it's worth taking the "risk" here and disallowing it entirely, I'm all for it. Disallowing the assumption that a null terminator follows the string would probably simplify code, allow those string views to be truly `constexpr` (as reading past array bounds cannot be), and probably be safer overall. My current passing test would become invalid but I could add another one, e.g. with a `{`a`, `b`, `c`, `\0`, `b`, `u`, `f`,}` array where the expectation is that it produces an `"abc"` view.

dunhor: Ahh, okay, I misread. I thought you were referring to constructing with arrays, my bad. Yes, I agree, the caller should assume that it's safe to read the null terminator in all situations. E.g. if I pass `(ptr, length)`, then it should be assumed that reading from `ptr[length]` is valid, won't AV, and gives the null character.

CookiePLMonster: Indeed, I was just editing my answer as you left yours, here it is unedited:

```
EDIT:
On a second look, my "passing" tests verify another constructor, but that is potentially alright. Securing this specific constructor against reading past the array bounds would change the contract, so it might be better to trust the user that they pass a valid size.
```

Looks like we effectively agreed on the same! My "passing" test assumes this exactly, i.e. that `ptr[length] == '\0`.

dunhor: Yeah, I just mis-understood your question. The tests as you have them written make sense to me and look correct. If I try and construct with an array, I expect the null character to be within the bounds of the array (otherwise failfast). If I construct with a `(pointer, length)` pair, I'm guaranteeing that it's safe to read from index `0` to index `length` and that `pointer[length]` is the null character. If constructing with a `(pointer, length)` pair does not produce a constant expression (e.g. would read past the end of an array), that's on the caller and they therefore should not try and do something like that in a `constexpr` context.

CookiePLMonster: There, I applied the changes we agreed upon. `WI_STL_FAIL_FAST_IF` is **not** `constexpr`, but the compiler realizes that the code branch containing it is never taken in the current `constexpr` contexts, and thus it's fine. Had it been `WI_STL_FAIL_FAST_IF(pos == view.npos);`, it would not compile.

include/wil/stl.h
```diff
@@ -172,6 +171,15 @@ namespace wil
         }
 
     private:
+        // Bounds-checked version of char_traits::length, like strnlen. Requires that the input contains a null terminator.
+        static constexpr size_type length_n(_In_reads_opt_(length) const TChar* str, size_type buf_size) noexcept
```
dunhor, line 175: One last minor comment. The `_In_reads_opt_(  --->  length  <---  )` does not match the `buf_size` argument nmae

dunhor (approved):

PR #280, opened 2022-11-28
Add RECT helpers

mcooley:
Helper functions to get the width or height of a rectangle are commonly copy-and-pasted in code that deals with win32 user interfaces. (Microsoft-internal folks can search the OS repo for "RectWidth" to see evidence of this.) It's a small thing since the implementations are trivial, but it would be nice to have a canonical copy of these in wil for new projects.

This PR adds five functions in win32_helpers.h:

- rect_width
- rect_height
- rect_is_empty
- rect_contains_point
- rect_from_size

The function templates work on the Windows SDK types RECT, RECTL, D2D_RECT_F, D2D_RECT_U, or any other type with the same shape. <s>There are also versions which operate on pointers to rectangles like the user32 functions PtInRect and IsRectEmpty do.</s>

### DIFF
--- a/include/wil/win32_helpers.h
+++ b/include/wil/win32_helpers.h
@@ -286,6 +286,56 @@ namespace wil
     }
 #pragma endregion
 
+#pragma region RECT helpers
+    template<typename rect_type>
+    constexpr auto rect_width(rect_type rect)
+    {
+        return rect.right - rect.left;
+    }
+
+    template<typename rect_type>
+    constexpr auto rect_width(rect_type* rect)
+    {
+        return rect->right - rect->left;
+    }
+
+    template<typename rect_type>
+    constexpr auto rect_height(rect_type rect)
+    {
+        return rect.bottom - rect.top;
+    }
+
+    template<typename rect_type>
+    constexpr auto rect_height(rect_type* rect)
+    {
+        return rect->bottom - rect->top;
+    }
+
+    template<typename rect_type>
+    constexpr auto rect_is_empty(rect_type rect)
+    {
+        return (rect.left >= rect.right) || (rect.top >= rect.bottom);
+    }
+
+    template<typename rect_type>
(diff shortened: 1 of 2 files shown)


CookiePLMonster:
Idea from someone who's been dealing with this literally 2 days ago - does it make sense to also introduce the "opposite" operators, to construct a type from width/height? basically `{x , y, x+width, y+height}`.

include/wil/win32_helpers.h
```diff
@@ -286,6 +286,56 @@ namespace wil
...
+    template<typename rect_type>
+    constexpr auto rect_is_empty(rect_type rect)
+    {
+        return (rect.left >= rect.right) || (rect.top >= rect.bottom);
+    }
+
+    template<typename rect_type>
+    constexpr auto rect_is_empty(rect_type* rect)
+    {
+        return (rect->left >= rect->right) || (rect->top >= rect->bottom);
```
dmachaj, line 323: It is not ideal that the pointer version duplicates the value type version for each of these methods.  Would it make more sense to call across instead?

```suggestion
        return rect_is_empty(*rect);
```

mcooley:
I removed the versions which take pointer parameters, since after some reflection I doubt they'll be needed widely. The remaining functions have been updated to take parameters by `const&`.

> Idea from someone who's been dealing with this literally 2 days ago - does it make sense to also introduce the "opposite" operators, to construct a type from width/height? basically {x , y, x+width, y+height}.

Something like `auto myRect = rect_from_size<RECT>(x, y, width, height);`?

CookiePLMonster:
> Something like `auto myRect = rect_from_size<RECT>(x, y, width, height);`?

Yep, sounds good!

CookiePLMonster (approved):

dmachaj (approved):

ChrisGuzak (approved):

PR #285, opened 2022-12-28
Add Tests with Address & UB Sanitizers Enabled

dunhor:
This enables Address Sanitizer for all MSVC targets and non-debug Clang targets as there doesn't appear to be proper debug libraries for Clang (or at least the wrong ones are being picked up when I try) and Undefined Behavior Sanitizer for non-debug Clang targets (again, no proper debug libraries are found) as there's no current MSVC support for UB Sanitizer. A-San found no active bugs. UB-San found some benign issues that have been fixed, one of which was in WRL which was also fixed in the OS and the header contents in the checked-in workarounds header have been updated.

I also took an opportunity to update some of the CMake to use more "modern" CMake practices. I also tried to update the `RelWithDebugInfo` target to compile with `/Ob2` as `/Ob1` is the default, for some reason, however this is currently blocked with [what appears to be a bug in Clang](https://github.com/llvm/llvm-project/issues/59690).

### DIFF
--- a/include/wil/wistd_functional.h
+++ b/include/wil/wistd_functional.h
@@ -250,21 +250,24 @@ namespace wistd     // ("Windows Implementation" std)
             return _Invoker::__call(__f_, wistd::forward<_ArgTypes>(__arg)...);
         }
 
-    }  // __function
-
-    template<class _Rp, class ..._ArgTypes>
-    class __WI_LIBCPP_TEMPLATE_VIS function<_Rp(_ArgTypes...)>
-        : public __function::__maybe_derive_from_unary_function<_Rp(_ArgTypes...)>,
-          public __function::__maybe_derive_from_binary_function<_Rp(_ArgTypes...)>
-    {
         // 'wistd::function' is most similar to 'inplace_function' in that it _only_ permits holding function objects
         // that can fit within its internal buffer. Therefore, we expand this size to accommodate space for at least 12
         // pointers (__base vtable takes an additional one).
-        static constexpr size_t __buffer_size = 13 * sizeof(void*);
+        constexpr const size_t __buffer_size = 13 * sizeof(void*);
 
+    }  // __function
+
+    // NOTE: The extra 'alignas' here is to work around the x86 compiler bug mentioned in
+    // https://github.com/microsoft/STL/issues/1533 to force alignment on the stack
+    template<class _Rp, class ..._ArgTypes>
+    class __WI_LIBCPP_TEMPLATE_VIS __WI_ALIGNAS(typename aligned_storage<__function::__buffer_size>::type)
+    function<_Rp(_ArgTypes...)>
+        : public __function::__maybe_derive_from_unary_function<_Rp(_ArgTypes...)>,
+          public __function::__maybe_derive_from_binary_function<_Rp(_ArgTypes...)>
+    {
         typedef __function::__base<_Rp(_ArgTypes...)> __base;
         __WI_LIBCPP_SUPPRESS_NONINIT_ANALYSIS
-        typename aligned_storage<__buffer_size>::type __buf_;
+        typename aligned_storage<__function::__buffer_size>::type __buf_;
         __base* __f_;
 
         __WI_LIBCPP_NO_CFI static __base *__as_base(void *p) {
--- a/include/wil/wistd_memory.h
+++ b/include/wil/wistd_memory.h
@@ -149,8 +149,8 @@ namespace wistd     // ("Windows Implementation" std)
     struct __second_tag {};
 
     template <class _T1, class _T2>
-    class __compressed_pair : private __compressed_pair_elem<_T1, 0>,
-                              private __compressed_pair_elem<_T2, 1> {
+    class __declspec(empty_bases) __compressed_pair : private __compressed_pair_elem<_T1, 0>,
+                                                      private __compressed_pair_elem<_T2, 1> {
       typedef __compressed_pair_elem<_T1, 0> _Base1;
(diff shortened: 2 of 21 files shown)


include/wil/wistd_functional.h
```diff
@@ -250,21 +250,24 @@ namespace wistd     // ("Windows Implementation" std)
...
 
+        // 'wistd::function' is most similar to 'inplace_function' in that it _only_ permits holding function objects
+        // that can fit within its internal buffer. Therefore, we expand this size to accommodate space for at least 12
+        // pointers (__base vtable takes an additional one).
+        constexpr const size_t __buffer_size = 13 * sizeof(void*);
+
     }  // __function
 
+    // NOTE: The extra 'alignas' here is to work around the x86 compiler bug mentioned in
+    // https://github.com/microsoft/STL/issues/1533 to force alignment on the stack
```
dunhor, line 261: Without this, passing `wistd::function` by value as function arguments can cause alignment issues

include/wil/wistd_memory.h
```diff
@@ -149,8 +149,8 @@ namespace wistd     // ("Windows Implementation" std)
     struct __second_tag {};
 
     template <class _T1, class _T2>
-    class __compressed_pair : private __compressed_pair_elem<_T1, 0>,
-                              private __compressed_pair_elem<_T2, 1> {
+    class __declspec(empty_bases) __compressed_pair : private __compressed_pair_elem<_T1, 0>,
```
dunhor, line 152: I've lost the pointer to the documentation on this, but without the `__declspec`, the size of the pair will be correct if one of the types is empty, however the address of `second` will still be beyond the bounds of the object

scripts/init.cmd
```diff
@@ -10,8 +10,8 @@ goto :init
 :usage
     echo USAGE:
     echo     init.cmd [--help] [-c^|--compiler ^<clang^|msvc^>] [-g^|--generator ^<ninja^|msbuild^>]
-    echo         [-b^|--build-type ^<debug^|release^|relwithdebinfo^|minsizerel^>] [-l^|--linker link^|lld-link]
-    echo         [-v^|--version X.Y.Z] [--cppwinrt ^<version^>] [--fast]
+    echo         [-b^|--build-type ^<debug^|release^|relwithdebinfo^|minsizerel^>] [-v^|--version X.Y.Z]
```
dunhor, line 13: NOTE: `lld-link` appears to be the default now

tests/CommonTests.cpp
```diff
@@ -159,7 +159,7 @@ enum class EClassTest
 };
 DEFINE_ENUM_FLAG_OPERATORS(EClassTest);
 
-enum ERawTest
+enum ERawTest : unsigned int
```
dunhor, line 162: Clearing flags is UB if there's no explicit underlying type because `~val` will set bits beyond the representation of the enum values. There's no issue with `enum class` as it has well-defined defaults for the underlying type.

tests/WinRTTests.cpp
```diff
@@ -240,8 +240,8 @@ void DoHStringDifferentValueComparisonTest(const wchar_t (&lhs)[LhsSize], const
     DoHStringComparisonTest<InhibitArrayReferences, IgnoreCase>(lhsUniqueStr, rhsUniqueStr, 1);
 
 #ifdef WIL_ENABLE_EXCEPTIONS
-    std::wstring lhsWstr(lhs, 7);
```
dunhor, line 243: I can't, for the life of me, understand why this was originally done, however it's obvious UB to do this and the tests have no reliance on it

tests/main.cpp
```diff
@@ -6,3 +6,15 @@
...
 #include "catch.hpp"
+
+#if WITEST_ADDRESS_SANITIZER
+extern "C" __declspec(dllexport)
+const char* __asan_default_options()
+{
+    return
+        // Tests validate OOM, so this is expected
+        "allocator_may_return_null=1"
+        // Some structs in Windows have dynamic size where we over-allocate for extra data past the end
```
dunhor, line 17: A-San complains that we are calling `delete` such that the memory allocated is > `sizeof(T)`

tests/sanitize-address/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    target_compile_definitions(witest.asan PRIVATE
+        # Not compatible with using lld-link
+        -D_DISABLE_VECTOR_ANNOTATION
+        # See below; not compatible with exceptions
+        -DCATCH_CONFIG_DISABLE_EXCEPTIONS
+        )
+
+    # Clang ASan on Windows has issues with Windows: https://github.com/google/sanitizers/issues/749
```
dunhor, line 19:
```suggestion
    # Clang ASan on Windows has issues with exceptions: https://github.com/google/sanitizers/issues/749

```

scripts/azure-pipelines.yml
```diff
@@ -18,7 +18,7 @@ jobs:
     displayName: 'Install Clang'
 
   - script: |
-      call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars32.bat"
+      call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsamd64_x86.bat""
```
dunhor, line 21:
```suggestion
      call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsamd64_x86.bat"

```

ChrisGuzak (approved):

PR #287, opened 2023-01-11
<wil/Tracelogging.h> does not compile in Clang and has some spelling errors

### LINKED ISSUE #286: <wil/Tracelogging.h> does not compile in Clang and has some spelling errors

dmachaj:
Closes #286

A private repo had a very old fork of <wil/Tracelogging.h> that also had some local fixes made to it over the years.  I recently deleted this fork in favor of using the latest version of WIL.  This set of changes is upstreaming those local fixes so that all may benefit.

The biggest change is that this header now compiles with Clang.

* The `Wmicrosoft-template-shadow` warning has been suppressed in Clang.  I don't understand this warning or how to fix it "correctly".
* Numerous spelling errors in comments.
* Remove pointless `const` qualifier on return types
* Tweak friend class typedefs to make Clang happy.
* Suppress unused-value warning around the same block of code where a MSVC warning is suppressed.
* Reorder a few class fields to match the order in the constructor.  At least one of these was a legitimate bug (line 748)
* Add `override` in one place.

Note that warnings are suppressed using push/pop so the suppression will not leak outside this header.

There is another change that is more borderline:
* `FallbackError` telemetry event should specify the ProductAndServicePerformance tag.

I created a new test file called `TraceLoggingTests.cpp`.  This file contains no tests.  It does include `<wil/Tracelogging.h>` to ensure that _it compiles_ in the first place.  That is enough for now.  Prior to this none of the test collateral would compile this file so build breaks might not be caught before PR completion.

I ran `init_all.cmd`, `build_all.cmd`, and `runtests.cmd`.  The Clang|debug build was already broken before I began.  Clang|reldebuginfo works and was useful to validate these changes.  The tests had one preexisting failure.


dmachaj:
@asklar @bartekk8 because some of the forked edits were made by you.

dunhor:
> /azp run

FYI for future reference, the pipeline triggers on all PRs to master, so this is not needed

ChrisGuzak (approved):

PR #297, opened 2023-02-03
Introduce make_variant_bstr*

talagrand:
Introduce make_variant_bstr* to make a unique_variant that contains a BSTR in a single step

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -4616,6 +4616,37 @@ namespace wil
     }
 #endif // WIL_ENABLE_EXCEPTIONS
 
+    inline wil::unique_variant make_variant_bstr_nothrow(PCWSTR source) WI_NOEXCEPT
+    {
+        wil::unique_variant result{};
+        V_VT(result.addressof()) = VT_BSTR;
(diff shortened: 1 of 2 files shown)


talagrand:
I realize we could fall down the rabbit hole with all the possibilities here but there's tons of examples for VT_BSTR today in Windows.

include/wil/resource.h
```diff
@@ -4616,6 +4616,33 @@ namespace wil
     }
 #endif // WIL_ENABLE_EXCEPTIONS
 
+    inline wil::unique_variant make_variant_bstr_nothrow(PCWSTR source) WI_NOEXCEPT
```
dunhor, line 4619: Only comment I have - and it's mostly to solicit opinions from others - is that checking for failure isn't as simple as using something like `operator bool` on the return value, so I wonder if changing the return type to be `HRESULT` and have the `unique_variant` be an out param might be more "friendly."

On the flip side, if it's an out param, the value would always need to be cleared, even if it was just constructed, so my preference is to probably keep it this way.

dunhor (approved):

PR #305, opened 2023-02-27
Use wistd::min instead of Windows.h min

CasualPokePlayer:
The latter is not guarenteed to be available for some projects which have NOMINMAX defined somewhere before a Windows.h include.

### DIFF
--- a/include/wil/result_macros.h
+++ b/include/wil/result_macros.h
@@ -2261,7 +2261,7 @@ __WI_POP_WARNINGS
         {
             size_t cchLen = UntrustedStringLength(reinterpret_cast<TString>(pStart), (pEnd - pStart) / sizeof((*ppszBufferString)[0]));
             *ppszBufferString = (cchLen > 0) ? reinterpret_cast<TString>(pStart) : nullptr;
-            auto pReturn = min(pEnd, pStart + ((cchLen + 1) * sizeof((*ppszBufferString)[0])));
+            auto pReturn = (wistd::min)(pEnd, pStart + ((cchLen + 1) * sizeof((*ppszBufferString)[0])));
             __analysis_assume((pReturn >= pStart) && (pReturn <= pEnd));
             return pReturn;
         }


include/wil/result_macros.h
```diff
@@ -2261,7 +2261,7 @@ __WI_POP_WARNINGS
         {
             size_t cchLen = UntrustedStringLength(reinterpret_cast<TString>(pStart), (pEnd - pStart) / sizeof((*ppszBufferString)[0]));
             *ppszBufferString = (cchLen > 0) ? reinterpret_cast<TString>(pStart) : nullptr;
-            auto pReturn = min(pEnd, pStart + ((cchLen + 1) * sizeof((*ppszBufferString)[0])));
```
dmachaj, line 2264: If you update some or all of the test projects to define `NOMINMAX` do they exhibit this build break?  If so, consider adding that to this PR so that it does not regress in the future.

dunhor (approved):

include/wil/result_macros.h
dunhor: One negative of that is that you then would not get errors when you forget the parenthesis around `wistd::min`

dmachaj: Fair.  There are many test projects.  So long as _at least one_ defines `NOMINMAX` that should provide _some_ coverage.

CasualPokePlayer: Various tests seem to just have failures with NOMINMAX defined due to themselves relying on Windows.h min.

CasualPokePlayer: With the above "fixed", and the PR changes reverted, the tests still build fine. This is probably due to the delayed template parsing behavior of msvc and clang-cl matching that behavior. On an interesting note, the -fno-delayed-template-parsing passed to clang-cl in the build scripts doesn't actually do anything here, as it appears to only respect that flag when -fno-ms-compatibility is also passed (probably a clang bug?).

dmachaj (approved):

include/wil/result_macros.h
dunhor: Hmm, that's interesting for a couple reasons. The first is that, as you note, Clang appears to produce an error when `-fno-ms-compatibility` is specified, but doesn't when the argument is omitted, even when `-fno-delayed-template-parsing` is specified. The second, and arguably even more interesting thing, is that MSVC _also_ does not produce an error when building with `/permissive-` (which we do). My guess is that MSVC sees that a portion of the calculation is dependent on the type of `TString` and therefore does not issue an error (even though the types of `cchLen` and `ppszBufferString` are known). I cannot say with absolute certainty if the standard requires an error here or not (or even if it optionally allows an error).

If we change to a slightly different example where the arguments very obviously do not depend on the template type:
```c++
template <typename T>
std::ptrdiff_t doit(const char* begin, const char* end)
{
    return min(end - begin, 42);
}
```
Both [Clang](https://godbolt.org/z/Ts9W4xPf8) and [MSVC](https://godbolt.org/z/cnoYGz64j) correctly produce errors. What's more interesting is that Clang with `-fms-compatibility -fno-delayed-template-parsing` _still_ [does not produce an error](https://godbolt.org/z/aj5h71MaM) for this code. There might be a historical reason for this (e.g. something in the STL or SDK causing issues), but I'd like to think that such issues have since been taken care of and ideally this behavior can be controlled by the `delayed-template-parsing` flag.

CasualPokePlayer: > My guess is that MSVC sees that a portion of the calculation is dependent on the type of `TString` and therefore does not issue an error (even though the types of `cchLen` and `ppszBufferString` are known).

> function declaration must be available as none of the arguments depend on a template parameter

Considering MSVC gives this error for your later example, I'm inclined to believe you're right here.

PR #314, opened 2023-04-03
Update most non-void returning const member functions to be [[nodiscard]]

dunhor:
It's become a generally accepted pattern in C++ to make non-`void` returning `const` member functions be `[[nodiscard]]` (or in the case of Herb's cpp2, the default everywhere). The idea is that a `const` function is not modifying any state, and therefore the caller should be interested in anything the function returns. I also went ahead and addressed things in the `wistd` namespace (even though `std` doesn't have the attribute) with the thought that it should benefit callers in general. There are a couple of exceptions to this, though:
* Functions with "out"/"in out" parameters were kept as-is. This was to avoid breaking too much valid/correct code that either doesn't care about the return value or can deduce the return value by the side effects to the output parameter (e.g. functions that do a QI).
* Functions that may have irrelevant return values/types for some subset of input values. E.g. waiting with `INFINITE` time. This was also done to avoid breaking too much correct code.
* Functions that are "too generic" to the point where we can't make any assumptions about their semantics were left alone. E.g. `wistd::function::operator()`.

Functions were identified by a combination of `clang-query` and manual code inspection. The mostly correct matcher argument for `clang-query` that I used was:

```
m cxxMethodDecl(isConst(), isExpansionInFileMatching("include[/\\]wil[/\\].*"), unless(hasAttr("attr::WarnUnusedResult")), unless(ofClass(isLambda())), unless(isDeleted()), unless(returns(voidType())))
```

The list of `const` member functions that didn't have the attribute added are:
* `wistd::function::operator()` - can't statically determine the semantics of the call.
* `wil::event_t::wait` - callers can - and generally do - assume the call can't fail when called with `INFINITE` wait time.
* `wil::com_ptr_t::(try_)(query|copy)_to` - callers can - and sometimes do - deduce the return value from whether or not the output pointer is null.

This change does not:
* Address any non-`const` member functions with the exception of functions with the same name as a `const` function to reduce confusion (e.g. `begin`, etc.).
* Address any non-member or `static` functions, even though these may have similar semantics of not modifying state.
* Re-visit whether some non-`const` member functions can/should be made `const`.
* Do anything to prohibit violating this rule in the future. We'll just have to be vigilant going forward.

I chose not to tackle these to limit the scope. We can try and tackle these in another change, if desired.

The plan is to just `(void)` any existing offending calls that are found. The idea is more to catch bugs moving forward, not find/fix bugs that may or may not already exist.

### DIFF
--- a/tests/ComTests.cpp
+++ b/tests/ComTests.cpp
@@ -562,7 +562,7 @@ TEST_CASE("ComTests::Test_ConstPointer", "[com][com_ptr]")
 
     REQUIRE(spUnk.get() != nullptr);
     REQUIRE(spUnk);
-    spUnk.addressof();
+    (void)spUnk.addressof();
     spUnk.copy_to(spUnkHelper.addressof());
     spUnk.copy_to(spInspectable.addressof());
     spUnk.copy_to(IID_PPV_ARGS(&spInspectable));
@@ -801,7 +801,7 @@ void TestSmartPointer(const Ptr& ptr1, const Ptr& ptr2)
         auto address = ptr1.addressof();
         REQUIRE(*address == ptr1.get());
         (void)static_cast<bool>(ptr1);
-        ptr1.get();
+        (void)ptr1.get();
         auto deref = ptr1.operator->();
         (void)deref;
         if (ptr1)
@@ -1452,7 +1452,7 @@ void TestSmartPointerQueryIidPpv(wistd::true_type, const Ptr& source)       // i
         if (source)
         {
             DestPtr dest;
-            source.query_to(IID_PPV_ARGS(&dest));
+            REQUIRE_NOERROR(source.query_to(IID_PPV_ARGS(&dest)));
             REQUIRE_ERROR(source.query_to(IID_PPV_ARGS(&never)));
             REQUIRE((dest && !never));
         }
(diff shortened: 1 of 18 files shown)


tests/ComTests.cpp
```diff
@@ -1452,7 +1452,7 @@ void TestSmartPointerQueryIidPpv(wistd::true_type, const Ptr& source)       // i
         if (source)
         {
             DestPtr dest;
-            source.query_to(IID_PPV_ARGS(&dest));
+            REQUIRE_NOERROR(source.query_to(IID_PPV_ARGS(&dest)));
```
dunhor, line 1455: Note that these changes were made before I reverted the updates to `query_to` et. al. I kept these checks because they seemed good regardless

dunhor:
Argh. Looks like we're randomly exhausting device space with building tests now :/

jonwis:
Kinda wish the standard had gone the other direction... everything is nodiscard unless you say "it's ok to discard this." I guess that makes updating existing codebases harder.

dunhor:
> Kinda wish the standard had gone the other direction...

As you guess, the answer is back-compat. Of course, if you start fresh, [you can choose whatever defaults you want](https://youtu.be/ELeZAKCN4tY?t=1359)

ChrisGuzak (approved):
thanks Duncan.

PR #315, opened 2023-04-04
Fix natvis for wistd types

dunhor:
History: The `wistd::` types used to be copied from MSVC's STL, however the open sourcing of WIL predates the open sourcing of the STL and the licensing for the STL hadn't been figured out by the time we went public. Therefore, we switched these implementations over to be a derivative of libcpp.

Since that transition, nobody has updated the natvis file to reflect the new implementations. This change rectifies that. Note that:
1. I've only tested on windbg (which is kind of annoying because there's already a version that ships with the debugger, or at least for the internal debugger). I have not yet figured out how to test on VS
2. There seems to occasionally be an issue with `__compressed_pair`. If I inspect something like `wistd::unique_ptr<int>`, everything looks right and the deleter shows as `default_delete`. However, if I instead inspect another type that uses something like `function_deleter`, I see the error `Unexpected failure to dereference object`. I'm unsure why this happens. I'm curious if it somehow has to do with EBO getting in the way, but that wouldn't explain why something like `default_delete` works fine. The experience otherwise isn't that bad. It'll still show the rest of the info, just give the error for the deleter, which nobody going to inspect anyway, especially since it has no state.

Also, as mentioned above, there seems to be a version of the natvis that ships with the internal debugger (and maybe the public version). The file seems to be pulled from the OS repo, which I've never sync'd before, so I'll be sure to do that for the next one.

### DIFF
--- a/natvis/wil.natvis
+++ b/natvis/wil.natvis
@@ -8,46 +8,80 @@
     PARTICULAR PURPOSE AND NONINFRINGEMENT.
 -->
 <AutoVisualizer xmlns="http://schemas.microsoft.com/vstudio/debugger/natvis/2010">
-    <Type Name="wistd::_Func_impl&lt;*&gt;">
-        <DisplayString>{_Callee._Object}</DisplayString>
+    <Type Name="wistd::default_delete&lt;*&gt;">
+        <DisplayString>default_delete</DisplayString>
+        <Expand />
+    </Type>
+
+    <Type Name="wistd::__compressed_pair_elem&lt;*,*,0&gt;">
+        <Intrinsic Name="get" Expression="__value_" />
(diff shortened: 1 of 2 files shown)


dunhor:
> I have not yet figured out how to test on VS

Bit the bullet and added the linker command to include the natvis in the pdb. Looks better in VS; tempted to call it a windbg bug:

image

image

dunhor:
On second look I now see why there originally was a `unique_ptr<wchar_t[*],*>` specification. That doesn't bind in windbg, but I guess is required for VS...

dunhor:
Okay, here's what it looks like in VS now

image

ChrisGuzak (approved):

PR #316, opened 2023-04-05
Work around noexcept-in-type-system VS2017 bug

### LINKED ISSUE #302: VS2017 noexcept-in-typesystem bug is hitting with the definition of 'unique_private_namespace'
1>wb2.cpp
1>e:\code\mytest\wb2\packages\microsoft.windows.implementationlibrary.1.0.230202.1\include\wil\resource.h(2550): error C2440: “specialization”: 无法从“void (__stdcall *)(HANDLE) noexcept”转换为“unknown-type”
1>e:\code\mytest\wb2\packages\microsoft.windows.implementationlibrary.1.0.230202.1\include\wil\resource.h(2550): note: 上下文不允许消除重载函数的歧义

dunhor:
Fixes #302 

TL;DR - as I recall from vague understanding in the past - is that MSVC's "noexcept in the type system" update for C++17 "infected" pre-C++17 compilations, specifically for `decltype`. From what I recall when this initially started to impact us, `decltype` includes `noexcept` in the type, but that makes it impossible to assign/bind anything to such a function pointer since the `noexcept` gets dropped (as it should pre-C++17). VS2017 seems to be _even more_ broken than this as allegedly even building with C++17 as the standard produces an error.

This change explicitly declares the function pointer type so that the issue will not be observed with the buggy version of the compiler. Ideally, we could just drop the type and use `auto`, but that's also a C++17 feature :/

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -2547,7 +2547,7 @@ namespace wil
     }
 
     template <ULONG flags = 0>
-    using unique_private_namespace = unique_any_handle_null_only<decltype(details::ClosePrivateNamespaceHelper<flags>), &details::ClosePrivateNamespaceHelper<flags>>;
+    using unique_private_namespace = unique_any_handle_null_only<void(__stdcall*)(HANDLE) WI_PFN_NOEXCEPT, &details::ClosePrivateNamespaceHelper<flags>>;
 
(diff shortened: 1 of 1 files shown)


ChrisGuzak (approved):

jonwis (approved):
sad-panda.wav.tiff.mp4

PR #317, opened 2023-04-09
Add property, event, and INPC helpers

asklar:
This adds the following helpers:
- `read_only_property<T>`
- `read_write_property<T>`
- `simple_event<T>` (maps to an event that notifies a `EventHandler<T>`)
- `typed_event<S, T>` (maps to an event that notifies a `TypedEventHandler<S, T>`)
- `notify_property_changed<CRTP>`- base class for a default INotifyPropertyChanged
- `property_with_notify<T>` - a notifiable property

added tests. The INPC test requires instantiating the xaml type PropertyChangedEventArgs, which requires having XAML running, so that necessitates the test being either UWP or having a xaml island. Chose the latter, which means having a sxs manifest with maxversiontested. It also requires updating the command line to cppwinrt to include extensions, hence the `-input sdk+` now instead of `-input sdk`


include/wil/property.h
```diff
@@ -0,0 +1,226 @@
...
+   *  // modify MyInt
+   *  // MyInt = ...
+   *
+   *  // now send a notification to update the bound UI elements
+   *  RaisePropertyChanged(L"MyInt");
+   * }
+   * @endcode
+  */
+  auto RaisePropertyChanged(std::wstring_view name) {
+    return m_propertyChanged(self(), Xaml_Data_PropertyChangedEventHandler{ name });
```
sylveon, line 166: Could consider enabling this code to use deducing this when available.

include/wil/property.h
```diff
@@ -0,0 +1,226 @@
...
+//*********************************************************
+
+#pragma once
+#ifndef __WIL_PROPERTY_INCLUDED
+#define __WIL_PROPERTY_INCLUDED
+
+namespace wil
+{
+
+template <typename T>
```
dunhor, line 19: Please format the files properly

include/wil/property.h
```diff
@@ -0,0 +1,226 @@
...
+struct read_only_property
```
dunhor, line 20: I fail to see the benefit of the `read_only_property` and `read_write_property` types. If I want a read only property, I'll declare a `const` public member. If I want a read-write property, I'll declare a non-`const` public member.

Additionally, since this does nothing to prevent assignment, I'm pretty sure something like this will work:
```c++
obj->readOnlyProperty = read_only_property<int>(42);
```

include/wil/property.h
```diff
@@ -0,0 +1,226 @@
...
+    template<typename U, typename = std::enable_if_t<std::is_convertible_v<U, T>>>
+    read_write_property& operator=(read_write_property<U>&& other) noexcept
+    {
+        this->m_value = std::move(other.m_value);
+        return *this;
+    }
+};
+
+#pragma region WinRT / XAML helpers
+#ifdef WINRT_Windows_Foundation_H
```
dunhor, line 75: We should avoid adding more of this. It's a maintainability nightmare as the functionality is very sensitive to include order.

IMO all of this should just go into `cppwinrt_helpers.h` (and delete the property types).

include/wil/property.h
```diff
@@ -0,0 +1,226 @@
...
+
+namespace details {
+    template<typename T>
+    struct event_base {
```
dunhor, line 79: What does this do apart from slightly changing the interface of C++/WinRT's event types? I'm not sure changing `add`/`remove` -> `operator()` and `operator()` -> `invoke` really provides much value

include/wil/property.h
asklar: Thanks for the feedback. This is to implement WinRT properties as C++/WinRT expects properties to work.

In C++/WinRT, for a readonly property you have to declare a data member + method. For a read/write property, you have to declare a data member + 2 methods. This rapidly adds up to a bunch of repetitive boilerplate code. The types in this PR encapsulate the boilerplate.

dunhor: I see. In that case, it should _definitely_ move to a C++/WinRT specific header.

But that said, something like `read_write_property` seems very risky as it does not provide any mechanism for synchronization.

include/wil/property.h
asklar: cppwinrt_helpers is for winrt-consuming code (e.g. using the system dispatcher), this new header is for helpers for authoring winrt components, so renamed the file accordingly. Not all code that uses winrt will need to author components.

include/wil/property.h
asklar: same thing as the properties - this exposes a wrapper type that conforms to cppwinrt's expectations of an event's implementation

include/wil/property.h
asklar: renamed the header. 
note that nothing prevents you from continuing to write properties the way you do today, this just provides a low-cost abstraction to do away with all the boilerplate for 90% of cases.

include/wil/property.h
asklar: do you mean returning *this or some sfinae magic to remove the need for self(), or something else? if re: removing self, you still need it when deducing this isn't available so having the one impl is simpler and clearer.

include/wil/property.h
sylveon: The header is cppwinrt_helpers, not cppwinrt_consuming. Certainly it could container both?

include/wil/property.h
sylveon: I do agree with the overall feeling. Why not simply make a `propert<T>` type and to make it read-only have people write `const property<int> Foo;`

roxk: First thing first: I think allowing assignment to `read_only_property` is intentional? To allow write in implementation type, while disallowing write in projected type. Please correct me if I'm wrong here.

If this observation is right, the problem is that it seems `const property<int>` isn't writable even in implementation type. It'd behave like a constant, whereas `read_only_property` actually is read-only in projected type, but writable in implementation type (classic `get; private set;` property in C#).

Consider the following table for a typical winrt property's getter and setter. Public means callable in projected type. Private means callable only in implementation type.

**Current implementation**
|getter|setter|property to use|
|--|--|--|
|public|public| `read_write_property` |
|public|private| `read_only_property` (allows mutation via assignment) |
|private|public| Not supported |
|private|private| Don't need wrapper |

**Suggested**
|getter|setter|property to use|
|--|--|--|
|public|public| `property` |
|public|private| Nothing (From what I read, the suggested `const property<T>` _likely_ doesn't allow any mutation, private or not) |
|private|public| Not supported |
|private|private| Don't need wrapper|

This makes `read_only_property` more useful than `const property<int>`.

Granted, we can make `const property<int>` allow write via assignment as well - but that seems more awkward as the write has to work around C++'s type system using `mutable` and what not, whereas `read_only_property`'s `read-only-ness` is implicit in its overloading of `operator()`. To put it the other way, we don't really want the property to be read-only in C++'s type system, but only read-only in projected type (aka `get; private set`).

Maybe we just need a better name to indicate the property is only `read-only` in WinRT? e.g. `public_get_private_set_property`? `winrt_read_only_property`?

roxk:
This PR would be very useful in my effort to make WinRT component authoring idl-free via [idlgen](https://github.com/roxk/idlgen) (sorry for the shameless plug). Specifically, `idlgen` could recognize the property helpers in this PR and simplify a lot of boilerplate and attribute needed to tell whether `uint64_t Foo();` in C++ is a `UInt64 Foo{get;};` or `UInt64 Foo();` in idl.

include/wil/property.h
dunhor: Gotcha

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,222 @@
...
+
+        template<typename... TArgs>
+        read_only_property(TArgs&&... args) : m_value(std::forward<TArgs>(args)...) { }
+
+    protected:
+        T m_value;
+    };
+
+    template <typename T>
+    struct read_write_property : read_only_property<T>
```
dunhor, line 37: I'll move the discussion here since the original comment got resolved and the file name changed. The original comment was regarding the fact that this type does nothing to facilitate synchronization.

I understand that people can still manually write properties like they do today. My concern is more that it's too easy to see this - or perhaps read code that uses it - and assume that it's the "Microsoft recommended" way to implement properties. I could _maybe_ understand if this is intended for something like a class whose properties are only going to be read/written by XAML, though in that case (and the general case) I'd think that this _at least_ requires a different name. E.g. something like `single_threaded_read_write_property`.

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,222 @@
...
+        template<typename U, typename = std::enable_if_t<std::is_convertible_v<U, T>>>
+        read_write_property& operator=(read_write_property<U>&& other) noexcept
+        {
+            this->m_value = std::move(other.m_value);
+            return *this;
+        }
+    };
+
+#pragma region WinRT / XAML helpers
+#ifdef WINRT_Windows_Foundation_H
```
dunhor, line 74: If we're going to persist with this pattern, then we need to at least ensure that the header can be included multiple times so that functionality can "light up"

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,222 @@
...
+    template<typename T,
+        typename Xaml_Data_PropertyChangedEventHandler = winrt::Windows::UI::Xaml::Data::PropertyChangedEventHandler,
+        typename Xaml_Data_PropertyChangedEventArgs = winrt::Windows::UI::Xaml::Data::PropertyChangedEventArgs>
+    struct property_with_notify : read_write_property<T> {
+        using Type = T;
+
+        using read_write_property<T>::operator();
+
+        void operator()(const T& value) {
+            if (value != operator()()) {
```
dunhor, line 185: Isn't `m_value` `protected`? You should be able to just reference it directly

tests/CMakeLists.txt
```diff
@@ -73,12 +73,12 @@ add_subdirectory(win7)
 set(DEBUG_BUILD FALSE)
 set(HAS_DEBUG_INFO FALSE)
 
-if (${CMAKE_BUILD_TYPE} STREQUAL "Debug")
+#if (${CMAKE_BUILD_TYPE} STREQUAL "Debug")
```
dunhor, line 76: I can only assume that you were testing something?

asklar: ah yes, sorry : ) I actually wasn't able to get the repo building anymore using the instructions in the README... CMAKE_BUILD_TYPE seems to evaluate to empty string even though I call init_all / init.cmd with the right params. Perhaps something is out of sync in the readme?

dunhor: Interesting. That sounds like it _might_ be related to using MSBuild as the generator? If that's the case, we may need to disable the sanitizers under those circumstances. I'd expect the `init_all` script to work, though as that forces Ninja as the generator

asklar: Filed #318

include/wil/cppwinrt_authoring.h
ChrisGuzak, line 74: nit, the IDE will collapse the ifdefs, so I don't see much use in the #pragma parts.

include/wil/cppwinrt_authoring.h
asklar: renamed to:
- `single_threaded_ro_property`
- `single_threaded_rw_property`
- `single_threaded_notifying_property`

a little verbose IMO, but i can live with that and i'll probably alias those types in my usage to something shorter.

jonwis:
Neat - I've written that `simple_property` thing probably 5 times now.  Here's what I settled on, cribbed from a lost local repo (I can't recreate the event version, yours looks great.):

```c++
namespace wil
{
    /// @cond
    namespace details
    {
        template<typename T> struct prop_storage
        {
            T m_value{};
            operator T& () { return m_value; }
            operator T const& () const { return m_value; }
            template<typename Q> auto operator=(Q&& q) {
                m_value = wistd::forward<Q>(q);
                return *this;
            }
        };
    }
    /// @endcond

    /** Produces the property operator() adapters around a type
    C++/WinRT property syntax requires operator()(value) to set and operator()() to get. This type
    is both the storage of the underlying property and the operator() overloads for access. It can
    be used in the same way as the property type as simple_property<Q> is-a Q.
    ~~~
    struct MyType : MyTypeT<MyType> {
        wil::simple_property<winrt::hstring> DisplayName{ L"puppies" };
        wil::simple_property<bool> IsEnabled { false };
        void Flip() {
            if (IsEnabled) {
                DisplayName = L"kittens";
                IsEnabled = false;
            } else {
                DisplayName = L"puppies";
                IsEnabled = true;
            }
        }
    };
    ~~~
    */
    template<typename T> struct simple_property : 
        wistd::conditional_t<wistd::is_scalar_v<T>, details::prop_storage<T>, T>
    {
        template<typename Q> auto& operator()(Q&& q)
        {
            *this = std::forward<Q>(q);
            return *this;
        }

        auto& operator()()
        {
            return *this;
        }

        template<typename Q> auto operator=(Q&& q)
        { 
            return static_cast<T&>(*this) = std::forward<Q>(q); 
        }
    };
}
```

This "is-a" means `PropertyName` can be used just like the underlying type - so you can say things like `DisplayName.empty()` or similar.

I couldn't get a "glom on" version like this to work nicely because `property_adapter<>` still requires storage for the `T&`. It's _probably_ possible with some `offsetof` hilarity in a macro.

```c++
template<typename T> struct property_adapter
{
    T& m_source;

    template<typename Q> auto& operator()(Q&& q)
    {
        return m_source = std::forward<Q>(q);
    }

    auto& operator()()
    {
        return m_source;
    }
};

struct thing
{
    int m_value{3};
    property_adapter<int> TheValue{ m_value };
};
```

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,219 @@
...
+            single_threaded_rw_property<T>(std::forward<TArgs...>(args)...) {
+            m_name = name;
+            m_npc = npc;
+            m_sender = sender;
+        }
+
+        single_threaded_notifying_property(const single_threaded_notifying_property&) = default;
+        single_threaded_notifying_property(single_threaded_notifying_property&&) = default;
+    private:
+        std::wstring m_name;
```
jonwis, line 205: Is there a way to make this a method instead of storage on the instance?

Or should this store the `std::wstring_view` so it's not making a copy here, since it's a preprocessor-generated string?

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,219 @@
...
+        winrt::event<Xaml_Data_PropertyChangedEventHandler> m_propertyChanged;
+    };
+
+    /**
+     * @brief Implements a property type with notifications
+     * @tparam T the property type
+     * @details Use the #INIT_NOTIFY_PROPERTY macro to initialize this property in your class constructor. This will set up the right property name, and bind it to the `notify_property_changed_base` implementation.
+    */
+    template<typename T,
+        typename Xaml_Data_PropertyChangedEventHandler = winrt::Windows::UI::Xaml::Data::PropertyChangedEventHandler,
```
dmachaj, line 175: Is it possible to detect the presence of WinUI and provide the same helpers for Microsoft::UI::Xaml types as well?

asklar: sure : )

include/wil/cppwinrt_authoring.h
asklar: if we can guarantee RTTI, then we could rely on it, but that seems like a niche case since most usages turn it off. using wstring_view seems like it might work though.

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,236 @@
...
+        std::wstring_view m_name;
+        winrt::event<Xaml_Data_PropertyChangedEventHandler>* m_npc;
+        winrt::Windows::Foundation::IInspectable m_sender;
+    };
+
+    /**
+    * @def INIT_NOTIFY_PROPERTY
+    * @brief use this to initialize a wil::single_threaded_notifying_property in your class constructor.
+    */
+#define INIT_NOTIFY_PROPERTY(NAME, VALUE)  \
```
dmachaj, line 230: It would be great if there were also type-checked macros for raising events as well as handling them.  Everything is a string so both of those types of code are subject to typo errors that will not be flagged by the compiler.

dmachaj: I missed that changing the property automatically raises the event (with the correct name!) so that is covered.  That just leaves event handlers.  This is "nice to have" but certainly non-blocking.

The ask is to have something like this:
```cpp
void OnPropertyChanged(const winrt::IInspectable& /*sender*/, const winrt::PropertyChangedEventArgs& args)
{
    // typo-safe property change handling
    if (args.propertyName() == NAMEOF(MyRuntimeClass, MyPropertyName))
    {
        // Handle property changed
    }

    // unsafe property change handling
    if (args.propertyName() == "MyPropertyNameWithTypo")
    {
        // Will never get called because I can't spell
    }
```

asklar: I'll send a separate PR for that since this feature is somewhat orthogonal to properties/notifications (though it is useful in the context of). See draft #321

dmachaj (approved):

asklar:
@dunhor looks like everything is green now, let me know if you have any more feedback, or if I need to press anything to merge : )

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,236 @@
...
+#ifndef __WIL_CPPWINRT_AUTHORING_INCLUDED
+#define __WIL_CPPWINRT_AUTHORING_INCLUDED
+    namespace details
+    {
+        template<typename T>
+        struct single_threaded_property_storage
+        {
+            T m_value{};
+            single_threaded_property_storage() = default;
+            single_threaded_property_storage(const T& value) : m_value(value) {}
```
dunhor, line 23:
```suggestion
            single_threaded_property_storage(T value) : m_value(std::move(value)) {}
```

Force the caller to do the copy and then move that value into the local member. This has a downside if the type is not movable, however that's _probably_ not the case for any type that this will apply to. An alternative would be to provide two constructors: one taking `const T&` and another taking `T&&`.

If you want to go off the crazy deep end, accept `Args&&...` and forward-construct the member.

dunhor: Though it looks like this is only used for scalar types? If so, this suggestion as well as the templated `operator=` below seems like it's probably overkill.

That said, the general idea of the suggestion applies below.

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,236 @@
...
+                return *this;
+            }
+        };
+    }
+
+    template <typename T>
+    struct single_threaded_ro_property : std::conditional_t<std::is_scalar_v<T>, wil::details::single_threaded_property_storage<T>, T>
+    {
+        single_threaded_ro_property() = default;
+        single_threaded_ro_property(const T& t) : base_type(t) { }
```
dunhor, line 38:
```suggestion
        single_threaded_ro_property(T value) : base_type(std::move(value)) { }
```
See above for more details

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,236 @@
...
+            template<typename Q> auto operator=(Q&& q)
+            {
+                m_value = wistd::forward<Q>(q);
+                return *this;
+            }
+        };
+    }
+
+    template <typename T>
+    struct single_threaded_ro_property : std::conditional_t<std::is_scalar_v<T>, wil::details::single_threaded_property_storage<T>, T>
```
dunhor, line 35: Given that this is intended only for WinRT properties, this is _probably fine_, but in general this should also be guarded by an `is_final` check as well

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,236 @@
...
+        {
+            return static_cast<T&>(*this) = std::forward<Q>(q);
+        }
+    };
+
+#endif // __WIL_CPPWINRT_AUTHORING_INCLUDED
+
+#if !defined(__WIL_CPPWINRT_AUTHORING_INCLUDED_FOUNDATION) && defined(WINRT_Windows_Foundation_H) // WinRT / XAML helpers
+#define __WIL_CPPWINRT_AUTHORING_INCLUDED_FOUNDATION
+    namespace details {
```
dunhor, line 78:
```suggestion
    namespace details
    {
```
This and below

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,236 @@
...
+        template<typename T>
+        struct event_base {
+            winrt::event_token operator()(T&& handler) {
+                return m_handler.add(std::forward<T>(handler));
+            }
+            auto operator()(const winrt::event_token& token) noexcept { return m_handler.remove(token); }
```
dunhor, line 84: Does `winrt::event` not throw if the token is not found?

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,236 @@
...
+     *         //   event Windows.Foundation.TypedEventHandler<ModalPage, String> OkClicked;
+     *         wil::typed_event<MarkupSample::ModalPage, winrt::hstring> OkClicked;
+     * @endcode
+    */
+    template<typename TSender, typename TArgs>
+    struct typed_event : wil::details::event_base<winrt::Windows::Foundation::TypedEventHandler<TSender, TArgs>> {};
+
+#endif // !defined(__WIL_CPPWINRT_AUTHORING_INCLUDED_FOUNDATION) && defined(WINRT_Windows_Foundation_H)
+
+#if !defined(__WIL_CPPWINRT_AUTHORING_INCLUDED_XAML_DATA) // INotifyPropertyChanged helpers
```
dunhor, line 117: Was this intended to contain a header guard as well?

dunhor: I see you have guards below, however this will cause include order dependency issues if this header is included before either of those

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,236 @@
...
+#define __WIL_CPPWINRT_AUTHORING_INCLUDED_XAML_DATA
+namespace details
```
dunhor, line 119:
```suggestion
    namespace details
    {
#ifdef WINRT_Microsoft_UI_Xaml_Data_H
        using Xaml_Data_PropertyChangedEventHandler = winrt::Microsoft::UI::Xaml::Data::PropertyChangedEventHandler;
        using Xaml_Data_PropertyChangedEventArgs = winrt::Microsoft::UI::Xaml::Data::PropertyChangedEventArgs;
#elif defined(WINRT_Windows_UI_Xaml_Data_H)
        using Xaml_Data_PropertyChangedEventHandler = winrt::Windows::UI::Xaml::Data::PropertyChangedEventHandler;
        using Xaml_Data_PropertyChangedEventArgs = winrt::Windows::UI::Xaml::Data::PropertyChangedEventArgs;
#endif
    }
```
Indentation

dunhor: Oh Github... Why would I ever want to suggest multi-line edits

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,236 @@
...
+{
+#ifdef WINRT_Microsoft_UI_Xaml_Data_H
+    using Xaml_Data_PropertyChangedEventHandler = winrt::Microsoft::UI::Xaml::Data::PropertyChangedEventHandler;
+    using Xaml_Data_PropertyChangedEventArgs = winrt::Microsoft::UI::Xaml::Data::PropertyChangedEventArgs;
+#elif defined(WINRT_Windows_UI_Xaml_Data_H)
+    using Xaml_Data_PropertyChangedEventHandler = winrt::Windows::UI::Xaml::Data::PropertyChangedEventHandler;
+    using Xaml_Data_PropertyChangedEventArgs = winrt::Windows::UI::Xaml::Data::PropertyChangedEventArgs;
+#endif
+}
```
dunhor, line 128:
```suggestion
    namespace details
    {
#ifdef WINRT_Microsoft_UI_Xaml_Data_H
        using Xaml_Data_PropertyChangedEventHandler = winrt::Microsoft::UI::Xaml::Data::PropertyChangedEventHandler;
        using Xaml_Data_PropertyChangedEventArgs = winrt::Microsoft::UI::Xaml::Data::PropertyChangedEventArgs;
#elif defined(WINRT_Windows_UI_Xaml_Data_H)
        using Xaml_Data_PropertyChangedEventHandler = winrt::Windows::UI::Xaml::Data::PropertyChangedEventHandler;
        using Xaml_Data_PropertyChangedEventArgs = winrt::Windows::UI::Xaml::Data::PropertyChangedEventArgs;
#endif
    }
```
Nifty. They just made it impossibly difficult to discover this feature

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,236 @@
...
+     * struct MyPage : MyPageT<MyPage>, wil::notify_property_changed_base<MyPage> {
+     *      wil::single_threaded_notifying_property<int> MyInt;
+     *
+     *      MyPage() : INIT_PROPERTY(MyInt, 42) { }
+     * };
+     * @endcode
+    */
+    template<typename T,
+        typename Xaml_Data_PropertyChangedEventHandler = wil::details::Xaml_Data_PropertyChangedEventHandler
+        // , typename = std::enable_if_t<std::is_convertible_v<T, winrt::Windows::Foundation::IInspectable>>
```
dunhor, line 146: SFINAE on a type doesn't really make sense. You _can_ do tricks like having `bool` template parameters or the `void_t` trick, but that's probably beyond what the intent is here. This is probably best done as a `static_assert`

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,236 @@
...
+    >
+    struct notify_property_changed_base {
```
dunhor, line 148:
```suggestion
    struct notify_property_changed_base
    {
```

dunhor: And throughout, in general

tests/CMakeLists.txt
```diff
@@ -42,8 +42,8 @@ endif()
 # functions not marked 'inline'. The reason we prefer 'RelWithDebInfo' over 'Release' is to get debug info, so manually
 # revert to the desired (and default) inlining behavior as that exercises more interesting code paths
 if (${CMAKE_BUILD_TYPE} STREQUAL "RelWithDebInfo")
-    # TODO: This is currently blocked by an apparent Clang bug: https://github.com/llvm/llvm-project/issues/59690
-    # replace_cxx_flag("/Ob1" "/Ob2")
+   # TODO: This is currently blocked by an apparent Clang bug: https://github.com/llvm/llvm-project/issues/59690
+   # replace_cxx_flag("/Ob1" "/Ob2")
```
dunhor, line 46:
```suggestion
    # TODO: This is currently blocked by an apparent Clang bug: https://github.com/llvm/llvm-project/issues/59690
    # replace_cxx_flag("/Ob1" "/Ob2")
```
Revert the whitespace changes

tests/CMakeLists.txt
```diff
@@ -77,7 +77,7 @@ if (${CMAKE_BUILD_TYPE} STREQUAL "Debug")
     set(DEBUG_BUILD TRUE)
     set(HAS_DEBUG_INFO TRUE)
 elseif(${CMAKE_BUILD_TYPE} STREQUAL "RelWithDebInfo")
-    set(HAS_DEBUG_INFO TRUE)
+   set(HAS_DEBUG_INFO TRUE)
```
dunhor, line 80:
```suggestion
    set(HAS_DEBUG_INFO TRUE)
```

include/wil/cppwinrt_authoring.h
asklar: no, it is a no-op if it's the wrong token.

include/wil/cppwinrt_authoring.h
asklar: this isn't sfinae, this is setting it up so that it can be instantiated with non-default types (e.g. if/when xaml and winui can coexist and you need to use mix-match types)

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,247 @@
...
+    struct single_threaded_ro_property : std::conditional_t<std::is_scalar_v<T> || std::is_final_v<T>, wil::details::single_threaded_property_storage<T>, T>
+    {
+        single_threaded_ro_property() = default;
+        single_threaded_ro_property(T value) : base_type(std::move(value)) { }
+        
+        const auto& operator()()
+        {
+            return *this;
+        }
+        template<typename Q> auto operator=(Q&& q) = delete;
```
jonwis, line 44: Doesn't deleting this nerf the ability of the implementation object to use this type for its own purposes?  This is a "construct only" property, not a "read only" property. I want to write this:

```c++
// IDL
runtimeclass MyType {
    String DisplayName { get; };
    bool TrySetDisplayName(String displayName);
}

// Implementation
struct MyType : MyTypeT<MyType> {
    wil::single_threaded_ro_property<winrt::hstring> DisplayName;
    bool TrySetDisplayName(winrt::hstring const& id) {
        if (NumberOfSecondsIsEven()) {
            DisplayName = id;
            return true;
        } else {
            return false;
        }        
    }
};
```

There's even a test that verifies this is blocked:

```c++
TEST_CASE("CppWinRTAuthoringTests::InStruct", "[property]")
{
    struct TestStruct
    {
        wil::single_threaded_ro_property<int> Prop1{ 42 };
        wil::single_threaded_rw_property<int> Prop2{ 1 };
        wil::single_threaded_ro_property<int> Prop3{ 44 };
    };

    TestStruct test;
    static_assert(!std::is_assignable_v<wil::single_threaded_ro_property<int>, int>, "cannot assign to a readonly property");
```

jonwis: I _think_ you mean "single_threaded_get_only" or "single_threaded_property" ?

jonwis (changes requested):

include/wil/cppwinrt_authoring.h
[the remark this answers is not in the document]
asklar: older iteration, already changed :)

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,236 @@
...
+{
+#ifndef __WIL_CPPWINRT_AUTHORING_INCLUDED
+#define __WIL_CPPWINRT_AUTHORING_INCLUDED
+    namespace details
+    {
+        template<typename T>
+        struct single_threaded_property_storage
+        {
+            T m_value{};
+            single_threaded_property_storage() = default;
```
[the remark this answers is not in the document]
asklar: it's useful for making it easier for tooling to treat the C++ code as the source of truth for generating metadata

include/wil/cppwinrt_authoring.h
jonwis: I don't understand? I want to write that type I specified above - there's a `DisplayName` read only property, in that the property isn't settable by the caller _except_ in calling the _method_ `TrySetDisplayName`.

The IDL is the _interface_ definition - what people can do with the object. It shouldn't dictate how you _implement_ the object.

It's fine if you want to call this `single_threaded_ro_property` so an AST walker can derive the IDL, but don't block the _implementation_ from updating its value.

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,247 @@
...
+            m_sender(sender)
+        {}
+
+        single_threaded_notifying_property(const single_threaded_notifying_property&) = default;
+        single_threaded_notifying_property(single_threaded_notifying_property&&) = default;
+        std::wstring_view Name() const noexcept { return m_name; }
+    private:
+        std::wstring_view m_name;
+        winrt::event<Xaml_Data_PropertyChangedEventHandler>* m_npc;
+        winrt::Windows::Foundation::IInspectable m_sender;
```
dunhor, line 236: Isn't the sender typically `this`? So wouldn't this keep a circular strong reference?

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,247 @@
...
+        single_threaded_ro_property() = default;
+        single_threaded_ro_property(T value) : base_type(std::move(value)) { }
+        
+        const auto& operator()()
+        {
+            return *this;
+        }
+        template<typename Q> auto operator=(Q&& q) = delete;
+    private:
+        using base_type = std::conditional_t<std::is_scalar_v<T>, wil::details::single_threaded_property_storage<T>, T>;
```
dunhor, line 46:
```suggestion
        using base_type = std::conditional_t<std::is_scalar_v<T> || std::is_final_v<T>, wil::details::single_threaded_property_storage<T>, T>;
```

include/wil/cppwinrt_authoring.h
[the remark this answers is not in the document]
asklar: XAML objects are much bigger than 4 pointer sizes anyway :)
Do you have an alternate proposal you'd like to put forth?

asklar: `this` in this context would refer to the property. The sender is the object which hosts the property (e.g. a Panel, that has a property `Width`, when the property changes, the sender is the panel, not the `Width` property)

dunhor: Sorry, when I say "this" I'm referring to the object hosting the property. E.g. if class `Foo` has a member of type `single_threaded_notifying_property`, then my understanding is that the sender (and what I call "this") is the `Foo` object.

If so, that's a circular reference. The `Foo` object has a member `single_threaded_notifying_property`, which is holding a strong reference to the `Foo` object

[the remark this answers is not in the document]
dunhor: Saying "the types are already large, so who cares if we make them larger" isn't a good mentality to have. We should be actively trying to make them smaller, if anything. Or at the very least, not making them larger when they don't have to be.

> Do you have an alternate proposal you'd like to put forth?

As much as I hate macros, that frankly seems like the best option that I can see.

asklar: the latest iteration already converted it to be a weak_ref

JaiganeshKumaran:
I would also recommend adding a CTAD guide so that you can avoid writing the property type (only works if it's static).

Example:
```cpp
inline static single_threaded_rw_property MyInt { 10 };
```

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,284 @@
...
+            static_cast<base_type&>(*this) = std::forward<Q>(q);
+            return *this;
+        }
+    };
+
+    template <typename T>
+    struct single_threaded_rw_property : single_threaded_property<T>
+    {
+        using base_type = single_threaded_property<T>;
+        template<typename... TArgs> single_threaded_rw_property(TArgs... value) : base_type(std::move(value)...) {}
```
dunhor, line 64: Two things here.

First, is there a reason for the disparity between accepting `TArgs...` here, but `T` for `single_threaded_property`?

Second, for a forwarding constructor, you should accept the arguments by `TArgs&&...` and use `std::forward<TArgs>(args)...`

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,284 @@
...
+    auto name() const noexcept { return m_##name; }         \
+    auto& name(type value)                                  \
+    {                                                       \
+        if (m_##name != value)                              \
+        {                                                   \
+            m_##name = std::move(value);                    \
+            RaisePropertyChanged(L#name);                   \
+        }                                                   \
+        return *this;                                       \
+    }                                                       \
```
dunhor, line 274:
```suggestion
#define WIL_NOTIFYING_PROPERTY(type, name, ...)             \
private:                                                    \
    type m_##name{__VA_ARGS__};                             \
public:                                                     \
    auto name() const noexcept { return m_##name; }         \
    auto& name(type value)                                  \
    {                                                       \
        if (m_##name != value)                              \
        {                                                   \
            m_##name = std::move(value);                    \
            RaisePropertyChanged(L"" #name);                \
        }                                                   \
        return *this;                                       \
    }                                                       \
```
Note that `L#name` isn't valid. We could also use `protected` over `private`, but this should be fine unless a legitimate need arises.

tests/CppWinRTAuthoringTests.cpp
```diff
@@ -0,0 +1,208 @@
...
+    test.MyEvent(token);
+
+    auto token2 = test.MyTypedEvent([](winrt::Windows::Foundation::IInspectable, int args) { REQUIRE(args == 42); });
+    test.MyTypedEvent.invoke(nullptr, 42);
+    test.MyTypedEvent(token2);
+}
+#endif // WINRT_Windows_Foundation_H
+
+// if using MSVC since we need the embedded manifest bits to use XAML islands
+#if _MSVC_LANG >= 201703L && defined(WINRT_Windows_UI_Xaml_Data_H)
```
dunhor, line 153: Note that this is _not_ checking for MSVC as `clang-cl` also provides a definition for `_MSVC_LANG`

include/wil/cppwinrt_authoring.h
asklar: L#x is valid. https://godbolt.org/z/73jWs4vv8

dunhor: [No it isn't](https://godbolt.org/z/r1797ar6b)

MSVC's preprocessor support is non-standard. There are flags to enable standards conformance, though

dunhor: Some more info: https://learn.microsoft.com/en-us/cpp/preprocessor/preprocessor-experimental-overview?view=msvc-170#lval

asklar: huh, TIL. Fixed.

tests/CppWinRTAuthoringTests.cpp
asklar: actually this shouldn't be needed now that I'm embedding the manifest via cmake in the first place

include/wil/cppwinrt_authoring.h
```diff
@@ -0,0 +1,284 @@
...
+
+        template<typename Q> auto& operator=(Q&& q)
+        {
+            static_cast<base_type&>(*this) = std::forward<Q>(q);
+            return *this;
+        }
+    };
+
+    template <typename T>
+    struct single_threaded_rw_property : single_threaded_property<T>
```
jonwis, line 61: I'm having a hard time telling the difference between this and the non-rw version. I presume it's for your eventual authoring support, where you generate the get/set based on whether it's rw or non-rw?

jonwis (approved):

include/wil/cppwinrt_authoring.h
asklar: yep exactly, that way we can easily differentiate them when we get tools to reason over them

citelao:
@asklar @jonwis, is there any documentation in the README about how to use these helpers? They look cool.

PR #324, opened 2023-05-09
wil::winrt_conditionally_implements for conditionally implementing an interface

oldnewthing:
This lets you selective hide interfaces on an object.

### DIFF
--- a/include/wil/cppwinrt.h
+++ b/include/wil/cppwinrt.h
@@ -403,6 +403,89 @@ namespace wil
             --winrt::get_module_lock();
         }
     };
+
+    /** Implements a C++/WinRT class where some interfaces are conditionally supported.
+    ~~~~
+    // Assume the existence of a class "Version2" which says whether
+    // the IMyThing2 interface should be supported.
+    struct Version2 { static bool IsEnabled(); };
+
+    // Declare implementation class which conditionally supports IMyThing2.
+    struct MyThing : wil::winrt_conditionally_implements<MyThingT<MyThing>,
+                         Version2, IMyThing2>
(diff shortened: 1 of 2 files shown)


include/wil/cppwinrt.h
```diff
@@ -403,6 +403,89 @@ namespace wil
...
+        }
+
+        template<std::size_t index, std::size_t upto, typename Tuple>
+        static constexpr void check_no_duplicates()
+        {
+            if constexpr (index < upto)
+            {
+                static_assert(!std::is_same_v<std::tuple_element_t<index, Tuple>, std::tuple_element_t<upto, Tuple>>,
+                    "Duplicate interfaces found in winrt_conditionally_implements");
+                check_no_duplicates<index + 2, upto, Tuple>();
```
jonwis, line 485: Feature creep - what if the "is enabled" method is constexpr? Is there a way to exclude the vtable entirely by filtering the `Rest...` somehow in that case?

jonwis (approved):
Looks fine, could use a test to verify though.

ChrisGuzak (approved):

include/wil/cppwinrt.h
oldnewthing: In that case, you don't need `winrt_conditionally_implements` at all.

```cpp
    struct TestClass : wil::winrt_conditionally_implements<
        winrt::implements<TestClass, std::conditional_t<also_closable, IClosable, std::true_type> >
```

If you mean that the template could autoconvert constexpr IsEnabled() into std::conditional_t, um, yeah, nice feature request.

PR #325, opened 2023-05-11
co_await resume_any_apartment(async) resumes in any apartment

### LINKED ISSUE #274: Consider adding no_switch await adapter for IAsyncInfo
*Originally posted on C++/WinRT [here](https://github.com/microsoft/cppwinrt/issues/1217#issuecomment-1289319275)*

As far as I can see (please correct me if I'm wrong) there is no equivalent of C#'s Task.ConfigureAwait(false) built-in to C++/WinRT when awaiting IAsyncInfo types. There is a sample of how to do it in the doc here https://learn.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/concurrency-2 (look for no_switch) and it's also in this 'The Old New Thing' blog post - https://devblogs.microsoft.com/oldnewthing/20191216-00/?p=103217. I wonder if it would be a good idea to add this to WIL? C# recommends that all libraries should use ConfigureAwait(false) (where it doesn't matter about running on the UI thread) and I assume that we should be doing the same in C++/WinRT so a no_switch await adapter ought to be widely used. (And any non-UI OS components not using it or an equivalent when awaiting IAsyncInfo should probably add it).

oldnewthing:
This lets you override the default "return to the original apartment" behavior of `co_await`'ing an `IAsyncInfo`.

This is the C++/WinRT counterpart to C# `ConfigureAwait(false)`.

This is basically a copy of the C++/WinRT awaiter, but with the one line of apartment-switching code deleted. Unfortunately, it's a good amount of code that has to be duplicated.

The code is made a bit more complicated by the fact that it has to deal with two sets of client options:

* C++/WinRT version 230207 introduced a breaking change in how custom awaiters support cancellation propagation, so we generate different cancellation support depending on which C++/WinRT version you are using.
* Avoiding a memory corruption bug in MSVC pre-standardization coroutines. This is fixed in VS version 16.11, but pre-16.11 and post-16.11 pre-standardization coroutines are ABI compatible, so changing behavior based on the VS version could lead to ODR violations. Instead, we key off the ABI breaking change in MSVC standardized coroutines in C++20, since the ABI break prevents you from linking pre-standardized coroutines with standardized coroutines. (This is the same solution that C++/WinRT itself uses.)

The `resume_any_apartment` helper activates if you include `cppwinrt_helpers.h` after `winrt/Windows.Foundation.h`. You can include `cppwinrt_helpers.h` multiple times, and any new features are activated based on whatever headers you've included since the previous time you include `cppwinrt_helpers.h`.

Fixes #274

### DIFF
--- a/include/wil/cppwinrt_helpers.h
+++ b/include/wil/cppwinrt_helpers.h
@@ -212,6 +212,150 @@ namespace wil::details
 #endif // __WIL_CPPWINRT_MICROSOFT_UI_DISPATCHING_HELPERS
 /// @endcond
 
+#if defined(WINRT_Windows_Foundation_H) && !defined(__WIL_CPPWINRT_WINDOWS_FOUNDATION_HELPERS)
+#define __WIL_CPPWINRT_WINDOWS_FOUNDATION_HELPERS
+/// @cond
+namespace winrt
+{
+    // Can get rid of this pair of forward declarations once we abandon support for C++/WinRT versions older than 230207.
+    // Look for v2_await_cancellation for other code that can be simplified.
+    template<typename Derived> struct cancellable_awaiter;
+    struct enable_await_cancellation;
+}
+
+namespace wil::details
+{
+    // Detect whether the C++/WinRT version uses v1 or v2 cancellation.
+    constexpr bool v2_await_cancellation = minor_version_from_string(CPPWINRT_VERSION) >= 230207;
+    template<typename Derived>
+    using winrt_cancellation_base = std::conditional_t<v2_await_cancellation, winrt::cancellable_awaiter<Derived>, winrt::enable_await_cancellation>;
+
+    template<typename Awaiter>
+    struct agile_disconnect_aware_handler
+    {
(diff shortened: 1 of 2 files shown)


kennykerr (approved):
Seems ok although it's hard to review as its very complicated. Can we not drop support for older implementations of coroutines?

oldnewthing:
> Seems ok although it's hard to review as its very complicated. Can we not drop support for older implementations of coroutines?

Alternatively, we could incorporate this into C++/WinRT, allowing us to take advantage of existing infrastructure. (It would turn into a boolean template parameter to async_awaiter that bypasses the `resume_apartment` call.)

We still have products in support that are locked to older versions of C++/WinRT. But then again, if you can't update C++/WinRT, you probably can't update wil either?

kennykerr:
Yes, I'd be ok with update cppwinrt if that'll be simpler.

jonwis (approved):

oldnewthing:
@jonwis I'm going to withdraw the PR for now. Looking at making this part of core C++/WinRT instead.

PR #336, opened 2023-06-19
Update README.md

keith-horton:
Added a statement on the new registry functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Some things that WIL includes to whet your appetite:
   that save you the work of manually specifying buffer sizes, calling a function twice
   to get the needed buffer size and then allocate and pass the right-size buffer,
   casting or converting between types, and so on.
-- [`include/wil/registry.h`](include/wil/registry.h): Type safe functions to read from, write to,
+- [`include/wil/registry.h`](include/wil/registry.h): Type-safe functions to read from, write to,
   and watch the registry. Registry watchers that can call a lambda function or a callback function
   you provide whenever a certain tree within the Windows registry changes.
 - [`include/wil/result.h`](include/wil/result.h)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,9 +17,9 @@ Some things that WIL includes to whet your appetite:
   that save you the work of manually specifying buffer sizes, calling a function twice
   to get the needed buffer size and then allocate and pass the right-size buffer,
   casting or converting between types, and so on.
-- [`include/wil/registry.h`](include/wil/registry.h): Registry watchers that can
-  call a lambda function or callback you provide whenever a certain tree within
-  the Windows registry changes.
+- [`include/wil/registry.h`](include/wil/registry.h): Type safe functions to read from, write to,
```
citelao, line 20:
```suggestion
- [`include/wil/registry.h`](include/wil/registry.h): Type-safe functions to read from, write to,
```

README.md
```diff
@@ -17,9 +17,9 @@ Some things that WIL includes to whet your appetite:
...
+  and watch the registry. Registry watchers that can call a lambda function or a callback function
```
citelao, line 21:
```suggestion
  and watch the registry. Also, registry watchers that can call a lambda function or a callback function
```

citelao (approved):
Some nits :)

jonwis (approved):

dunhor (approved):

README.md
```diff
@@ -17,9 +17,9 @@ Some things that WIL includes to whet your appetite:
   that save you the work of manually specifying buffer sizes, calling a function twice
   to get the needed buffer size and then allocate and pass the right-size buffer,
   casting or converting between types, and so on.
-- [`include/wil/registry.h`](include/wil/registry.h): Registry watchers that can
-  call a lambda function or callback you provide whenever a certain tree within
-  the Windows registry changes.
+- [`include/wil/registry.h`](include/wil/registry.h): Type-safe functions to read from, write to,
```
dunhor, line 20: Generic comment, not specific to this change (& I don't suggest changing it here): we should add line breaks after each sentence instead of concatenating them all together and breaking at some character width in the file (we can still break long sentences if we don't want to rely on word wrap, but we should still break on punctuation). Ultimately this adds a sentence and a word, but it's harder to see that just looking at the diff.

PR #341, opened 2023-06-23
Additional test coverage for restricted error info

jonwis:
In #337, @dmachaj asked for specific tests on the C++/WinRT and C++ integration for error origination.  This change adds that coverage for posterity.

### DIFF
--- a/tests/CppWinRTTests.cpp
+++ b/tests/CppWinRTTests.cpp
@@ -15,6 +15,7 @@ using namespace winrt::Windows::ApplicationModel::Activation;
 
 #include "catch.hpp"
 #include <roerrorapi.h>
+#include "common.h"
 
 // HRESULT values that C++/WinRT throws as something other than winrt::hresult_error - e.g. a type derived from
 // winrt::hresult_error, std::*, etc.
@@ -651,6 +652,7 @@ TEST_CASE("CppWinRTTests::ResumeForegroundTests", "[cppwinrt]")
 TEST_CASE("CppWinRTTests::ThrownExceptionWithMessage", "[cppwinrt]")
 {
     SetRestrictedErrorInfo(nullptr);
+
     []()
     {
         try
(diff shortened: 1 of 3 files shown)


tests/CppWinRTTests.cpp
```diff
@@ -646,3 +647,26 @@ TEST_CASE("CppWinRTTests::ResumeForegroundTests", "[cppwinrt]")
...
 #endif // coroutines
+
+TEST_CASE("CppWinRTTests::ThrownExceptionWithMessage", "[cppwinrt]")
+{
+    SetRestrictedErrorInfo(nullptr);
+    []()
+    {
+        try
+        {
+            throw winrt::hresult_access_denied(L"Puppies not allowed");
```
dmachaj, line 658: Puppies should never be denied access :p

tests/ResultTests.cpp
```diff
@@ -578,6 +578,20 @@ TEST_CASE("ResultTests::AutomaticOriginationOnFailure", "[result]")
     REQUIRE(S_FALSE == GetRestrictedErrorInfo(&restrictedErrorInformation));
 }
 
+void ExpectRestrictedError(HRESULT err, LPCWSTR text)
+{
+    wil::com_ptr_nothrow<IRestrictedErrorInfo> errorInfo;
+    REQUIRE_SUCCEEDED(GetRestrictedErrorInfo(&errorInfo));
```
dmachaj, line 584: Perhaps worth a comment regarding the (extremely) nonobvious behavior that calling `GetRestrictedErrorInfo` removes the error information for the current thread.  That leaves it clean for the next test case.

tests/CppWinRTTests.cpp
dmachaj, line 658: The origination from cppwinrt can vary based on the mechanism through which it is thrown (although on cpp20 and post-September-2022 it should be far more consistent).  Consider also validating `winrt::check_hresult(E_ACCESSDENIED, L"Puppies are wonderful");`

dmachaj (approved):

dunhor (approved):

PR #342, opened 2023-06-27
fix a potential compile error: 'PropVariantInit': undeclared identifier

Demonese:
Although generally no developer will disable language extensions - this is really asking for trouble.
We still need to respect the definition of PROPVARIANT and PropVariantInit in the Windows SDK header files.

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -4589,7 +4589,10 @@ namespace wil
 
 #if defined(__propidl_h__) && !defined(_WIL__propidl_h__) && !defined(WIL_KERNEL_MODE)
 #define _WIL__propidl_h__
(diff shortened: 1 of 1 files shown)


dunhor:
For those interested, as I was, the definition for `PropVariantInit` looks like the following in the SDK:
```c++
#if defined(_MSC_EXTENSIONS)
#define _PROPVARIANTINIT_DEFINED_
#   ifdef __cplusplus
inline void PropVariantInit (_Out_ PROPVARIANT * pvar )
{
    memset ( pvar, 0, sizeof(PROPVARIANT) );
}
#   else
#   define PropVariantInit(pvar) memset ( (pvar), 0, sizeof(PROPVARIANT) )
#   endif
#endif /* _MSC_EXTENSIONS */
```

dunhor (approved):

PR #343, opened 2023-06-29
Sync sources with OS

### LINKED ISSUE #340: Build Issues with registry changes
Here's a list of things I've noticed. Note that this might not be exhaustive as I may have missed some errors and the number of errors is capped at 100.

* Use of undefined types `uint8_t`, etc. -> does not include dependencies (in this case `stdint.h`)
* Unprotected use of STL -> *should* be able to use `wistd` alternatives. From what I've seen, this includes:
  * `std::move`
  * `std::enable_if_t`
  * `std::is_same_v`
* Removal of the `wil::max_registry_key_name_length` constant, but I don't see a replacement added

dunhor:
Waiting for the last few build flavors to indicate success.

Fixes #340

### DIFF
--- a/include/wil/registry_helpers.h
+++ b/include/wil/registry_helpers.h
@@ -16,6 +16,7 @@
 #include <iterator>
 #endif
 
+#include <stdint.h>
 #include <Windows.h>
 #include "resource.h"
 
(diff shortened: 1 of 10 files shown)


include/wil/registry_helpers.h
```diff
@@ -182,7 +183,7 @@ namespace wil
                 if (last - first < 3)
                 {
                     // it doesn't have the required 2 terminating null characters - return an empty string
-                    return { ::std::wstring{} };
+                    return ::std::vector<::std::wstring>(1);
```
dunhor, line 186: NOTE: I kept the behavior here as-is, but it kind of seems odd to return a vector with a single empty element. Should this not just be an empty vector?

dunhor: *Side note: this change is specifically because some projects in the OS use an OLD STL version that does not have the `std::initializer_list` constructor for `std::vector`*

Issue #216, opened 2021-10-15
[Kernel] Some wil/resource.h features require Windows 10


jeanga:
When I compile my driver with wil/resource.h, I get error messages:
```
  ...\vcpkg_installed\x64-windows-static\include\wil\resource.h(6490,15): error C2039: 'ExReleasePushLockExclusive': is not a member of '`global namespace'' [F:\Projects\IRM\src\IRM\IRM.vcxproj]
  ...\vcpkg_installed\x64-windows-static\include\wil\resource.h(6490,41): error C3861: 'ExReleasePushLockExclusive': identifier not found [F:\Projects\IRM\src\IRM\IRM.vcxproj]
  ...\vcpkg_installed\x64-windows-static\include\wil\resource.h(6497,15): error C2039: 'ExReleasePushLockShared': is not a member of '`global namespace'' [F:\Projects\IRM\src\IRM\IRM.vcxproj]
  ...\vcpkg_installed\x64-windows-static\include\wil\resource.h(6497,38): error C3861: 'ExReleasePushLockShared': identifier not found [F:\Projects\IRM\src\IRM\IRM.vcxproj]
  ...\vcpkg_installed\x64-windows-static\include\wil\resource.h(6514,11): error C2039: 'ExAcquirePushLockExclusive': is not a member of '`global namespace'' [F:\Projects\IRM\src\IRM\IRM.vcxproj]
  ...\vcpkg_installed\x64-windows-static\include\wil\resource.h(6514,37): error C3861: 'ExAcquirePushLockExclusive': identifier not found [F:\Projects\IRM\src\IRM\IRM.vcxproj]
  ...\IRM\vcpkg_installed\x64-windows-static\include\wil\resource.h(6524,11): error C2039: 'ExAcquirePushLockShared': is not a member of '`global namespace'' [F:\Projects\IRM\src\IRM\IRM.vcxproj]
```

My driver is compiled with the following (relevant) options: /D _WIN32_WINNT=0x0601 /D WINVER=0x0601 /D WINNT=1 /D NTDDI_VERSION=0x06010000 which means that I target Windows 7 and those APIs (ExReleasePushLockExclusive, ExReleasePushLockShared) are Windows 10 only.

It is my impression that these code portions should have been guarded to prevent these APIs' use when targeting Windows 7.

Thank you for WIL!

PR #344, opened 2023-07-01
Fixed including push_lock on windows versions below 1809

ROCKFAL1:
Issue #216

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -6561,7 +6561,7 @@ namespace wil
     }
 
 //! WDM version of EX_PUSH_LOCK is available starting with Windows 10 1809
-#if NTDDI_VERSION >= NTDDI_WIN10_RS5
+#if (NTDDI_VERSION >= NTDDI_WIN10_RS5)
     namespace details
     {
         _IRQL_requires_max_(APC_LEVEL)
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -6560,6 +6560,8 @@ namespace wil
         return{};
     }
 
+//! WDM version of EX_PUSH_LOCK is available starting with Windows 10 1809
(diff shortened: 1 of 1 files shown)


include/wil/resource.h
```diff
@@ -6560,6 +6560,8 @@ namespace wil
         return{};
     }
 
+//! WDM version of EX_PUSH_LOCK is available starting with Windows 10 1809
+#if NTDDI_VERSION >= NTDDI_WIN10_RS5
```
jonwis, line 6564: Thanks - nitpicking, please add parentheses to follow the pattern elsewhere in the repo:

```c
#if (NTDDI_VERSION >= NTDDI_WIN10_RS5)
```

jonwis (changes requested):

include/wil/resource.h
ROCKFAL1: Fixed

ROCKFAL1: I used the information from the MSDN documentation. But it seems that it is not quite correct: the `Ex*PushLock` API, apparently appeared earlier than indicated [here](https://github.com/MicrosoftDocs/windows-driver-docs-ddi/blob/staging/wdk-ddi-src/content/wdm/nf-wdm-exinitializepushlock.md).
Based on the contents of the wdm.h header file, I came to the conclusion that this API appeared starting with Windows 8.1
```cpp
#ifndef EX_NO_PUSH_LOCKS
#if (NTDDI_VERSION < NTDDI_WINBLUE)
#ifndef EX_LEGACY_PUSH_LOCKS
#define EX_LEGACY_PUSH_LOCKS
#endif
#endif // (NTDDI_VERSION < NTDDI_WINBLUE)
#endif // EX_NO_PUSH_LOCKS
```

jonwis (approved):

Issue #338, opened 2023-06-20
wil::reg functions need an enumeration function to walk values and keys


keith-horton:
Feature Request: need to add functions to enumerate values and keys which return an object that supports iterator semantics for ranged-for loops and algorithms, as well as supporting ranges.

e.g., walking dword values under a registry key:

for (const auto& value_name : wil::reg::enumerate_values(key)
{
    auto value = wil::reg::get_value_dword(key, value_name);
}

ChrisGuzak:
and 
```cpp
for (auto&& [key, value] : wil::reg::enumerate_names_and_values(key)
{

}
```

ChrisGuzak:
here was [my first](https://dev.azure.com/chrisguzak/Projects/_search?action=contents&text=registry_key_range&type=code&lp=code-Project&filters=ProjectFilters%7BProjects%7DRepositoryFilters%7BChrisGuzakWork%7D&pageSize=25&result=DefaultCollection/Projects/ChrisGuzakWork/GBmaster//Laboratory/RegistryHelpersProposal.h) run at this. I'm sure it needs improvements.

keith-horton:
We had a tested implementation earlier - it was removed as folks requested the initial commit be much simpler.
[wil_iterator.txt](https://github.com/microsoft/wil/files/11889773/wil_iterator.txt)

PR #345, opened 2023-07-10
User/khorton/registry enumeration

keith-horton:
Initial commit for registry enumeration

### DIFF
--- a/include/wil/registry.h
+++ b/include/wil/registry.h
@@ -161,23 +161,12 @@ namespace wil
             return regview.create_key(subKey, hkey.put(), access);
         }
 #endif // #define __WIL_WINREG_STL
-
-#if defined(WIL_ENABLE_EXCEPTIONS)
         //
-        // wil::reg::value_enumerator<HKEY> enumerate_values(HKEY key)
-        // wil::reg::key_enumerator<HKEY> enumerate_keys(HKEY key)
-        //
-        //  - Enumerates registry keys and values under the given HKEY
-        //  - Throws a std::exception on failure (including wil::ResultException)
+        //  wil::key_iterator and wil::value_iterator objects enable enumerating registry keys and values.
         //
         // Examples of usage when std::wstring is included:
         //
         //     for (const auto& key_data : wil::make_range(wil::reg::key_iterator{hkey}, wil::reg::key_iterator{}))
-        //     {
-        //         key_data.name; // the std::wstring of the enumerated key
-        //     }
-        //
-        //     for (const auto& key_data : wil::reg::enumerate_keys(hkey))
         //     {
         //         key_data.name; // the std::wstring of the enumerated key
         //     }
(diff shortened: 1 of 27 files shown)


tests/RegistryTests.cpp
```diff
@@ -2971,3 +2971,244 @@ TEST_CASE("BasicRegistryTests::cotaskmem_array-bytes", "[registry]]")
...
+        REQUIRE(value_iterator == uniquevalue_enum.end());
+    }
+
+    SECTION("enumerate_values with one value - std::for_each usage")
+    {
+        wil::unique_hkey hkey{wil::reg::create_unique_key(HKEY_CURRENT_USER, testSubkey, wil::reg::key_access::readwrite)};
+        wil::reg::set_value(hkey.get(), L"first_value", 0);
+
+        uint32_t count = 0;
+        auto enumerator = wil::reg::enumerate_values(hkey.get());
```
ChrisGuzak, line 3028: consider if `wil::make_range` should be followed for this name. e.g. `wil::reg::make_value_range`, `wil::reg::make_key_range`.

I'm not sure how much thought went into "make range" so not sure how important that is to follow as precedent.

tests/RegistryTests.cpp
```diff
@@ -2971,3 +2971,244 @@ TEST_CASE("BasicRegistryTests::cotaskmem_array-bytes", "[registry]]")
...
+        std::for_each(enumerator.begin(), enumerator.end(), [&](PCWSTR key_name) {
```
ChrisGuzak, line 3029: `value_name` or just `name` as this is a "value" case. here and below

ChrisGuzak (approved):
check my `make_range` comment. might require a wildisc email to conclude on that.

tests/RegistryTests.cpp
keith-horton: Interesting. I didn't know wil had a make_range. Will next look at integrating that with this enumerator

tests/RegistryTests.cpp
keith-horton: yep

tests/RegistryTests.cpp
dunhor: IMO we should follow the `std::filesystem::directory_iterator` pattern here where the iterator _itself_ exposes `begin()` and `end()` functions. E.g.

```c++
for (auto&& entry : std::directory_iterator(path)) { /* ... */ }
```

And then of course, a default-constructed iterator can be used as the `end` iterator

```c++
std::for_each(std::filesystem::directory_iterator(path), std::filesystem::directory_iterator, callback);
```

keith-horton: I can rename this to be called "iterator" .. but to be honest, I don't understand a model where someone creates a "iterator" who *creates* other iterator types by exposing begin() and end(). That's completely confusing and goes against any iterator model I've seen in the STL.

Today it looks like this:
    for (const auto& value_data : wil::reg::enumerate_values(hkey))

but I can rename it to look like the filesystem name:
    for (const auto& value_data : wil::reg::value_iterator(hkey))

(same for wil::reg::key_iterator).

Duncan, Chris - thoughts?

keith-horton: Or back to Chris's point: avoid 'iterator' nomenclature completely and embrace 'range'.

I'm perfectly happy to just call it:
    for (const auto& value_data : wil::reg::make_value_range(hkey))

(though that doesn't really follow "make_range" semantics)

or something like:

    for (const auto& value_data : wil::reg::make_value_range(reg_view_details::reg_view{ hkey })
    
... which can be moved out of the reg_view_details namespace of course.

dunhor: > That's completely confusing and goes against any iterator model I've seen in the STL.

The STL _is the one that set this precedent_ (the example I provide is _from_ the STL). Of course, not _all_ iterators in the STL behave this way as it wouldn't really make sense for container iterator types such as `std::vector<T>::iterator` or adapter iterator types such as `std::back_inserter<T>` to behave this way, and some iterator types that _could_ behave that way seem like they haven't been updated, such as `std::regex_iterator<T>` which predates the `std::filesystem` APIs.

> avoid 'iterator' nomenclature completely and embrace 'range'.

Ranges are built on top of iterators. In fact, the _definition_ of the concept `std::ranges::range` is that iterators can be made from the type. However, if we were to _truly_ embrace ranges, we'd call it a `view`, not a `range` (so something like `wil::reg::keys_view` and maybe `wil::reg::views::keys` - or if we really want to get crazy - `wil::reg::ranges::keys_view` and `wil::reg::ranges::views::keys`).

One benefit that using an iterator type here has is that it's easier to invoke "legacy" functions that expect iterators. E.g. compare the following two examples:

```c++
do_something(wil::reg::keys_iterator(path), wil::reg::keys_iterator());
```

vs

```c++
auto range = wil::reg::keys_view(path);
do_something(range.begin(), range.end());
```

Using an iterator type saves the local from needing to be created.

That said, I don't personally have too strong of opinions. `wil::reg::keys_iterator`, `wil::reg::make_keys_range`, and `wil::reg::keys_view` all sound good enough to me.

keith-horton: Oh gosh, sorry - I must have replied entirely too late to see that of course, your example was from std::filesystem.

I'll look at their directory_iterator interface and usage.

keith-horton: OK - I'm pushing changes that I think strongly aligns with this model. Thoughts?
this us of wil::reg::key_iterator always works perfectly with STL algorithms. (I have unit tests for std::for_each and std::count)

     for (const auto& key_data : wil::make_range(wil::reg::key_iterator{hkey}, wil::reg::key_iterator{}))
     {
         key_data.name; // the std::wstring of the enumerated key
     }

     for (const auto& key_data : wil::reg::enumerate_keys(hkey))
     {
         key_data.name; // the std::wstring of the enumerated key
     }

tests/RegistryTests.cpp
keith-horton: I think I fixed all instances.

include/wil/registry_helpers.h
```diff
@@ -1218,6 +1218,531 @@ namespace wil
...
+                return const_cast<PWSTR>(name.data());
+            }
+#endif // #if defined(_STRING_)
+        };
+
+        // forward declaration to allow friend-ing the template iterator class
+        template <typename T> class iterator_t;
+
+        template <typename T =
+#if defined(_STRING_)
```
dunhor, line 1321: We should avoid doing stuff like this as much as possible. Simply changing include order, adding a new include to an unrelated header, etc. can change the default, which may cause existing code to break.

include/wil/registry.h
```diff
@@ -162,6 +162,108 @@ namespace wil
...
+        //
+        // When std::wstring is *not* included, the above functions return a 'name' field of type wil::unique_process_heap_string
+        //
+        /**
+         * \brief Returns an enumerator object that exposes begin() and end() to iterate through the registry values under the specified key
+         * \param key An open registry key
+         *        Note: the caller must guarantee the registry key must be valid for the lifetime of the returned registry enumerator object
+         * \return wil::reg::value_enumerator The object to iterate values - exposing the iterator semantics of being() and end()
+         */
+        inline ::wil::reg::value_enumerator<HKEY> enumerate_values(HKEY key)
```
dunhor, line 205: When I see a function named `enumerate*` I expect it to take a callback as a function argument, not return an iterator. Using a name like `make_` or `create_` would alleviate such confusion

include/wil/registry.h
```diff
@@ -162,6 +162,108 @@ namespace wil
         }
 #endif // #define __WIL_WINREG_STL
 
+#if defined(WIL_ENABLE_EXCEPTIONS)
```
dunhor, line 165: If I'm reading this correctly, these iterator types are not available to consumers without exceptions, which seems like an unnecessary limitation. We should follow the design of the `IVector*`, etc. iterator types and make them available in throwing, `_nothrow`, and `_failfast` forms.

include/wil/registry_helpers.h
```diff
@@ -1218,6 +1218,531 @@ namespace wil
...
+                {
+                    memset(name.get(), 0, length * sizeof(wchar_t));
+                }
+            }
+            inline ::wil::unique_process_heap_string copy_name(const ::wil::unique_process_heap_string& str, size_t capacity)
+            {
+                if (!str)
+                {
+                    return {};
+
```
dunhor, line 1259: nit: unnecessary empty line

include/wil/registry_helpers.h
```diff
@@ -1218,6 +1218,531 @@ namespace wil
...
+#endif // #if defined(__WIL_WINREG_STL)
+
+            // function overloads to allow *_iterator_data objects to hold different types to contain the string value
+            // if more string values are desired, all 5 functions must be implement for that type:
+            // clear_name, copy_name, compare_name, resize_name, address_of_name
+            inline void clear_name(::wil::unique_process_heap_string& name, size_t length)
+            {
+                if (length > 0)
+                {
+                    memset(name.get(), 0, length * sizeof(wchar_t));
```
dunhor, line 1251: Why not just `*name = 0`?

include/wil/registry_helpers.h
```diff
@@ -1218,6 +1218,531 @@ namespace wil
...
+            {
+                if (new_length > current_length)
+                {
+                    auto newString{ ::wil::make_unique_string_nothrow<::wil::unique_process_heap_string>(name.get(), new_length) };
+                    name.swap(newString);
+                    return new_length;
+                }
+
+                // continue to use the existing buffer since the requested length is less than or equals to the current length
+                clear_name(name, current_length);
```
dunhor, line 1281: I'm confused. Why the call to `clear_name` here when the above (allocation code path) copies the string with a new length?

include/wil/registry_helpers.h
```diff
@@ -1218,6 +1218,531 @@ namespace wil
...
+            inline size_t resize_name(::std::wstring& name, size_t, size_t new_length)
+            {
+                name.resize(new_length);
+                clear_name(name, name.size());
+                return name.size();
+            }
+            inline PWSTR address_of_name(::std::wstring& name) WI_NOEXCEPT
+            {
+                // cannot return a const PCWSTR -- the registry APIs require a PWSTR
+                return const_cast<PWSTR>(name.data());
```
dunhor, line 1312: `std::basic_string::data` already returns a non-`const` pointer. The `const_cast` is unnecessary

tests/RegistryTests.cpp
DrusTheAxe: I saw this and first thought was why wil:: and not some std type but wistd:: namespace

Psychlist1972:
Any idea when this is planned to make it into the NuGet package for wil?

ChrisGuzak:
> Any idea when this is planned to make it into the NuGet package for wil?

being discussed now. we want to flush the PR queue first. hopefully soon.

dunhor:
This should now be available in [Microsoft.Windows.ImplementationLibrary 1.0.231028.1](https://www.nuget.org/packages/Microsoft.Windows.ImplementationLibrary/1.0.231028.1)

Psychlist1972:
> This should now be available in [Microsoft.Windows.ImplementationLibrary 1.0.231028.1](https://www.nuget.org/packages/Microsoft.Windows.ImplementationLibrary/1.0.231028.1)

Excellent! Thank you.

Do you know if there are any additional steps or latency to get this into the internal build system? I will be able to build with VS from our public repo, but want to verify it'll build under razzle when the repo is sync'd

dunhor:
If you are working in the OS codebase, the code is sync'd with the rs_we_adept_e4d2 branch and has been there since this past Friday if you would like to track its availability.

Psychlist1972:
> If you are working in the OS codebase, the code is sync'd with the rs_we_adept_e4d2 branch and has been there since this past Friday if you would like to track its availability.

Great. Thanks :)

DefaultRyan:
This new helper looks useful, but I don't see any documentation on how to use it in the wiki.

Also, does this PR close out #338 ?

keith-horton:
> This new helper looks useful, but I don't see any documentation on how to use it in the wiki.
> 
> Also, does this PR close out #338 ?

Sorry - it looks like we haven't updated https://github.com/microsoft/wil/wiki/Registry-Helpers :( :(

I wrote canonical examples in the comments. e.g., https://github.com/microsoft/wil/blob/5cbb58909f87d255ced18fa43ea72430d91f2f4f/include/wil/registry.h#L165C12-L165C12

(hopefully that's a direct link - it's to line 165 in https://github.com/Microsoft/wil/blob/master/include/wil/registry.h

tests/RegistryTests.cpp
ChrisGuzak: Now documented [here](https://github.com/microsoft/wil/wiki/Registry-Helpers#enumeration) on the wiki

PR #351, opened 2023-08-15
Re-duplicate tracelogging macros to make the MSVC preprocessor happy

dunhor:
The MSVC preprocessor apparently won't expand macro arguments that contain commas, meaning that the `_TLGWRITE_GENERIC_PARTB_FIELDS` solution that we have currently causes issues because it combines with the previous argument throughout the entire preprocessor chain. This change re-duplicates the macros that use it to avoid this issue. Note that the same issue likely still exists if `_GENERIC_PARTB_FIELDS_ENABLED` contains multiple comma-separated values, _however_ this doesn't appear to be an issue in practice, likely because the macros don't care and the argument gets forwarded all the way until it gets read by the lexer, which will treat commas correctly.

### DIFF
--- a/include/wil/Tracelogging.h
+++ b/include/wil/Tracelogging.h
@@ -1408,238 +1408,479 @@ namespace wil
     };
 
 #ifdef _GENERIC_PARTB_FIELDS_ENABLED
-#define _TLGWRITE_GENERIC_PARTB_FIELDS  , _GENERIC_PARTB_FIELDS_ENABLED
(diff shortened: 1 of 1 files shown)


dunhor:
Oh, also expanded the tests to actually validate that things compile...

Also, `/experimental:preprocessor` gives the correct MSVC behavior, however we cannot rely on that.

dunhor:
For those interested, this illustrates the issue with the MSVC preprocessor: https://godbolt.org/z/js6he3h1Y

jonwis (approved):

PR #355, opened 2023-09-02
Fix typos

m417z:


include/wil/common.h
```diff
@@ -457,13 +457,13 @@ WI_HEADER_INITITALIZATION_FUNCTION(InitializeDesktopFamilyApis, []
 The above example is used within WIL to decide whether or not the library containing WIL is allowed to use
 desktop APIs.  Building this functionality as #IFDEFs within functions would create ODR violations, whereas
 doing it with global function pointers and header initialization allows a runtime determination. */
-#define WI_HEADER_INITITALIZATION_FUNCTION(name, fn)
```
jonwis, line 460: Unfortunately we have 117+ instances of the misspelling in the Windows tree, so we'll need to global search-and-replace when this merges back in, @dunhor

jonwis (approved):
Looks fine, see comment about re-ingestion to Windows.

include/wil/common.h
dunhor: Yeah, we probably need to keep misspellings in actual names (functions, macros, etc.). We can define both and have the old misspelled one "forward" to the new correct spelling

dunhor: Argh, I can't make a suggestion outside of this `#if` chain because GitHub doesn't let me add comments to lines unchanged... This can be solved by the following:
```c++
#define WI_HEADER_INITITALIZATION_FUNCTION(name, fn) WI_HEADER_INITIALIZATION_FUNCTION(name, fn)
```

ALTERNATIVELY if this isn't likely used all that much in the wild, we can throw this definition in the "internal" header so that folks in the OS repo don't get broken

asklar (approved):

include/wil/common.h
m417z: @dunhor rebased and added the old spelling per your suggestion.

dunhor (approved):

PR #358, opened 2023-09-09
Remove redundant string initialization

m417z:


ChrisGuzak (approved):

dunhor (approved):

m417z:
@dunhor these initializations, e.g. `wchar_t message[2048];` to `wchar_t message[2048]{};`, were added in https://github.com/microsoft/wil/pull/343. But I noticed that some places were unaffected by the change for some reason, perhaps the transformation didn't process templates or some ifdef regions.

Examples of such places:
https://github.com/search?q=repo%3Amicrosoft%2Fwil%20%22%5B0%5D%20%3D%20L%27%5C0%27%22&type=code

dunhor:
@m417z if you see a change that says something like "sync with OS" those changes were (typically) originally authored by someone else. I lack the context for the original change - whether it was something "caught" by manual inspection or some overzealous code analysis tool, etc. Personally, I'm not too fond of initializing the entire buffer to zero when it's unnecessary, so I'll take the "missing" locations as a win.

PR #359, opened 2023-09-15
wil::to_array_view provides safer access to IBuffer / IMemoryBuffer

oldnewthing:
The existing `.data()` member gives you a pointer to the start of the buffer but you are on your own to get the buffer size by other means.

The new `wil::to_array_view` function returns the buffer bytes in the form of a `winrt::array_view`, which encodes both the start of the buffer as well as its length.

By default, you get an `array_view<uint8_t>`, but you can customize the data type by calling `to_array_view<T>` to get an `array_view<T>`. Any partial `T`s are ignored. For example, if the buffer is 6 bytes long, then calling `to_array_view<int32_t>` will give you a view of size 1.

To get the `IMemoryBuffer` and `IMemoryBufferReference` versions, you must include `winrt/Windows.Foundation.h` before `wil/cppwinrt_helpers.h`, and you must include `MemoryBuffer.h` at some point before using `to_array_view`.

To get the `IBuffer` version, you must include `winrt/Windows.Storage.Streams.h` before `wil/cppwinrt_helpers.h`.

The `IBuffer` version defaults to returning an `array_view` for the buffer's length. If you want an `array_view` for the buffer's capacity, you can use `to_array_view_for_capacity`.

We follow the pattern of allowing `cppwinrt_helpers.h` to be included multiple times, and each time you do it, new features light up based on what headers you have included prior to including `cppwinrt_helpers.h`.

Fixes microsoft/cppwinrt#1353

### DIFF
--- a/include/wil/cppwinrt_helpers.h
+++ b/include/wil/cppwinrt_helpers.h
@@ -212,6 +212,48 @@ namespace wil::details
 #endif // __WIL_CPPWINRT_MICROSOFT_UI_DISPATCHING_HELPERS
 /// @endcond
 
+#if defined(WINRT_Windows_Foundation_H) && !defined(__WIL_CPPWINRT_WINDOWS_FOUNDATION_HELPERS)
+#define __WIL_CPPWINRT_WINDOWS_FOUNDATION_HELPERS
+namespace Windows::Foundation
+{
+    /// @cond
+    struct IMemoryBufferByteAccess;
+    /// @endcond
+}
+
+namespace wil
+{
+    //! Returns a view into the underlying bytes of a memory buffer
+    //! provided in the form of an IMemoryBufferReference.
+    //! The caller is responsible for ensuring that the memory buffer's
+    //! lifetime encompasses the lifetime of the returned view.
+    //! By default, returns an array_view<uint8_t>, but you can provide an alternate
+    //! type such as to_array_view<double>.
+    //! You must include memorybuffer.h in order to use this overload of to_array_view.
+    template<typename T = uint8_t, int V = 0>
+    winrt::array_view<T> to_array_view(winrt::Windows::Foundation::IMemoryBufferReference const& reference)
+    {
+        uint8_t* data;
+        uint32_t capacity;
+        winrt::check_hresult(reference.as<std::enable_if_t<!V, ::Windows::Foundation::IMemoryBufferByteAccess>>()->GetBuffer(&data, &capacity));
+        return { reinterpret_cast<T*>(data), static_cast<uint32_t>(capacity / sizeof(T)) };
+    }
+
+    //! Returns a view into the underlying bytes of a memory buffer
+    //! provided in the form of an IMemoryBuffer.
+    //! The caller is responsible for ensuring that the memory buffer's
+    //! lifetime encompasses the lifetime of the returned view.
+    //! By default, returns an array_view<uint8_t>, but you can provide an alternate
+    //! type such as to_array_view<double>.
+    //! You must include memorybuffer.h in order to use this overload of to_array_view.
+    template<typename T = uint8_t, int V = 0>
+    winrt::array_view<T> to_array_view(winrt::Windows::Foundation::IMemoryBuffer const& buffer)
+    {
+        return to_array_view<T, V>(buffer.CreateReference());
+    }
+}
+#endif
+
 #if defined(WINRT_Windows_Foundation_Collections_H) && !defined(__WIL_CPPWINRT_WINDOWS_FOUNDATION_COLLECTION_HELPERS)
 #define __WIL_CPPWINRT_WINDOWS_FOUNDATION_COLLECTION_HELPERS
 namespace wil
@@ -310,6 +352,34 @@ namespace wil
         {
             return to_vector(src.First());
         }
+    }
+}
+#endif
+
+#if defined(WINRT_Windows_Storage_Streams_H) && !defined(__WIL_CPPWINRT_WINDOWS_STORAGE_STREAMS_HELPERS)
+#define __WIL_CPPWINRT_WINDOWS_STORAGE_STREAMS_HELPERS
+namespace wil
+{
+    //! Returns a view into the underlying bytes of an IBuffer up to its Length.
+    //! The caller is responsible for ensuring that the IBuffer's
+    //! lifetime encompasses the lifetime of the returned view.
+    //! By default, returns an array_view<uint8_t>, but you can provide an alternate
+    //! type such as to_array_view<double>.
+    template<typename T = uint8_t>
+    winrt::array_view<T> to_array_view(winrt::Windows::Storage::Streams::IBuffer const& buffer)
+    {
+        return { reinterpret_cast<T*>(buffer.data()), static_cast<uint32_t>(buffer.Length() / sizeof(T)) };
+    }
+
+    //! Returns a view into the underlying bytes of an IBuffer up to its Capacity.
+    //! The caller is responsible for ensuring that the IBuffer's
+    //! lifetime encompasses the lifetime of the returned view.
+    //! By default, returns an array_view<uint8_t>, but you can provide an alternate
+    //! type such as to_array_view<double>.
+    template<typename T = uint8_t>
+    winrt::array_view<T> to_array_view_for_capacity(winrt::Windows::Storage::Streams::IBuffer const& buffer)
+    {
+        return { reinterpret_cast<T*>(buffer.data()), static_cast<uint32_t>(buffer.Capacity() / sizeof(T)) };
     }
 }
 #endif
--- a/tests/CppWinRTTests.cpp
+++ b/tests/CppWinRTTests.cpp
@@ -9,7 +9,9 @@
 #include <winrt/Windows.ApplicationModel.Activation.h>
 #include <wil/cppwinrt_helpers.h>
 #include <winrt/Windows.System.h>
+#include <winrt/Windows.Storage.Streams.h>
 #include <wil/cppwinrt_helpers.h> // Verify can include a second time to unlock more features
+#include <memorybuffer.h>
 
 using namespace winrt::Windows::ApplicationModel::Activation;
 
@@ -180,6 +182,42 @@ TEST_CASE("CppWinRTTests::VectorToVector", "[cppwinrt]")
 
     winrt::clear_factory_cache();
     winrt::uninit_apartment();
+}
+
+TEST_CASE("CppWinRTTests::BufferToArrayView", "[cppwinrt]")
+{
+    // Create a buffer of capacity 8 and length 4.
+    auto buffer = winrt::Windows::Storage::Streams::Buffer(8);
+    buffer.Length(4);
+    // Get a Capacity-based int view and set the ints to 256 and 512.
+    {
+        auto view = wil::to_array_view_for_capacity<int32_t>(buffer);
+        REQUIRE(view.size() == 2);
+        view[0] = 256;
+        view[1] = 512;
+    }
+    // Get a Length-based byte view and confirm that the four bytes are { 0, 1, 0, 0 }.
+    // (Assumes little-endian system.)
+    {
+        auto view = wil::to_array_view(buffer);
+        REQUIRE(view.size() == 4);
+        REQUIRE(view == winrt::array_view(std::array<uint8_t, 4>{ 0, 1, 0, 0 }));
+    }
+    // Create an IMemoryBuffer around the Buffer.
+    auto mbuffer = winrt::Windows::Storage::Streams::Buffer::CreateMemoryBufferOverIBuffer(buffer);
+    // Verify that the buffer is the 2 ints 256 and 512.
+    {
+        auto view = wil::to_array_view<int>(mbuffer);
+        REQUIRE(view.size() == 2);
+        REQUIRE(view == winrt::array_view(std::array{ 256, 512 }));
+    }
+    // Verify that the buffer reference is the 8 bytes { 0, 1, 0, 0, 0, 2, 0, 0 }.
+    // (Assumes little-endian system.)
+    {
+        auto view = wil::to_array_view(mbuffer.CreateReference());
+        REQUIRE(view.size() == 8);
+        REQUIRE(view == winrt::array_view(std::array<uint8_t, 8>{ 0, 1, 0, 0, 0, 2, 0, 0 }));
+    }
 }
 
 TEST_CASE("CppWinRTTests::WilToCppWinRTExceptionTranslationTest", "[cppwinrt]")


include/wil/cppwinrt_helpers.h
```diff
@@ -212,6 +212,48 @@ namespace wil::details
...
+
+namespace wil
+{
+    //! Returns a view into the underlying bytes of a memory buffer
+    //! provided in the form of an IMemoryBufferReference.
+    //! The caller is responsible for ensuring that the memory buffer's
+    //! lifetime encompasses the lifetime of the returned view.
+    //! By default, returns an array_view<uint8_t>, but you can provide an alternate
+    //! type such as to_array_view<double>.
+    //! You must include memorybuffer.h in order to use this overload of to_array_view.
```
oldnewthing, line 232: Unfortunately, we have to wait for the client to `#include <memorybuffer.h>` to get access to `IMemoryBufferByteAccess`. C++/WinRT avoids this requirement by manually declaring the interface. Should we do that here?

include/wil/cppwinrt_helpers.h
```diff
@@ -212,6 +212,48 @@ namespace wil::details
...
+    template<typename T = uint8_t, int V = 0>
+    winrt::array_view<T> to_array_view(winrt::Windows::Foundation::IMemoryBufferReference const& reference)
+    {
+        uint8_t* data;
+        uint32_t capacity;
+        winrt::check_hresult(reference.as<std::enable_if_t<!V, ::Windows::Foundation::IMemoryBufferByteAccess>>()->GetBuffer(&data, &capacity));
```
dmachaj, line 238: This is WIL so shouldn't this be `THROW_IF_FAILED`?

tests/CppWinRTTests.cpp
```diff
@@ -182,6 +184,42 @@ TEST_CASE("CppWinRTTests::VectorToVector", "[cppwinrt]")
...
+TEST_CASE("CppWinRTTests::BufferToArrayView", "[cppwinrt]")
+{
+    // Create a buffer of capacity 8 and length 4.
+    auto buffer = winrt::Windows::Storage::Streams::Buffer(8);
+    buffer.Length(4);
+    // Get a Capacity-based int view and set the ints to 256 and 512.
+    {
+        auto view = wil::to_array_view_for_capacity<int32_t>(buffer);
+        REQUIRE(view.size() == 2);
+        view[0] = 256;
```
dmachaj, line 196: For bit-twiddling I think hex is more understandable than decimal.

dunhor:
One thing to possibly consider: the language _really_ wants you to use `std::byte` for "here's a pointer to memory occupied by _data_. Personally... I'm not all that sold, so I don't really care, but maybe something to consider.

include/wil/cppwinrt_helpers.h
dunhor: Any concern to us just including the necessary header? Or can we add its header to the ifdef?

include/wil/cppwinrt_helpers.h
```diff
@@ -212,6 +212,48 @@ namespace wil::details
 #endif // __WIL_CPPWINRT_MICROSOFT_UI_DISPATCHING_HELPERS
 /// @endcond
 
+#if defined(WINRT_Windows_Foundation_H) && !defined(__WIL_CPPWINRT_WINDOWS_FOUNDATION_HELPERS)
+#define __WIL_CPPWINRT_WINDOWS_FOUNDATION_HELPERS
+namespace Windows::Foundation
```
dunhor, line 217: Do we need to do the `ABI` namespace mumbo gumbo here?

include/wil/cppwinrt_helpers.h
dunhor: IMO producing C++/WinRT errors in C++/WinRT specific code is probably nicer

include/wil/cppwinrt_helpers.h
```diff
@@ -212,6 +212,48 @@ namespace wil::details
...
+namespace wil
+{
+    //! Returns a view into the underlying bytes of a memory buffer
+    //! provided in the form of an IMemoryBufferReference.
+    //! The caller is responsible for ensuring that the memory buffer's
+    //! lifetime encompasses the lifetime of the returned view.
+    //! By default, returns an array_view<uint8_t>, but you can provide an alternate
+    //! type such as to_array_view<double>.
+    //! You must include memorybuffer.h in order to use this overload of to_array_view.
+    template<typename T = uint8_t, int V = 0>
```
dunhor, line 233: What's the `V` here for? Just to prevent compilation errors when `IMemoryBufferByteAccess` is not defined?

include/wil/cppwinrt_helpers.h
dunhor: Answering my own question. This isn't a WinRT/MIDLRT generated header, so that shouldn't be a concern unless someone includes it within a `namespace ABI` guard

dunhor:
We seem to be sporadically hitting a lot of `STATUS_ACCESS_VIOLATION` lately... Re-ran the pipeline

dunhor (approved):

include/wil/cppwinrt_helpers.h
oldnewthing: Unfortunately, `memorybuffer.h` relies on `#pragma once` and does not have a `#define` for detecting whether it has been included.

I'm okay with explicitly including the header if you think it's sufficiently unobtrusive.

include/wil/cppwinrt_helpers.h
oldnewthing: Yes, the V is to prevent the template from being instantiated before IMemoryBufferByteAccess is defined.

dunhor: Yeah, was about to ask if the `is_same_v<T, T>` trick didn't delay evaluation for some reason

include/wil/cppwinrt_helpers.h
dunhor: Hmm, looks like the only thing it includes is `objidl.h`. While I'd like to believe that it is innocuous, I'm not so sure of that. I'd be okay defining it ourselves _but_ wouldn't we have the same issue with `IUnknown` as well?

dunhor: Ultimately, I think this is probably fine. The alternative would be to move the definition to a new header that does the necessary inclusion that the user would then need to explicitly include, so not _that_ much different than needing to include an additional header. We may want to consider adding a comment above the line where the compilation error will be produced saying something like "if you receive an error saying that this type is not defined, include this header."

include/wil/cppwinrt_helpers.h
dmachaj: That's fair

jonwis (approved):

tests/CppWinRTTests.cpp
oldnewthing: Got out of the bit twiddling business entirely.

include/wil/cppwinrt_helpers.h
oldnewthing: I'll require the caller to include for now. We can relax that later as a non-breaking change.

sylveon:
Suggestion: add an `assert(buffer.Length() % sizeof(T) == 0)`

oldnewthing:
> Suggestion: add an `assert(buffer.Length() % sizeof(T) == 0)`

I don't think so. You don't always control who gave you the buffer, and you don't want to crash if somebody gave you a wrong-sized buffer. We assume you just want to ignore the fractional bytes. This can happen if, for example, somebody creates a buffer based on GlobalSize(), which is allowed to return a value larger than the value passed to GlobalAlloc.

dmachaj (approved):

dunhor:
For some reason, your changes are really giving the pipeline troubles. I think every build/re-build has failed with an AV in the same place and I just re-ran master which got passed that point. Can you make sure to build & run the tests with Clang targeting x86 w/ RelWithDebInfo to see if it repros locally for you?

JaiganeshKumaran:
I don't really like having basic helpers on WinRT types in WIL - this should be in C++/WinRT.

sylveon:
C++/WinRT is nearly frozen at this point, so it makes sense to focus C++/WinRT itself on the projection only and keep helpers in a repo that can evolve faster.

I wouldn't be surprised if an eventual C++/WinRT 3 extracts most built-in helpers to WIL.

oldnewthing:
@dunhor

> Can you make sure to build & run the tests with Clang targeting x86 w/ RelWithDebInfo to see if it repros locally for you?

* `scripts\init -c clang -g ninja -b debug`
* `cd build\clang32debug`
* `ninja` (let build complete)
* `cd ..\..`
* `scripts\runtests`

> Running tests from D:\repos\wil\build\clang32debug
> Running app tests...
> \===============================================================================
> All tests passed (87175 assertions in 115 test cases)
> 
> Running cpplatest tests...
> \===============================================================================
> All tests passed (108794 assertions in 222 test cases)
> 
> Running noexcept tests...
> \===============================================================================
> All tests passed (40415 assertions in 157 test cases)
> 
> Running normal tests...
> \===============================================================================
> All tests passed (107300 assertions in 199 test cases)
> 
> Running win7 tests...
> \===============================================================================
> All tests passed (93425 assertions in 161 test cases)

dunhor:
That's Debug, not RelWithDebInfo :). I checked out your repo and looks like this repros for me pretty reliably:
```
0:000> k
 # ChildEBP RetAddr      
00 012fd648 001fd933     KERNELBASE!RaiseException+0x64 [minkernel\kernelbase\xcpt.c @ 955] 
01 012fd694 000c1c5a     witest_cpplatest!witest::FakeFailfastWithContext+0x13 [tests\common.h @ 212] 
02 012fec0c 001fd8e9     witest_cpplatest!wil::details::LogFailure+0x1aa [include\wil\result_macros.h @ 3569] 
03 (Inline) --------     witest_cpplatest!wil::details::in1diag3::FailFast_IfFailed+0x40 [include\wil\result_macros.h @ 5014] 
04 (Inline) --------     witest_cpplatest!wil::err_failfast_policy::HResult+0x40 [include\wil\result_macros.h @ 6191] 
05 (Inline) --------     witest_cpplatest!wil::com_ptr_t<INever,wil::err_failfast_policy>::com_ptr_t+0x5e [include\wil\com.h @ 874] 
06 (Inline) --------     witest_cpplatest!wil::com_query_failfast+0x6a [include\wil\com.h @ 1259] 
07 (Inline) --------     witest_cpplatest!TestGlobalQuery<ComObject,ComObject *>::<lambda_5>::operator()+0x6c [tests\ComTests.cpp @ 1206] 
08 012fec44 001fd83b     witest_cpplatest!witest::DoesCodeThrow<`lambda at tests\ComTests.cpp:1206:13' &>+0x99 [tests\common.h @ 196] 
09 (Inline) --------     witest_cpplatest!witest::ReportsError<`lambda at tests\ComTests.cpp:1206:13'>::<lambda_1>::operator()+0x8 [tests\common.h @ 269] 
0a (Inline) --------     witest_cpplatest!wistd::__invoke+0x8 [include\wil\wistd_type_traits.h @ 4012] 
0b (Inline) --------     witest_cpplatest!wistd::__invoke_void_return_wrapper<void>::__call+0x8 [include\wil\wistd_functional.h @ 100] 
0c 012fec4c 001fd9a3     witest_cpplatest!wistd::__function::__func<`lambda at tests/common.h:267:40',void ()>::operator()+0xb [include\wil\wistd_functional.h @ 250] 
0d 012fec7c 001fc6fe     witest_cpplatest!witest::details::DoesCodeCrash+0x43 [tests\common.h @ 240] 
0e (Inline) --------     witest_cpplatest!witest::DoesCodeCrash+0x41 [tests\common.h @ 260] 
0f 012fed08 001d4e03     witest_cpplatest!witest::ReportsError<`lambda at tests\ComTests.cpp:1206:13'>+0x8e [tests\common.h @ 267] 
10 012ff5a8 001d4788     witest_cpplatest!TestGlobalQuery<ComObject,ComObject *>+0x4d3 [tests\ComTests.cpp @ 1206] 
11 012ff5d4 001d4745     witest_cpplatest!TestQueryCombination<ComObject,ComObject>+0x38 [tests\ComTests.cpp @ 1652] 
12 012ff5d8 001d46ef     witest_cpplatest!TestQuery<ComObject,ComObject>+0x5 [tests\ComTests.cpp @ 1667] 
13 012ff5e0 000f6f43     witest_cpplatest!TestQuery<ComObject,ComObject,ComObject>+0xf [tests\ComTests.cpp @ 1675] 
14 012ff648 0006cbfd     witest_cpplatest!C_A_T_C_H_T_E_S_T_102+0x43 [tests\ComTests.cpp @ 1685] 
15 012ff670 0006afc4     witest_cpplatest!Catch::RunContext::invokeActiveTestCase+0x4d [tests\catch.hpp @ 13026] 
16 012ff7c8 0006a72d     witest_cpplatest!Catch::RunContext::runCurrentTest+0x174 [tests\catch.hpp @ 13000] 
17 012ff914 0006f8d2     witest_cpplatest!Catch::RunContext::runTest+0x16d [tests\catch.hpp @ 12760] 
18 (Inline) --------     witest_cpplatest!Catch::`anonymous namespace'::TestGroup::execute+0x143 [tests\catch.hpp @ 13352] 
19 012ffac4 0006f456     witest_cpplatest!Catch::Session::runInternal+0x432 [tests\catch.hpp @ 13562] 
1a 012ffad0 00089785     witest_cpplatest!Catch::Session::run+0x36 [tests\catch.hpp @ 13518] 
1b (Inline) --------     witest_cpplatest!Catch::Session::run+0x2f [tests\catch.hpp @ 13236] 
1c 012ffc1c 009f2021     witest_cpplatest!main+0x85 [tests\catch.hpp @ 17543]
```
I'll start looking into it to see what's triggering the issue; IIRC issues like this are generally due to Clang handling SEH a bit weird

oldnewthing:
@dunhor Oops, sorry, missed that detail. I cannot repro with clang32relwithdebinfo either.

> Running tests from D:\repos\wil\build\clang32relwithdebinfo
> Running app tests...
> \===============================================================================
> All tests passed (87175 assertions in 115 test cases)
>
> Running cpplatest tests...
> \===============================================================================
> All tests passed (109193 assertions in 222 test cases)

From the stack, it appears to be an order-of-uninitialization issue when an app fails to uninitialize COM, and COM is forced to uninitialize itself from inside `DLL_PROCESS_DETACH`.

tests/CppWinRTTests.cpp
```diff
@@ -182,6 +184,44 @@ TEST_CASE("CppWinRTTests::VectorToVector", "[cppwinrt]")
     winrt::uninit_apartment();
 }
 
+TEST_CASE("CppWinRTTests::BufferToArrayView", "[cppwinrt]")
+{
+    std::array<int32_t, 2> testData = { 314159265, 27182818 };
```
dunhor, line 189:
```suggestion
    winrt::init_apartment();
    auto uninit = wil::scope_exit([]{ winrt::uninit_apartment(); });

    std::array<int32_t, 2> testData = { 314159265, 27182818 };
```

This seems to avoid the crash. Note that I still hit the crash locally for other tests, however AFAIK we haven't hit the same in the build labs

dunhor: OMG! This comment exists for the CppWinRTAuthoringTests and I didn't even know about it!:
```c++
// This test cannot run in the same process as the malloc spies tests in wiTest.cpp
// MSFT_internal: https://task.ms/44191550
```
Making that `[LocalOnly]` isn't really the correct move, though...

ChrisGuzak (approved):

tests/CppWinRTTests.cpp
oldnewthing: @dunhor Can we proceed with this PR? We sort of gave up on the NotifyPropertyChanged test (PRs 368, 374).

dunhor: I've disabled the NotifyPropertyChanged test to run during CI. We can either do the same thing here or we can update the test to call `CoInitialize` to avoid the crash during shutdown.

dunhor: Err, actually disabling for just CI is probably not the best move. If we go the disable route, we should disable even when run locally so that the scripts don't break.

jonwis:
Once the build finished again I'll poke the merge button. Thanks for your patience @oldnewthing.

dunhor:
> Once the build finished again I'll poke the merge button. Thanks for your patience @oldnewthing.

@jonwis see the discussion above. There's a crash in the tests caused by a bug in COM that's caused by a combination of using `IMallocSpy` and not cleaning up C++/WinRT's `CoIncrementMTAUsage`. Raymond and I have chatted about this a bit offline (as it turns out, this is not the first test to fail because of it), and I've floated a few ideas:

1. Simplest thing is to manually call `CoInitialize` as I've suggested above. This bypasses C++/WinRT's logic & we don't hit the issue because we've properly cleaned up COM before process exit
2. Another simple thing to do is to disable the test, but obviously not ideal
3. Another idea I've floated is to tag the tests that use `IMallocSpy` specially and then perform two passes of test invocation: one that excludes the `IMallocSpy`-using tests and one that only runs those tests

I'm personally a fan of option 1: it's simple and effective. Option 3 would help us avoid these issues moving forward, however.

jonwis:
Filed #393 to address the discovered issue

include/wil/cppwinrt_helpers.h
```diff
@@ -226,6 +226,51 @@ struct dispatcher_traits<winrt::Microsoft::UI::Dispatching::DispatcherQueue>
 #endif // __WIL_CPPWINRT_MICROSOFT_UI_DISPATCHING_HELPERS
 /// @endcond
 
+#if defined(WINRT_Windows_Foundation_H) && !defined(__WIL_CPPWINRT_WINDOWS_FOUNDATION_BUFFER_HELPERS) ||\
+    defined(WIL_DOXYGEN)
+/// @cond
+#define __WIL_CPPWINRT_WINDOWS_FOUNDATION_BUFFER_HELPERS
+namespace Windows::Foundation
+{
+    struct IMemoryBufferByteAccess;
```
dunhor, line 235:
```suggestion
struct IMemoryBufferByteAccess;
```
To make clang-format happy

dunhor: Can I on a fork?

One thing I'd really like to do if I ever got the time to do it is to add a pipeline/action/something that I can run that will run clang format & commit the result, either as a separate command (e.g. `/azp format`) or just part of the pipeline by default.

sylveon:
This is very late but, any reason for not preferring std::span?

PR #365, opened 2023-09-26
Implementing a simple registry watcher

keith-horton:
Implementing a simple registry watcher, which maintains the same contract as the unique_threadpool objects it uses for the callbacks - it guarantees all callbacks have completed and no more will be invoked when the d'tor has completed

### DIFF
--- a/include/wil/registry.h
+++ b/include/wil/registry.h
@@ -3109,12 +3109,117 @@ namespace wil
             }
         };
 
+        constexpr DWORD registry_notify_filter = REG_NOTIFY_CHANGE_LAST_SET | REG_NOTIFY_CHANGE_NAME | REG_NOTIFY_THREAD_AGNOSTIC;
+
         inline void delete_registry_watcher_state(_In_opt_ registry_watcher_state* watcherStorage) { watcherStorage->Release(); }
 
         typedef resource_policy<registry_watcher_state*, decltype(&details::delete_registry_watcher_state),
             details::delete_registry_watcher_state, details::pointer_access_none> registry_watcher_state_resource_policy;
     }
     /// @endcond
+
+    template <typename err_policy = ::wil::err_exception_policy>
+    class slim_registry_watcher_t
+    {
+    public:
+        // HRESULT or void error handling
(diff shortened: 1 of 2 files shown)


include/wil/registry.h
```diff
@@ -3109,13 +3109,118 @@ namespace wil
...
+    class slim_registry_watcher_t
+    {
+    public:
+        // HRESULT or void error handling
+        typedef typename err_policy::result result;
+
+        slim_registry_watcher_t() WI_NOEXCEPT = default;
+
+        // Exception-based constructors
+        slim_registry_watcher_t(HKEY rootKey, _In_ PCWSTR subKey, bool isRecursive, ::wistd::function<void(::wil::RegistryChangeKind)>&& callback)
```
dmachaj, line 3131: Perhaps it would be safer if there were a series of constructors that took various kinds of smart pointers (weak and strong).  The registry watcher can then keep the callback alive (if given a strong pointer) or allow it to safely destruct (if given a weak pointer).

keith-horton: I'm at a loss to see what benefit differing behaviors provides based off the type of callback one provides (well, what type of container holds their callback).

It seems some folks really wanted to decouple the object enabling the callbacks from their callback code that runs. That doesn't make sense to me - if I want to unregister, I don't want more callbacks. But I'm only 1 person, so cool - if some folks want to employ weak-ref models with their callback designs, and they don't want to worry about properly synchronizing when objects are destroyed -- then they have an option.

But I think that should be a strong interface contract at the class level, not implicit behavior on what type of callback container they provide.

Hope this makes sense (?)

dmachaj: I don't think I expressed my suggestion well.  What I'm trying to suggest is that `slim_registry_watcher` is a tuple containing everything you already have AND some sort of lifetime management for the thing that is being called back.  That could be a strong reference, a weak reference, or for non-class-based code it can be left off.

Providing a strong or weak pointer should completely close the race with object destruction.  A strong reference prevents destruction.  A weak reference allows the callback to know that it is already destructed so it can NOOP.

PR #367, opened 2023-09-26
Disable 'NotifyPropertyChanged' since it crashes when run locally

dunhor:
The test currently is marked with `[LocalOnly]` so that it doesn't run on CI machines, however this breaks scripts even when running locally (and is annoying with a post-mortem debugger). Disabling it entirely for now


dunhor:
Abandoning for now since (1) it appears as though running with a filter prevents the "hiding" of tests, and (2) @oldnewthing is investigating a separate fix

PR #368, opened 2023-09-26
Fix malloc spy conflict with CppWinRTAuthoringTests::NotifyPropertyChanged

oldnewthing:
The CppWinRTAuthoringTests::NotifyPropertyChanged test did not clean up everything before it exited. This resulted in COM remaining initialized at process termination, and that in turn resulted in adverse interactions with the malloc spy test, since the "emergency cleanup" calls CoTaskMemFree, and since COM was put into "possible malloc spy" mode, it tries to take a lock to protect the potential malloc spy. But that lock no longer exists, since COM has shut down, and things spiral downhill.

The NotifyPropertyChanged test tries to clean up the WindowsXamlManager, but it didn't realize that Close() is an asynchronous call that queues work back to the current thread's DispatcherQueue for final cleanup. The old test just returned immediately and never pumped messages any more, which resulted in the cleanup never happening.

To ensure we clean up the WindowsXamlManager, we create our own DispatcherQueue for the XAML work. Since it's our DispatcherQueue, we can call ShutdownQueueAsync to clean it up. ShutdownQueueAsync completes when the queue has shut down, but before the thread uninitializes COM. We wait for the DispatcherQueue thread to exit, so that we know for sure that it has definitely uninitialized COM. Only then do we allow the test to complete.

This cleans up the leaked COM initialization and avoids a condition crash during "emergency cleanup", thereby resolving a longstanding problem with the automated unit tests.

The diff looks big because of a change in indentation. Turn on "ignore whitespace" to see the actual change.

### DIFF
--- a/tests/CppWinRTAuthoringTests.cpp
+++ b/tests/CppWinRTAuthoringTests.cpp
@@ -212,7 +212,7 @@ TEST_CASE("CppWinRTAuthoringTests::EventsAndCppWinRt", "[property]")
 TEST_CASE("CppWinRTAuthoringTests::NotifyPropertyChanged", "[property][LocalOnly]")
 {
 #if defined(WIL_ENABLE_EXCEPTIONS)
-    auto uninit = wil::RoInitialize_failfast(RO_INIT_SINGLETHREADED);
+    auto uninit = wil::RoInitialize_failfast(RO_INIT_MULTITHREADED);
     // We need to initialize the XAML core in order to instantiate a PropertyChangedEventArgs.
     // Do all the work on a separate DispatcherQueue thread so we can shut it down cleanly
     // and make sure COM is completely cleaned up before we finish the test.
--- a/tests/CppWinRTAuthoringTests.cpp
+++ b/tests/CppWinRTAuthoringTests.cpp
@@ -209,7 +209,7 @@ TEST_CASE("CppWinRTAuthoringTests::EventsAndCppWinRt", "[property]")
 #include <winrt/Windows.System.h>
 #include <winrt/Windows.UI.Xaml.Hosting.h>
 
-TEST_CASE("CppWinRTAuthoringTests::NotifyPropertyChanged", "[property][LocalOnly]")
+TEST_CASE("CppWinRTAuthoringTests::NotifyPropertyChanged", "[property]")
 {
 #if defined(WIL_ENABLE_EXCEPTIONS)
     auto uninit = wil::RoInitialize_failfast(RO_INIT_MULTITHREADED);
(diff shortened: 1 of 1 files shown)


tests/CppWinRTAuthoringTests.cpp
```diff
@@ -206,57 +206,70 @@ TEST_CASE("CppWinRTAuthoringTests::EventsAndCppWinRt", "[property]")
...
-    auto manager = winrt::Windows::UI::Xaml::Hosting::WindowsXamlManager::InitializeForCurrentThread();
-    {
-        struct Test : winrt::implements<Test, winrt::Windows::UI::Xaml::Data::INotifyPropertyChanged>, wil::notify_property_changed_base<Test>
+    // Do all the work on a separate DispatcherQueue thread so we can shut it down cleanly
+    // and make sure COM is completely cleaned up before we finish the test.
+
+    auto controller = winrt::Windows::System::DispatcherQueueController::CreateOnDedicatedThread();
+    winrt::handle dispatcherThreadHandle;
+
+    controller.DispatcherQueue().TryEnqueue([&]
```
dmachaj, line 223: >TryEnqueue [](http://example.com/codeflow?start=33&length=10)

Do you need to `REQUIRE` that `TryEnqueue` succeeds?

tests/CppWinRTAuthoringTests.cpp
```diff
@@ -206,57 +206,70 @@ TEST_CASE("CppWinRTAuthoringTests::EventsAndCppWinRt", "[property]")
...
+                        notified = true;
+                    });
+
+                testImpl->MyProperty(43);
+                REQUIRE(notified);
+                test.PropertyChanged(token);
+            }
+            manager.Close();
+        });
+    controller.ShutdownQueueAsync().get();
```
dmachaj, line 269: >get [](http://example.com/codeflow?start=36&length=3)

Is this guaranteed to complete in a reasonable amount of time?  get is a potentially infinite wait.

dunhor (approved):

tests/CppWinRTAuthoringTests.cpp
```diff
@@ -206,57 +206,70 @@ TEST_CASE("CppWinRTAuthoringTests::EventsAndCppWinRt", "[property]")
 #endif // WINRT_Windows_Foundation_H
 
 #if defined(WINRT_Windows_UI_Xaml_Data_H)
+#include <winrt/Windows.System.h>
 #include <winrt/Windows.UI.Xaml.Hosting.h>
 
-// This test cannot run in the same process as the malloc spies tests in wiTest.cpp
-// MSFT_internal: https://task.ms/44191550
 TEST_CASE("CppWinRTAuthoringTests::NotifyPropertyChanged", "[property][LocalOnly]")
```
dunhor, line 212:
```suggestion
TEST_CASE("CppWinRTAuthoringTests::NotifyPropertyChanged", "[property]")
```

The `[LocalOnly]` tag disables for CI. Presumably this can be enabled now

asklar (approved):
thanks @oldnewthing!

tests/CppWinRTAuthoringTests.cpp
oldnewthing: I'll fold it with the WFSO.

tests/CppWinRTAuthoringTests.cpp
oldnewthing: Sure.

dmachaj (approved):

dunhor (approved):

PR #373, opened 2023-09-28
fix #347:warning C6001

### LINKED ISSUE #347: warning C6001
packages\Microsoft.Windows.ImplementationLibrary.1.0.230629.1\include\wil\resource.h(5599): warning C6001: 使用未初始化的内存“*hClientHandle”。

kouzhudong:
https://github.com/microsoft/wil/issues/347

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -5597,7 +5597,7 @@ namespace wil
     /// @cond
     namespace details
     {
-        inline void __stdcall CloseWlanHandle(_Frees_ptr_ HANDLE hClientHandle)
+        inline void __stdcall CloseWlanHandle(_In_ HANDLE hClientHandle)
         {
             ::WlanCloseHandle(hClientHandle, nullptr);
         }


oldnewthing:
@dunhor

>   REQUIRE( WaitForSingleObject(dispatcherThreadHandle.get(), 5000) == ((((DWORD )0x00000000L) ) + 0 ) )
> with expansion:
>   258 (0x102) == 0

This means that the XAML test failed to complete within 5 seconds. Which is strange since we see no test failures from it. Maybe the system is CPU-overloaded?

I'm going to rewrite the last-minute fix suggested by @dmachaj about "What if ShutdownAsync hangs?" Before that last-minute fix, we gave the thread 5 seconds to call CoUninitialize, but with the fix, the entire test has 5 seconds to run. I'm going to go back to giving just the CoUninitialize a 5-second timeout. Or maybe just wait infinitely and let Catch2's default timeout detect the hang.

dunhor:
Sounds good. I'll hold off on disabling the test until we know for sure if the test is stable or not

PR #382, opened 2023-10-24
add wil::create_file_create, create_file_open, etc. tame use of CreateFileW

ChrisGuzak:
I distilled the documented behavior of [CreateFileW](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew) into these helpers with the goal enabling **simple use** of the functionality without having to construct the long and complicated, and interdependent parameter list. 

```cpp
auto handle = wil::open_file(path.c_str());
```
The design uses "open", "create" and "trucate" to distinguish the cases that open existing files and use that creates and modifies existing files. Here is more complicated use that demonstrates the handle and error code being returned (used in **success cases**, so providing access to it is essential.). 

```cpp
auto path = wil::ExpandEnvironmentStringsW<std::wstring>(LR"(%TEMP%\create_file_test.txt)");
auto [fileHandle, error] = wil::try_create_new_file(path.c_str())
```
Here, are the method names that intend to describe the semantics.
```cpp
auto handle = wil::create_new_file(path.c_str());
auto handle = wil::open_or_create_file(path.c_str());
auto handle = wil::open_or_truncate_existing_file(path.c_str());
auto handle = wil::truncate_existing_file(path.c_str());

```
PR includes tests now.

### DIFF
--- a/include/wil/filesystem.h
+++ b/include/wil/filesystem.h
@@ -1054,7 +1054,7 @@ namespace wil
     // Helpers to make the CreateFileW API easier to use. This segregates the OPEN_EXISTING "open an existing file"
     // cases from the "create a new file" that has 4 variations represented in the create_file_create_behavior enum.
 
-    // https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew?devlangs=cpp&f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(FILEAPI%252FCreateFileW)%3Bk(CreateFileW)%3Bk(DevLang-C%252B%252B)%3Bk(TargetOS-Windows)%26rd%3Dtrue
+    // https://learn.microsoft.com/windows/win32/api/fileapi/nf-fileapi-createfilew
 
     struct file_and_error_result
     {
--- a/include/wil/filesystem.h
+++ b/include/wil/filesystem.h
@@ -19,6 +19,7 @@
 #include <combaseapi.h> // Needed for CoTaskMemFree() used in output of some helpers.
 #include <winbase.h> // LocalAlloc
 #include <PathCch.h>
+#include "wistd_type_traits.h"
 #include "result.h"
 #include "win32_helpers.h"
 #include "resource.h"
@@ -716,10 +717,10 @@ namespace wil
             {
                 for (auto const& info : create_next_entry_offset_iterator(reinterpret_cast<FILE_NOTIFY_INFORMATION *>(readerState->m_readBuffer)))
                 {
-                    wchar_t realtiveFileName[MAX_PATH];
-                    StringCchCopyNW(realtiveFileName, ARRAYSIZE(realtiveFileName), info.FileName, info.FileNameLength / sizeof(info.FileName[0]));
+                    wchar_t relativeFileName[MAX_PATH];
+                    StringCchCopyNW(relativeFileName, ARRAYSIZE(relativeFileName), info.FileName, info.FileNameLength / sizeof(info.FileName[0]));
 
-                    readerState->m_callback(static_cast<FolderChangeEvent>(info.Action), realtiveFileName);
+                    readerState->m_callback(static_cast<FolderChangeEvent>(info.Action), relativeFileName);
                 }
             }
             else if (result == ERROR_NOTIFY_ENUM_DIR)
@@ -1047,8 +1048,6 @@ namespace wil
         THROW_IF_FAILED(GetFileInfoNoThrow<infoClass>(fileHandle, result));
         return result;
     }
-#endif // _CPPUNWIND
-#endif // WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP) && (_WIN32_WINNT >= _WIN32_WINNT_WIN7)
 
 // Helpers to make the CreateFileW API easier to use. This segregates the OPEN_EXISTING "open an existing file"
 // cases from the "create a new file" that has 4 variations represented in the create_file_create_behavior enum.
@@ -1118,15 +1117,15 @@ auto [handle, error] = wil::create_file_try_create(filePath.c_str());
 ~~~
 */
 template<create_file_create_behavior CreateDisposition = create_file_create_behavior::overwrite_existing>
-std::pair<wil::unique_hfile, DWORD> create_file_try_create(PCWSTR path, DWORD dwDesiredAccess = FILE_READ_ACCESS | FILE_WRITE_ACCESS,
+wistd::pair<wil::unique_hfile, DWORD> create_file_try_create(PCWSTR path, DWORD dwDesiredAccess = FILE_READ_ACCESS | FILE_WRITE_ACCESS,
     DWORD dwShareMode = FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
     LPSECURITY_ATTRIBUTES lpSecurityAttributes = nullptr,
     DWORD dwFlagsAndAttributes = FILE_ATTRIBUTE_NORMAL,
     HANDLE hTemplateFile = nullptr) noexcept
 {
-    return std::pair<wil::unique_hfile, DWORD>{CreateFileW(
+    return wistd::pair<wil::unique_hfile, DWORD>{CreateFileW(
         path, dwDesiredAccess, dwShareMode, lpSecurityAttributes, static_cast<DWORD>(CreateDisposition), dwFlagsAndAttributes, hTemplateFile),
-        GetLastError()};
+        ::GetLastError()};
 }
 
 /** create, throws on error.
@@ -1146,13 +1145,15 @@ wil::unique_hfile create_file_create(PCWSTR path, DWORD dwDesiredAccess = FILE_R
     DWORD dwFlagsAndAttributes = FILE_ATTRIBUTE_NORMAL,
     HANDLE hTemplateFile = nullptr)
 {
-    auto [fileHandle, error] = std::pair<wil::unique_hfile, DWORD>{CreateFileW(
+    auto [fileHandle, error] = wistd::pair<wil::unique_hfile, DWORD>{CreateFileW(
         path, dwDesiredAccess, dwShareMode, lpSecurityAttributes, static_cast<DWORD>(CreateDisposition), dwFlagsAndAttributes, hTemplateFile),
-        GetLastError()};
+        ::GetLastError()};
     THROW_WIN32_IF(error, !fileHandle.is_valid());
     return std::move(fileHandle);
 }
 
+#endif // _CPPUNWIND
+#endif // WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP) && (_WIN32_WINNT >= _WIN32_WINNT_WIN7)
 }
 
 #ifndef NO_FILE_TYPE_OPERATORS
--- a/include/wil/filesystem.h
+++ b/include/wil/filesystem.h
@@ -198,6 +198,7 @@ namespace wil
     };
     DEFINE_ENUM_FLAG_OPERATORS(RemoveDirectoryOptions);
 
+    /// @cond
     namespace details
     {
         // Reparse points should not be traversed in most recursive walks of the file system,
@@ -209,6 +210,7 @@ namespace wil
                     (IsReparseTagDirectory(info.ReparseTag) || (info.ReparseTag == IO_REPARSE_TAG_WCI))));
         }
     }
+    /// @endcond
 
     // Retrieve a handle to a directory only if it is safe to recurse into.
     inline wil::unique_hfile TryCreateFileCanRecurseIntoDirectory(PCWSTR path, PWIN32_FIND_DATAW fileFindData, DWORD access = GENERIC_READ | /*DELETE*/ 0x00010000L, DWORD share = FILE_SHARE_READ)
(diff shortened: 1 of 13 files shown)


include/wil/filesystem.h
```diff
@@ -1045,6 +1045,125 @@ namespace wil
     }
 #endif // _CPPUNWIND
 #endif // WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP) && (_WIN32_WINNT >= _WIN32_WINNT_WIN7)
+
+// Helpers to make the CreateFileW API easier to use. This segregates the OPEN_EXISTING "open an existing file"
+// cases from the "create a new file" that has 4 variations represented in the create_file_create_behavior enum.
+
+// https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew?devlangs=cpp&f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(FILEAPI%252FCreateFileW)%3Bk(CreateFileW)%3Bk(DevLang-C%252B%252B)%3Bk(TargetOS-Windows)%26rd%3Dtrue
```
jonwis, line 1052: Consider removing the `en-us/` part and cleaning up the stuff after the `?=` maybe.

include/wil/filesystem.h
```diff
@@ -1045,6 +1045,125 @@ namespace wil
...
+~~~
+*/
+template<create_file_create_behavior CreateDisposition = create_file_create_behavior::overwrite_existing>
+std::pair<wil::unique_hfile, DWORD> create_file_try_create(PCWSTR path, DWORD dwDesiredAccess = FILE_READ_ACCESS | FILE_WRITE_ACCESS,
+    DWORD dwShareMode = FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
+    LPSECURITY_ATTRIBUTES lpSecurityAttributes = nullptr,
+    DWORD dwFlagsAndAttributes = FILE_ATTRIBUTE_NORMAL,
+    HANDLE hTemplateFile = nullptr) noexcept
+{
+    return std::pair<wil::unique_hfile, DWORD>{CreateFileW(
```
jonwis, line 1123: Is this execution order (createfile first, getlasterror after) guaranteed?

include/wil/filesystem.h
```diff
@@ -1045,6 +1045,125 @@ namespace wil
...
+    return handleResult;
+}
+
+/** create, non-throwing, returns handle and error code as a pair.
+~~~
+auto [handle, error] = wil::create_file_try_create(filePath.c_str());
+~~~
+*/
+template<create_file_create_behavior CreateDisposition = create_file_create_behavior::overwrite_existing>
+std::pair<wil::unique_hfile, DWORD> create_file_try_create(PCWSTR path, DWORD dwDesiredAccess = FILE_READ_ACCESS | FILE_WRITE_ACCESS,
```
jonwis, line 1117: Should we also fix the camel-case on the parameter names here?

include/wil/filesystem.h
```diff
@@ -1045,6 +1045,125 @@ namespace wil
...
+auto handle = wil::create_file_create(filePath.c_str());
+~~~
+
+There are 4 create behaviors, overwrite_existing is default. To specify the others use the template parameter.
+~~~
+auto [fileHandle, error] = create_file_try_create<create_file_create_behavior::create_only_if_new>(overWriteTarget.c_str());
+~~~
+*/
+template<create_file_create_behavior CreateDisposition = create_file_create_behavior::overwrite_existing>
+wil::unique_hfile create_file_create(PCWSTR path, DWORD dwDesiredAccess = FILE_READ_ACCESS | FILE_WRITE_ACCESS,
```
jonwis, line 1139: Using template specialization for this purpose seems strange, compared to specifically named wrapper functions.

include/wil/filesystem.h
```diff
@@ -1045,6 +1045,125 @@ namespace wil
...
+}
+
+}
+
+#ifndef NO_FILE_TYPE_OPERATORS
+// namespace scope
+inline bool operator!=(const FILE_ID_128& left, const FILE_ID_128& right)
+{
+    static_assert(sizeof(left) == sizeof(right), "size must match");
+    return memcmp(&left, &right, sizeof(left)) != 0;
```
jonwis, line 1159: Nit - write this in terms of the other?

jonwis (commented):
I'm of multiple minds on this. I like that it's clearing up the "floor wax and dessert topping" form of the API, I don't know that this is the way that's most intuitive in naming for me yet.

include/wil/filesystem.h
dmachaj, line 1117: >pair [](http://example.com/codeflow?start=5&length=4)

Don't you need to `#idef` that `std::pair` is already included?

include/wil/filesystem.h
```diff
@@ -1045,6 +1045,125 @@ namespace wil
...
 }
 
+/** create, throws on error.
+~~~
+auto handle = wil::create_file_create(filePath.c_str());
+~~~
+
+There are 4 create behaviors, overwrite_existing is default. To specify the others use the template parameter.
+~~~
+auto [fileHandle, error] = create_file_try_create<create_file_create_behavior::create_only_if_new>(overWriteTarget.c_str());
```
dmachaj, line 1135: >create_file_try_create [](http://example.com/codeflow?start=27&length=22)

Comment is copy/pasted from previous without being renamed.

include/wil/filesystem.h
```diff
@@ -1045,6 +1045,125 @@ namespace wil
...
+~~~
+*/
+template<create_file_create_behavior CreateDisposition = create_file_create_behavior::overwrite_existing>
+wil::unique_hfile create_file_create(PCWSTR path, DWORD dwDesiredAccess = FILE_READ_ACCESS | FILE_WRITE_ACCESS,
+    DWORD dwShareMode = FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE,
+    LPSECURITY_ATTRIBUTES lpSecurityAttributes = nullptr,
+    DWORD dwFlagsAndAttributes = FILE_ATTRIBUTE_NORMAL,
+    HANDLE hTemplateFile = nullptr)
+{
+    auto [fileHandle, error] = std::pair<wil::unique_hfile, DWORD>{CreateFileW(
```
dmachaj, line 1145: >auto [](http://example.com/codeflow?start=4&length=4)

`create_file_open` calls `create_file_try_open` and throws if it fails.  `create_file_create` duplicates the CreateFile call similar to `create_file_try_create` and throws the result.  It seems like these should be consistent (duplicate, or throwing calls non-throwing).

include/wil/filesystem.h
```diff
@@ -1045,6 +1045,125 @@ namespace wil
...
+    return std::move(fileHandle);
+}
+
+}
+
+#ifndef NO_FILE_TYPE_OPERATORS
+// namespace scope
+inline bool operator!=(const FILE_ID_128& left, const FILE_ID_128& right)
+{
+    static_assert(sizeof(left) == sizeof(right), "size must match");
```
dmachaj, line 1158: >sizeof(left) == sizeof(right), "size must match" [](http://example.com/codeflow?start=18&length=48)

How could this possibly fail?  This method is not templated and these types are the same.  Should it be comparing the size of some field within this type?

include/wil/filesystem.h
```diff
@@ -1043,8 +1046,224 @@ namespace wil
         THROW_IF_FAILED(GetFileInfoNoThrow<infoClass>(fileHandle, result));
         return result;
     }
+
+    // Helpers to make the CreateFileW API easier to use. This segregates the OPEN_EXISTING "open an existing file"
+    // cases from the "create a new file" that has 4 variations represented in the create_file_create_behavior enum.
+
+    // https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew?devlangs=cpp&f1url=%3FappId%3DDev16IDEF1%26l%3DEN-US%26k%3Dk(FILEAPI%252FCreateFileW)%3Bk(CreateFileW)%3Bk(DevLang-C%252B%252B)%3Bk(TargetOS-Windows)%26rd%3Dtrue
+
+    struct file_and_error_result
```
dunhor, line 1055: Maybe I'm just weird, but I'd be fine with even the `try_` versions returning a `wil::unique_handle` with the expectation that the user queries it for validity (& then calls `GetLastError` if desired). That would be consistent with how other "try" functions tend to work such as `wil::com_ptr_t::try_query`. The usage is also not _that_ differnet. E.g. consider the two possible use cases:
```c++
// CURRENT
auto [handle, err] = wil::try_open_file(path);
RETURN_IF_WIN32_ERROR(err);

// ALTERNATIVE
auto handle = wil::try_open_file(path);
RETURN_LAST_ERROR_IF(!handle);
```
The usage is pretty similar. It might be slightly more work if someone wants to change behavior based on the specific error, but that's probably not the most common use.

include/wil/filesystem.h
```diff
@@ -1043,8 +1046,224 @@ namespace wil
...
+    */
+    inline file_and_error_result try_open_file(PCWSTR path, DWORD dwDesiredAccess = FILE_READ_ACCESS,
+        DWORD dwShareMode = FILE_SHARE_READ,
+        bool inheritHandle = false,
+        DWORD dwFlagsAndAttributes = FILE_ATTRIBUTE_NORMAL) noexcept
+    {
+        SECURITY_ATTRIBUTES secAttributes{ sizeof(secAttributes) };
+        secAttributes.bInheritHandle = inheritHandle;
+        return { CreateFileW(path, dwDesiredAccess, dwShareMode, &secAttributes, OPEN_EXISTING, dwFlagsAndAttributes, nullptr),
+            ::GetLastError() };
```
dunhor, line 1078: It might be worth conditionally calling `GetLastError` depending on whether or not `CreateFileW` returned an invalid handle (& deduce `ERROR_SUCCESS` if not).

include/wil/filesystem.h
```diff
@@ -1043,8 +1046,224 @@ namespace wil
...
+    auto handle = wil::open_file(filePath.c_str());
+    ~~~
+    */
+    inline wil::unique_hfile open_file(PCWSTR path, DWORD dwDesiredAccess = FILE_READ_ACCESS,
+        DWORD dwShareMode = FILE_SHARE_READ,
+        bool inheritHandle = false,
+        DWORD dwFlagsAndAttributes = FILE_ATTRIBUTE_NORMAL)
+    {
+        auto result = try_open_file(path, dwDesiredAccess, dwShareMode, inheritHandle, dwFlagsAndAttributes);
+        THROW_WIN32_IF(result.last_error, !result.file.is_valid());
```
dunhor, line 1092: Probably needs "if exceptions enabled" pre-processor check

include/wil/filesystem.h
```diff
@@ -1043,8 +1046,224 @@ namespace wil
...
+    /** create using CREATE_NEW, returns handle and error code.
+    ~~~
+    auto [handle, error = wil::try_create_new_file(filePath.c_str());
+    ~~~
+    */
+    inline file_and_error_result try_create_new_file(PCWSTR path,
+        DWORD dwDesiredAccess = FILE_READ_ACCESS | FILE_WRITE_ACCESS,
+        DWORD dwShareMode = FILE_SHARE_READ,
+        LPSECURITY_ATTRIBUTES lpSecurityAttributes = nullptr,
+        DWORD dwFlagsAndAttributes = FILE_ATTRIBUTE_NORMAL,
```
dunhor, line 1123: I realize this is preserving argument order for `CreateFileW`, which might be good for reducing confusion, _but_ we have the opportunity to change it if we wanted to. E.g. I find myself specifying `dwFlagsAndAttributes` far more often than I find myself specifying `lpSecurityAttributes`.

include/wil/filesystem.h
```diff
@@ -1043,8 +1046,224 @@ namespace wil
...
+
+    /** Non-throwing open existing using OPEN_EXISTING.
+    ~~~
+    auto handle = wil::try_open_file(filePath.c_str());
+    ~~~
+    */
+    inline file_and_error_result try_open_file(PCWSTR path, DWORD dwDesiredAccess = FILE_READ_ACCESS,
+        DWORD dwShareMode = FILE_SHARE_READ,
+        bool inheritHandle = false,
+        DWORD dwFlagsAndAttributes = FILE_ATTRIBUTE_NORMAL) noexcept
```
dunhor, line 1073: Personally I'd feel like it's mor common to specify flags (open directories, open reparse points) than it is to inherit handles. Consider swapping parameter order

include/wil/filesystem.h
dmachaj: That's a good point.  Won't the value of GetLastError be undefined on success?  Or is `CreateFileW` guaranteed to clear it to success on entry?

include/wil/filesystem.h
dmachaj: +1

Isn't it the case that if you want to specify an optional parameter then that means you must _also_ specify every optional parameter that comes before that on the argument list?

dunhor: > you must also specify every optional parameter that comes before that on the argument list

Correct

include/wil/filesystem.h
dunhor: From the documentation on [GetLastError](https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror#remarks):

> some functions call SetLastError with a zero when they succeed, wiping out the error code set by the most recently failed function.

Sounds like it's not guaranteed. The only relevant info on the documentation for [CreateFileW](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilew#return-value) seems to be:

> If the function fails, the return value is INVALID_HANDLE_VALUE. To get extended error information, call [GetLastError](https://learn.microsoft.com/en-us/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror).

There's nothing that indicates that the function calls `SetLastError` for success cases

include/wil/filesystem.h
```diff
@@ -1043,8 +1046,224 @@ namespace wil
...
+    {
+        SECURITY_ATTRIBUTES secAttributes{ sizeof(secAttributes) };
+        secAttributes.bInheritHandle = inheritHandle;
+        return { CreateFileW(path, dwDesiredAccess, dwShareMode, &secAttributes, OPEN_EXISTING, dwFlagsAndAttributes, nullptr),
```
dunhor, line 1077: nit: seems like this could also be updated to call `details::create_file`

include/wil/filesystem.h
```diff
@@ -1043,8 +1046,224 @@ namespace wil
...
+            path, dwDesiredAccess, dwShareMode, lpSecurityAttributes, dwFlagsAndAttributes, hTemplateFile);
+        THROW_WIN32_IF(result.last_error, !result.file.is_valid());
+        return std::move(result.file);
+    }
+
 #endif // _CPPUNWIND
 #endif // WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP) && (_WIN32_WINNT >= _WIN32_WINNT_WIN7)
 }
 
+#ifndef NO_FILE_TYPE_OPERATORS
```
dunhor, line 1254: nit: consider prefixing names like this with something like `WIL_`

include/wil/filesystem.h
ChrisGuzak: I want to provide the last error in success cases too, so that motivated me to use the struct. if you still disagree let me know and I can reconsider.

include/wil/filesystem.h
ChrisGuzak: the docs are clear that the last error is meaningful in success cases and that ensures it always is.

include/wil/filesystem.h
ChrisGuzak: I like the order that we were provided.
read/write mode and sharing options are the most commonly specified.
FILE_FLAG_BACKUPSEMANTICs is needed in the attributes for folders. so that is not as convenient as I'd like. but maybe that means we should add `wil::open_directory()`

include/wil/filesystem.h
ChrisGuzak: agree. Let me try that.

include/wil/filesystem.h
ChrisGuzak: it seems to compile. I'll leave it I guess.

include/wil/filesystem.h
ChrisGuzak: could be, yes. see my update!

include/wil/filesystem.h
dunhor: I still think that _most_ consumers don't need/care about the last set error value on success. Or at least I've never personally called `GetLastError` when I've gotten back a valid handle. That said, I was mostly just bringing it up for discussion; I'm okay with it either way.

include/wil/filesystem.h
dunhor: Ahh, I missed a few mentions of when error values are set:
> If the specified file exists and is writable, the function truncates the file, the function succeeds, and last-error code is set to ERROR_ALREADY_EXISTS

I'm still of the opinion that most of the time this information is not needed, but it makes more sense now. Alternatively, and this might be over-designing things, we could only return the last set error for the functions where it makes sense (`CREATE_ALWAYS` and `OPEN_ALWAYS`). We could also possibly "translate" this into something like a `bool already_exists` instead of an error code.

include/wil/filesystem.h
dunhor: Another data point/thing to consider: the exception versions don't communicate the error code

include/wil/filesystem.h
dunhor: Oh, that's because you defined these functions inside of `#ifdef _CPPUNWIND` (starts on line 1027). Seems like we'd probably want the `try_` versions to be available without exceptions, right?

include/wil/filesystem.h
```diff
@@ -1043,8 +1050,219 @@ namespace wil
         THROW_IF_FAILED(GetFileInfoNoThrow<infoClass>(fileHandle, result));
         return result;
     }
+
+    // Helpers to make the CreateFileW API easier to use. This segregates the OPEN_EXISTING "open an existing file"
+    // cases from the "create a new file" that has 4 variations represented in the create_file_create_behavior enum.
```
dunhor, line 1055: Is `create_file_create_behavior` a remnant of a previous version of this change that got removed?

include/wil/filesystem.h
ChrisGuzak: I'm not very sympathetic for projects that build with exceptions disabled. for them, go direct to CreateFileW

include/wil/filesystem.h
```diff
@@ -1047,8 +1050,217 @@ namespace wil
...
+    auto handle = wil::open_file(filePath.c_str());
+    ~~~
+    */
+    inline wil::unique_hfile open_file(PCWSTR path, DWORD dwDesiredAccess = FILE_READ_ACCESS,
+        DWORD dwShareMode = FILE_SHARE_READ, DWORD dwFlagsAndAttributes = FILE_ATTRIBUTE_NORMAL,
+        bool inheritHandle = false) noexcept
+    {
+        auto result = try_open_file(path, dwDesiredAccess, dwShareMode, inheritHandle, dwFlagsAndAttributes);
+        THROW_WIN32_IF(result.last_error, !result.file.is_valid());
+        return std::move(result.file);
```
dmachaj, line 1093: It is dis-recommended to `return std::move(X)` because it defeats some types of optimizations.  The compiler will automatically move the return.

dmachaj (approved):

include/wil/filesystem.h
dunhor: I might be wrong, but I don't think you can get NRVO from a member object.

In any case, I don't think the alternative would compile because `result.file` is not an x-value. E.g. see: https://godbolt.org/z/P8Pr73oen

dmachaj: Oh I missed that this is returning a move from the struct (and not the entire struct).  Nevermind then.

tests/FileSystemTests.cpp
```diff
@@ -650,7 +649,6 @@ TEST_CASE("FileSystemTests::GetFileInfo<FileStreamInfo>", "[filesystem]")
     wistd::unique_ptr<FILE_STREAM_INFO> streamInfo;
     auto hr = wil::GetFileInfoNoThrow<FileStreamInfo>(handle.get(), streamInfo);
     REQUIRE(hr == S_OK);
-#endif
 }
 
 TEST_CASE("FileSystemTests::QueryFullProcessImageNameW", "[filesystem]")
```
dunhor, line 654: nit: Not critical, but this could be pulled out of the `ifdef` for exception support

dunhor (commented):
I'm still of the opinion that flags & attributes should come before security attributes, but otherwise things look good to me

dunhor (approved):

PR #386, opened 2023-11-02
Fix folder_change_reader_state destructor hang. 

razvan-pricope:
If there is no pending operation, CancelIoEx returns an error and GetOverlappedResult hangs.
This is a fix for an issue introduced in #326. If **CancelIoEx** returns `FALSE` && `GetLastError() == ERROR_NOT_FOUND` then there is no pending operation, and `GetOverlappedResult` might hang. The hang can happen if you delete the monitored folder after receiving a completion notification but before calling again `StartIo`

### DIFF
--- a/include/wil/filesystem.h
+++ b/include/wil/filesystem.h
@@ -624,10 +624,15 @@ namespace wil
 
                     if (m_folderHandle)
                     {
-                        CancelIoEx(m_folderHandle.get(), &m_overlapped);
+                       BOOL cancelRes = CancelIoEx(m_folderHandle.get(), &m_overlapped);
 
-                        DWORD bytesTransferredIgnored = 0;
-                        GetOverlappedResult(m_folderHandle.get(), &m_overlapped, &bytesTransferredIgnored, TRUE);
+                       // no pending operation to cancel. Maybe StartIo returned 
+                       // an error?
+                       if (!(cancelRes == FALSE && ::GetLastError() == ERROR_NOT_FOUND)) 
+                       {
+                           DWORD bytesTransferredIgnored = 0;
+                           GetOverlappedResult(m_folderHandle.get(), &m_overlapped, &bytesTransferredIgnored, TRUE);
+                       }
                     }
 
                     // Wait for callbacks to complete.
--- a/include/wil/filesystem.h
+++ b/include/wil/filesystem.h
@@ -624,15 +624,15 @@ namespace wil
 
                     if (m_folderHandle)
                     {
-                       BOOL cancelRes = CancelIoEx(m_folderHandle.get(), &m_overlapped);
+                        BOOL cancelRes = CancelIoEx(m_folderHandle.get(), &m_overlapped);
 
-                       // no pending operation to cancel. Maybe StartIo returned 
-                       // an error?
-                       if (!(cancelRes == FALSE && ::GetLastError() == ERROR_NOT_FOUND)) 
-                       {
-                           DWORD bytesTransferredIgnored = 0;
-                           GetOverlappedResult(m_folderHandle.get(), &m_overlapped, &bytesTransferredIgnored, TRUE);
-                       }
+                        // no pending operation to cancel. Maybe StartIo returned 
+                        // an error?
+                        if (!(cancelRes == FALSE && ::GetLastError() == ERROR_NOT_FOUND)) 
+                        {
+                            DWORD bytesTransferredIgnored = 0;
+                            GetOverlappedResult(m_folderHandle.get(), &m_overlapped, &bytesTransferredIgnored, TRUE);
+                        }
                     }
 
                     // Wait for callbacks to complete.
--- a/tests/FileSystemTests.cpp
+++ b/tests/FileSystemTests.cpp
@@ -773,6 +773,64 @@ TEST_CASE("FileSystemTests::CreateFileW helpers", "[filesystem]")
 }
 
 #endif
+TEST_CASE("FileSystemTest::FolderChangeReader destructor does not hang", "[filesystem]")
+{
+    std::wstring testRootDir;
+    testRootDir = wil::ExpandEnvironmentStringsW<std::wstring>(L"%TEMP%\\wil_test_filesystem");
+    std::wstring testFile = testRootDir + L"\\test.dat";
+    bool deleteDir = false;
+    wil::unique_handle opCompleteEv(::CreateEventW(nullptr, TRUE, FALSE, nullptr));
+    wil::unique_handle wilReaderNotify(::CreateEventW(nullptr, TRUE, FALSE, nullptr));
 
+    REQUIRE_FALSE(DirectoryExists(testRootDir.c_str()));
+    REQUIRE(SUCCEEDED(wil::CreateDirectoryDeepNoThrow(testRootDir.c_str())));
+    REQUIRE(DirectoryExists(testRootDir.c_str()));
+
+    /**
+     * Move to a new thread.
+     * The destructor of unique_folder_change_reader might hang. If this happens, 
+     * we want to report an test error instead of hanging forever. 
+     */
+    auto reader = wil::make_folder_change_reader_nothrow(testRootDir.c_str(), false, wil::FolderChangeEvents::All,
+    [&](wil::FolderChangeEvent, PCWSTR)
+    {
+        if (deleteDir)
+            RemoveDirectoryW(testRootDir.c_str());
+
+        SetEvent(opCompleteEv.get());
+    });
+
+    struct ThreadContext
+    {
+        wil::unique_folder_change_reader_nothrow reader;
+        HANDLE threadStopEv = NULL;
+        static DWORD WINAPI ThreadFn(LPVOID lpParam)
+        {
+            auto ctx = reinterpret_cast<ThreadContext*>(lpParam);
+            WaitForSingleObject(ctx->threadStopEv, INFINITE);
+            delete ctx;
+            return 0;
+        }
+    };
+    auto ctx = new ThreadContext();
+    ctx->reader = std::move(reader);
+    ctx->threadStopEv = wilReaderNotify.get();
+    wil::unique_handle readerThreadH(::CreateThread(nullptr, 0, (LPTHREAD_START_ROUTINE)ThreadContext::ThreadFn,
+        ctx, 0, nullptr));
+
+    wil::unique_hfile testFileOut(::CreateFileW(testFile.c_str(), GENERIC_ALL,
+        FILE_SHARE_READ, nullptr, CREATE_ALWAYS, 0, nullptr));
+    REQUIRE(testFileOut);
+    testFileOut.reset();
+    WaitForSingleObject(opCompleteEv.get(), INFINITE);
+
+    ResetEvent(opCompleteEv.get());
+    deleteDir = true;
+    REQUIRE(DeleteFileW(testFile.c_str()));
+    WaitForSingleObject(opCompleteEv.get(), INFINITE);
+    
+    SetEvent(wilReaderNotify.get());
+    REQUIRE(WaitForSingleObject(readerThreadH.get(), 30 * 1000) == WAIT_OBJECT_0);
+}
 
 #endif // WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
--- a/tests/FileSystemTests.cpp
+++ b/tests/FileSystemTests.cpp
@@ -5,9 +5,8 @@
 #include <wil/common.h>
 #ifdef WIL_ENABLE_EXCEPTIONS
 #include <string>
-#endif
-
 #include <thread>
+#endif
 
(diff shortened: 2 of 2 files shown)


dunhor:
Hmm, for some reason the pipeline is having trouble with ASan. Looks like it's complaining about a large allocation [which I believe should be covered with this](https://github.com/microsoft/wil/blob/7e14e28547603125e3d12f68e9d2ee443851386f/tests/main.cpp#L16C38-L16C38).

I'll have to take a closer look at it on Monday

dunhor (approved):

include/wil/filesystem.h
```diff
@@ -624,10 +624,15 @@ namespace wil
 
                     if (m_folderHandle)
                     {
-                        CancelIoEx(m_folderHandle.get(), &m_overlapped);
-
-                        DWORD bytesTransferredIgnored = 0;
-                        GetOverlappedResult(m_folderHandle.get(), &m_overlapped, &bytesTransferredIgnored, TRUE);
+                       BOOL cancelRes = CancelIoEx(m_folderHandle.get(), &m_overlapped);
```
ChrisGuzak, line 627: nit, the indentation here looks wrong. 4 spaces, no tabs please.

include/wil/filesystem.h
```diff
@@ -624,10 +624,15 @@ namespace wil
...
                     {
-                        CancelIoEx(m_folderHandle.get(), &m_overlapped);
-
-                        DWORD bytesTransferredIgnored = 0;
-                        GetOverlappedResult(m_folderHandle.get(), &m_overlapped, &bytesTransferredIgnored, TRUE);
+                       BOOL cancelRes = CancelIoEx(m_folderHandle.get(), &m_overlapped);
+
+                       // no pending operation to cancel. Maybe StartIo returned 
+                       // an error?
+                       if (!(cancelRes == FALSE && ::GetLastError() == ERROR_NOT_FOUND))
```
ChrisGuzak, line 631: parens around sub-expressions. 
here is a re-write using DeMorgan, but this seems strange to me. is this right?
`if (cancelRes || (::GetLastError() != ERROR_NOT_FOUND))`

ChrisGuzak (commented):
Please add a test that demonstrates the problem and show that this fixes it.

include/wil/filesystem.h
razvan-pricope: fixed. Don't know how I managed to add 3 spaces instead of 4.

include/wil/filesystem.h
razvan-pricope: Yes, it looks odd to me too (hard to parse) 
Also I am not sure when we should call `GetOverlappedResult`:
- If `CancelIoEx` returned `TRUE` -- we should always call it.
- If `CancelIoEx` returned false:
-- do we want to call it when `GetLastError() != ERROR_NOT_FOUND` ? 
-- should we never call it? (maybe there is another error that could cause the call to hang).

razvan-pricope:
> Please add a test that demonstrates the problem and show that this fixes it.

Done.

include/wil/filesystem.h
ChrisGuzak: I don't know either so would need to consult with experts. maybe I can get some of their attention. do you have a repro case we can point to? ideally in the form of a test.

ChrisGuzak: From an expert... (thank you!)

> If I’m reading it right, CancelIoEx can fail because:
Kernelbase passed ntoskrnl bad memory (shouldn’t happen);
The handle is invalid (shouldn’t happen);
The request is not outstanding so couldn’t be cancelled (what this fix is about);
The request is outstanding but was already cancelled by another thread so this thread won’t cancel it.  I think wil is already protecting itself against that by the m_cancelLock above this call; this is the challenging case because if it happened it’s still important to wait for the request to complete, even if this thread didn’t initiate the cancel.  So my 2c would be that this change looks fine, although it shouldn’t be blindly copy-pasted to an environment without that cancel lock.

ChrisGuzak (approved):
my comments are only on the tests, but please clean that up if you can.

tests/FileSystemTests.cpp
```diff
@@ -773,6 +773,64 @@ TEST_CASE("FileSystemTests::CreateFileW helpers", "[filesystem]")
...
+TEST_CASE("FileSystemTest::FolderChangeReader destructor does not hang", "[filesystem]")
+{
+    std::wstring testRootDir;
+    testRootDir = wil::ExpandEnvironmentStringsW<std::wstring>(L"%TEMP%\\wil_test_filesystem");
+    std::wstring testFile = testRootDir + L"\\test.dat";
+    bool deleteDir = false;
+    wil::unique_handle opCompleteEv(::CreateEventW(nullptr, TRUE, FALSE, nullptr));
+    wil::unique_handle wilReaderNotify(::CreateEventW(nullptr, TRUE, FALSE, nullptr));
+
+    REQUIRE_FALSE(DirectoryExists(testRootDir.c_str()));
```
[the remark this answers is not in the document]
razvan-pricope: That's not a macro function, it's a function defined in [FileSystemTests.cpp](https://github.com/microsoft/wil/blob/master/tests/FileSystemTests.cpp#L29)

tests/FileSystemTests.cpp
```diff
@@ -773,6 +773,64 @@ TEST_CASE("FileSystemTests::CreateFileW helpers", "[filesystem]")
...
+    auto reader = wil::make_folder_change_reader_nothrow(testRootDir.c_str(), false, wil::FolderChangeEvents::All,
+    [&](wil::FolderChangeEvent, PCWSTR)
+    {
+        if (deleteDir)
+            RemoveDirectoryW(testRootDir.c_str());
 
+        SetEvent(opCompleteEv.get());
+    });
+
+    struct ThreadContext
```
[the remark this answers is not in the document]
razvan-pricope: I cannot use `std::thread` because `join()` doesn't have any timeout. However, I will rewrite it using `std::future` / `std::async` to be able to `wait_for(...)` the result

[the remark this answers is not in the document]
razvan-pricope: actually i cannot use `std::future` because I cannot detach it in case of a hang. So I rewrote the code using `std::thread` + `native_handle()`

tests/FileSystemTests.cpp
```diff
@@ -773,6 +773,64 @@ TEST_CASE("FileSystemTests::CreateFileW helpers", "[filesystem]")
 }
 
 #endif
+TEST_CASE("FileSystemTest::FolderChangeReader destructor does not hang", "[filesystem]")
+{
+    std::wstring testRootDir;
+    testRootDir = wil::ExpandEnvironmentStringsW<std::wstring>(L"%TEMP%\\wil_test_filesystem");
+    std::wstring testFile = testRootDir + L"\\test.dat";
+    bool deleteDir = false;
+    wil::unique_handle opCompleteEv(::CreateEventW(nullptr, TRUE, FALSE, nullptr));
```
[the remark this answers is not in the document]
razvan-pricope: hmm, seems to be hitting some [build errors](https://dev.azure.com/msft-wil/Windows%20Implementation%20Library/_build/results?buildId=970&view=logs&j=ab477ad6-9781-5b54-82ca-a0141b26726e&t=8fddd505-58bd-5f5a-42fd-95acd1c9abe1&l=226) when building in azure 😕that I don't have when building locally (I am using msvc, not clang). However, If I include `#include <wil/resource.h>` I get an error when building locally....

razvan-pricope:

include/wil/filesystem.h
razvan-pricope: So, how should we go forward with this change?

tests/FileSystemTests.cpp
```diff
@@ -773,6 +776,66 @@ TEST_CASE("FileSystemTests::CreateFileW helpers", "[filesystem]")
...
+    wil::unique_hfile testFileOut(::CreateFileW(testFile.c_str(), GENERIC_ALL,
+        FILE_SHARE_READ, nullptr, CREATE_ALWAYS, 0, nullptr));
+    REQUIRE(testFileOut);
+    testFileOut.reset();
+    WaitForSingleObject(opCompletedEv.get(), INFINITE);
+
+    deleteDir = true;
+    REQUIRE(DeleteFileW(testFile.c_str()));
+    WaitForSingleObject(opCompletedEv.get(), INFINITE);
+    std::this_thread::sleep_for(std::chrono::seconds(5)); //enough time for the StartIO call to fail
```
razvan-pricope, line 829: I am not really happy with this sleep here, but without it, there might be a race between calling the destructor and calling again `StartIo`

tests/FileSystemTests.cpp
```diff
@@ -7,6 +7,8 @@
 #include <string>
 #endif
 
+#include <thread>
```
dunhor, line 10: I'm a little surprised this builds (though maybe it gives the same warnings as other code). This should ideally be moved up inside of the `#ifdef WIL_ENABLE_EXCEPTIONS` above

tests/FileSystemTests.cpp
```diff
@@ -773,6 +776,66 @@ TEST_CASE("FileSystemTests::CreateFileW helpers", "[filesystem]")
...
+
+        SetEvent(opCompletedEv.get());
+    });
+    auto readerThread = std::thread([&wilReaderNotify, r = std::move(reader)]() mutable {
+
+        /**
+         * moving here to make sure the reader is destructed on a different thread 
+         */
+        auto r2 = std::move(r);
+        WaitForSingleObject(wilReaderNotify.get(), INFINITE);
```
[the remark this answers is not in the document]
dunhor: They should also have their own `SetEvent` functions

tests/FileSystemTests.cpp
```diff
@@ -773,6 +776,66 @@ TEST_CASE("FileSystemTests::CreateFileW helpers", "[filesystem]")
...
+     */
+    auto reader = wil::make_folder_change_reader_nothrow(testRootDir.c_str(), false, wil::FolderChangeEvents::All,
+    [&](wil::FolderChangeEvent, PCWSTR)
+    {
+        if (deleteDir)
+            RemoveDirectoryW(testRootDir.c_str());
+
+        SetEvent(opCompletedEv.get());
+    });
+    auto readerThread = std::thread([&wilReaderNotify, r = std::move(reader)]() mutable {
```
dunhor, line 811: Since the test allows waiting for the thread to time out, probably don't want to capture `wilReaderNotify` by reference. IMO - keep the `HANDLE` stashed locally and then `std::move` the `unique_event` into the lambda so that the thread can be the one to destroy the event.

tests/FileSystemTests.cpp
dunhor: How reliable is the repro if you: (1) decrease the timeout (e.g. maybe to something like 100ms), or (2) instead do something like `std::this_thread::yield`. If the issue reproduces (without this fix) a decent percentage of time (let's say >50% but maybe even less than that), IMO that's still decent enough that we'll catch regressions pretty quickly

dunhor: Wild thought: is the `m_tpIo` accessible off the `unique_folder_change_reader_nothrow` type? If so, I'm curious if the thread can call `WaitForThreadpoolIoCallbacks` itself. That would ensure that the callback - in its entirety - has completed, which includes the call to `StartIo`.

dunhor: Looks like the `get` function in the `unique_storage` class is public, so this seems possible

tests/FileSystemTests.cpp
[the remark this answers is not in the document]
razvan-pricope: fixed.

tests/FileSystemTests.cpp
razvan-pricope: fixed

tests/FileSystemTests.cpp
dunhor: > Wild thought: is the `m_tpIo` accessible off the `unique_folder_change_reader_nothrow` type? If so, I'm curious if the thread can call `WaitForThreadpoolIoCallbacks` itself. That would ensure that the callback - in its entirety - has completed, which includes the call to `StartIo`.

Curious if you gave this a try. I also wouldn't be opposed to adding a macro that conditionalizes making members public vs private for unit tests

razvan-pricope: Does not seem to be possible with the current implementation. Calling
>PTP_IO readerRaw = reader.get()->m_tpIo;

gives me

> wil\include\wil\resource.h(347,107): error C2338: static_assert failed: 'get(): the raw handle value is not available for this resource class

dunhor: Ahh, I see now that `folder_change_reader_state_resource_policy` is marked as `pointer_access_none`. In theory we could do something like `#if WIL_TESTING` to make `pointer_access_none` act like `pointer_access_all`, _but_ that also sounds like something that we can address in the future

dunhor (approved):
Thanks for finding, diagnosing, and fixing this!

Issue #384, opened 2023-10-27
Add a string literal marker for wil::zwstring_view


jonwis:
@dmachaj points out that it'd be handy to have a literal that produced a wil::zwstring_view for use in arrays or just generally "easy to make zviews".

PR #392, opened 2023-11-13
Define a string literal marker for zwstring_view

tusharchitti:
Issue : https://github.com/microsoft/wil/issues/384

Added a literal that produced a wil::zwstring_view.

Added a simple test case to validate it.

### DIFF
--- a/include/wil/stl.h
+++ b/include/wil/stl.h
@@ -187,6 +187,10 @@ namespace wil
 
     using zstring_view = basic_zstring_view<char>;
     using zwstring_view = basic_zstring_view<wchar_t>;
+
+    constexpr zwstring_view operator "" _zw(const wchar_t* str, std::size_t len) noexcept {
+        return zwstring_view(str, len);
+    }
 #endif // _HAS_CXX17
 
 } // namespace wil
--- a/include/wil/stl.h
+++ b/include/wil/stl.h
@@ -188,13 +188,16 @@ namespace wil
     using zstring_view = basic_zstring_view<char>;
     using zwstring_view = basic_zstring_view<wchar_t>;
 
-    constexpr zstring_view operator "" _zv(const char* str, std::size_t len) noexcept {
-        return zstring_view(str, len);
-    }
+    inline namespace literals {
+        constexpr zstring_view operator "" _zv(const char* str, std::size_t len) noexcept {
+            return zstring_view(str, len);
+        }
 
-    constexpr zwstring_view operator "" _zv(const wchar_t* str, std::size_t len) noexcept {
-        return zwstring_view(str, len);
+        constexpr zwstring_view operator "" _zv(const wchar_t* str, std::size_t len) noexcept {
+            return zwstring_view(str, len);
+        }
     }
+    
 #endif // _HAS_CXX17
 
 } // namespace wil
--- a/include/wil/stl.h
+++ b/include/wil/stl.h
@@ -188,11 +188,11 @@ namespace wil
     using zstring_view = basic_zstring_view<char>;
     using zwstring_view = basic_zstring_view<wchar_t>;
 
-    constexpr zstring_view operator "" zv(const char* str, std::size_t len) noexcept {
+    constexpr zstring_view operator "" _zv(const char* str, std::size_t len) noexcept {
         return zstring_view(str, len);
     }
 
-    constexpr zwstring_view operator "" zv(const wchar_t* str, std::size_t len) noexcept {
+    constexpr zwstring_view operator "" _zv(const wchar_t* str, std::size_t len) noexcept {
         return zwstring_view(str, len);
     }
 #endif // _HAS_CXX17
--- a/include/wil/stl.h
+++ b/include/wil/stl.h
@@ -188,7 +188,11 @@ namespace wil
     using zstring_view = basic_zstring_view<char>;
     using zwstring_view = basic_zstring_view<wchar_t>;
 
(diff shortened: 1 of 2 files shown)


tusharchitti:

include/wil/stl.h
```diff
@@ -187,6 +187,10 @@ namespace wil
 
     using zstring_view = basic_zstring_view<char>;
     using zwstring_view = basic_zstring_view<wchar_t>;
+
```
dmachaj, line 190: I can't recall if we ever added a section to the documentation for zwstring_view.  If we did, would you be willing to add a note about the new literal support?  That is definitely helpful to have and documentation is how people will learn it exists.

dmachaj (commented):
Thank you for contributing :).  I'm happy to see this support getting added.

include/wil/stl.h
```diff
@@ -187,6 +187,10 @@ namespace wil
 
     using zstring_view = basic_zstring_view<char>;
     using zwstring_view = basic_zstring_view<wchar_t>;
+
+    constexpr zwstring_view operator "" _zw(const wchar_t* str, std::size_t len) noexcept {
```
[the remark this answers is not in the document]
tusharchitti: @dmachaj  what marker name do you recommend ? Is there any convention that i should follow ?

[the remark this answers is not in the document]
tusharchitti: i am updating it with zv. (similar to sv (string view))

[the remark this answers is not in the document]
tusharchitti: Looks like for user defined literal _ is necessary.

build log:
D:\a\1\s\include\wil/stl.h(191,28): error: user-defined literal suffixes not starting with '_' are reserved; no literal will invoke this operator [-Werror,-Wuser-defined-literals]
    constexpr zstring_view operator "" zv(const char* str, std::size_t len) noexcept {
                           ^
D:\a\1\s\include\wil/stl.h(195,29): error: user-defined literal suffixes not starting with '_' are reserved; no literal will invoke this operator [-Werror,-Wuser-defined-literals]
    constexpr zwstring_view operator "" zv(const wchar_t* str, std::size_t len) noexcept {
                            ^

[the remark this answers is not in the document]
dmachaj: That's a bummer but it makes sense for the standard library to reserve names.  `_zv` seems reasonable.

I'd also like us to consider `_zsv` as well.  That would be similar to how `zwstring_view` is like `wstring_view` with a z on the front.  So the literal would be `sv` with a z on the front to make `zsv`.  (And then add the underscore because it is required).

I sent mail to a few other reviewers to ask for additional opinions on this topic.

include/wil/stl.h
```diff
@@ -187,6 +187,14 @@ namespace wil
 
     using zstring_view = basic_zstring_view<char>;
     using zwstring_view = basic_zstring_view<wchar_t>;
+
+    constexpr zstring_view operator "" _zv(const char* str, std::size_t len) noexcept {
```
dunhor, line 191: IMO - we probably want to put these in an `inline namespace` (e.g. `literals` like the STL) so that folks don't need to `using namespace wil;` and pull in _everything_

dmachaj: That's a really good point.  `using namespace wil::literals;` is fine.  `using namespace wil;` is recommended against in most cases.

dunhor: If we want to go _really_ crazy, we could even do `namespace wil::inline namespace literals::inline namespace string_view_literals::operator""_zv` (okay, yes, I know that's not valid syntax). I don't think that's entirely necessary, but would help if WIL has multiple `operator""` definitions and `_zv` conflicts with another library/definition.

tests/StlTests.cpp
```diff
@@ -11,7 +11,7 @@ struct dummy
 {
     char value;
 };
-
+using namespace wil;
```
ChrisGuzak, line 14: suspicious namespace alias.

include/wil/stl.h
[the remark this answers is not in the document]
tusharchitti: the email concluded at using "_zv".

include/wil/stl.h
tusharchitti: created an inline namespace.

tusharchitti: something is off with using "using namespace wil::literals" , the current build has failed. Rechecking

duncanhorn: If you haven't figured it out already, it's because the definition of `zstring_view` - and therefore the existence of the namespace - is guarded by `#if _HAS_CXX17`

dmachaj (approved):

include/wil/stl.h
```diff
@@ -187,6 +187,17 @@ namespace wil
 
     using zstring_view = basic_zstring_view<char>;
     using zwstring_view = basic_zstring_view<wchar_t>;
+
+    inline namespace literals {
```
dunhor, line 191: Please move curly braces to a new line

jonwis:
```suggestion
    inline namespace literals
    {
```

tests/StlTests.cpp
ChrisGuzak: our coding guidelines say never create namespace aliases like this for wil and std. 

I understand string literal require an aliases as discussed in the closed comments. I assume you should use that here to demonstrate the best usage.

jonwis (approved):
Thanks!

PR #397, opened 2023-11-21
Update filesystem.h (extra checking the argument of is_extended_length_path)

shalala66:
Since "path" is the argument of wcsncmp and due to SAL (declared in corecrt_wstring.h) for wcsncmp function the "path" can not be NULL that related with "Warning C6011":

_// corecrt_wstring.h_
```
_Check_return_
_ACRTIMP int __cdecl wcsncmp(
    _In_reads_or_z_(_MaxCount) wchar_t const* _String1,
    _In_reads_or_z_(_MaxCount) wchar_t const* _String2,
    _In_                       size_t         _MaxCount
    );
```

Also additionally,  parameter of "is_extended_length_path" function is also declared as _In_ (SAL):

_// filesystem.h (current header)_
inline bool is_extended_length_path(_In_ PCWSTR path)

To prevent this the "path" parameter should be checked via assert or adding if statement:

assert(path != NULL);

### DIFF
--- a/include/wil/filesystem.h
+++ b/include/wil/filesystem.h
@@ -19,6 +19,7 @@
 #include <combaseapi.h> // Needed for CoTaskMemFree() used in output of some helpers.
 #include <winbase.h> // LocalAlloc
 #include <PathCch.h>
+#include <cassert>
 #include "wistd_type_traits.h"
 #include "result.h"
 #include "win32_helpers.h"
@@ -29,6 +30,7 @@ namespace wil
     //! Determines if a path is an extended length path that can be used to access paths longer than MAX_PATH.
     inline bool is_extended_length_path(_In_ PCWSTR path)
     {
+        assert(path != NULL);
         return wcsncmp(path, L"\\\\?\\", 4) == 0;
     }
 


dunhor:
What is the warning that this is trying to fix?

include/wil/filesystem.h
```diff
@@ -29,6 +30,7 @@ namespace wil
     //! Determines if a path is an extended length path that can be used to access paths longer than MAX_PATH.
     inline bool is_extended_length_path(_In_ PCWSTR path)
     {
+        assert(path != NULL);
```
dunhor, line 33: `_In_` presumes not-null, so I'm not entirely sure what this is trying to catch

dunhor:
Also, please update the title to be more descriptive of the change

include/wil/filesystem.h
```diff
@@ -19,6 +19,7 @@
 #include <combaseapi.h> // Needed for CoTaskMemFree() used in output of some helpers.
 #include <winbase.h> // LocalAlloc
 #include <PathCch.h>
+#include <cassert>
```
ChrisGuzak, line 22: some systems use a different assert macro controlled by build system specific macros. notice how there are no other uses of cassert in wil.

include/wil/filesystem.h
shalala66: Yeah, I also already mentioned it in last sentence (Debug ver. or) via offering to add "if statement" to resolve it.

include/wil/filesystem.h
shalala66: You are right, but using assert (or such kind of kindred statements) is for a debug mode as You know. Because of this, I already mentioned about it in my 1st explanation, about using an "if statement" to prevent of sending the "nullptr" (since the "wcsncmp" function doesn't validate its parameters for other situations beside this via giving an exception):

inline bool is_extended_length_path(_In_ PCWSTR path)
{
        if(s)
            return wcsncmp(path, L"\\\\?\\", 4) == 0;
}

dunhor:
Closing for now unless additional information can be provided that explains the issue this is trying to fix. A couple comments:

1. Access to the `cassert` header should not be assumed; prefer use of `WI_ASSERT` for asserts
2. The input to `is_extended_length_path` should be non-null. If the issue is that `wcsncmp` is crashing because of null pointer, prefer letting it crash over asserts
3. If the analyzer is giving warnings because it thinks that the argument can be null even though the parameter is marked as `_In_`, we should prefer investigating that issue

shalala66:
In this scenario (example on the below section) of dynamic memory allocation part which could causes if the malloc function could not able to allocate the memory buffer, it returns NULL and could be arised the "null pointer dereferencing" which can cause to undefined behaviour later. In normal circumstances any program codes should check the pointers returned by the malloc function and pertinently handle the situation when memory could not be allocated. This part of the checking maybe is omitted, so we can avoid undefined behaviors by performing this checking inside the "is_extended_length_path" function just in case. The first appeal to the memory by null pointer or in a differ terms dereferencing the null pointer usually causes the "structured exception" in Windows systems.  As earlier I partly mentioned there are two implicitly potential behaviours related with "Warning C6387" coherented with "NULL value or nullptr of path", since "path" is the argument of wcsncmp and due to SAL for wcsncmp function the "path" can not be NULL or nullptr (although the function parameter is marked as  _In_. Through this kind of example _In_ mark could be  bypassed indirectly) related with "Warning C6011" where it took its initial point which corresponds to "null pointer dereferencing":

```
#include <iostream>
#include <Windows.h>


#if !defined(_MSC_VER)
#define __WI_ANALYSIS_ASSUME(_exp)                          ((void)sizeof(!(_exp))) // Validate syntax on non-debug builds
#else
#define __WI_ANALYSIS_ASSUME(_exp)                          __noop(_exp)
#endif
#define __WI_ASSERT_FAIL_ANNOTATION(msg) __annotation(L"Debug", L"AssertFail", msg)
#define WI_ASSERT(condition)                                (__WI_ANALYSIS_ASSUME(condition), ((!(condition)) ? (__WI_ASSERT_FAIL_ANNOTATION(L"" #condition), DbgRaiseAssertionFailure(), FALSE) : TRUE))


inline bool is_extended_length_path(_In_ PCWSTR path)
{
    WI_ASSERT(path);
    return wcsncmp(path, L"\\\\?\\", 4) == 0;
}


int main(int argc, const char* argv[])
{
    WCHAR* path = (WCHAR*)malloc(0);
    is_extended_length_path(reinterpret_cast<PCWSTR>(path));
    free(path);

    return 0;
}

```
0x00000000 - 0x00000FFF in x86_32 and 0x0000000000000000 - 0x000000000000FFFF in x86_64: In Windows 9x architecture this region of memory for user level was not available for accessing from user mode because of though it's part of user mode region, but it's reserved by OS and played a trap role to detect a null pointers, plus to it contains the null address. Although they are have the close meaning in Windows NT architecture this region is for detection the pointers with wrong values. Trying to access this memory section causes to raising a "memory access violation" exeption. In this case null pointer must be store the value which indicates the fault and serves that this pointer is invalid. That is the target of reservation the null address by the system and of course this is the null pointer related section. But this region is not only included null address. Following the null address, are located the 65535 addresses. That is why the size of this region is 64KiB. It happens due to "data alignment" within comply the sophisticated rules to treatment all addresses with the same way. Despite to it's non-optimize, to deny access to some of addresses all 65536 addresses should to be denied, but this is small amount of memory in compairing with large total amount of user level memory. Null pointer dereferencing may causes a lot of problems started from large or small amount of memory corruptions associated even with compiler optimizations which according to the compiler the nullptr can not be dereferenced and could apply swapping of assignments which causes the "structural exception", but it will happen too late and despite this fact the datas in memory will be corrupted. That is why applying SEH (-> VEH) is not a single option at some in this case. Kindred situations related with compiler optimizations applying swap in assignments could be arises with the use of memset function also. Because of all, I suggest make this kind of changes via adding some checking statement due to on my opinion this code snippet will be more secure hence clear.

shalala66:
@dunhor , @ChrisGuzak

dunhor:
I'm not entirely sure I follow the justification in your response:

>  This part of the checking maybe is omitted, so we can avoid undefined behaviors by performing this checking inside the "is_extended_length_path" function just in case

First, adding an assert will _not_ avoid undefined behavior. Second, as mentioned before, the argument to `is_extended_length_path` is declared as `_In_` which means that it's the caller's responsibility to check for null before calling the function.

> As earlier I partly mentioned there are two implicitly potential behaviours related with "Warning C6387" coherented with "NULL value or nullptr of path", since "path" is the argument of wcsncmp and due to SAL for wcsncmp function the "path" can not be NULL or nullptr (although the function parameter is marked as _In_. Through this kind of example _In_ mark could be bypassed indirectly) related with "Warning C6011" where it took its initial point which corresponds to "null pointer dereferencing"

I pasted your code into [godbolt](https://godbolt.org/z/xdx1ndeK4) and failed to get the C6011 warning you describe, even with `/Wall /analyze`. If you can provide an example project or link to godbolt.org that reproduces this warning, that would be quite useful. Even just copy/pasting the warning you see in its entirety could be useful. I suspect that the warning is _not_ complaining about the call to `wcsncmp`, but instead to your call to `is_extended_length_path` because _you_ forgot to check for a null return value from `malloc`.

shalala66:
First, assert will avoid undefined behavior in this scenario. Second, I did not forget, I just depicted some variation of scenario. Third, maybe you don't understand the real case although I explained it as soon as obviously. Because, you didn't give "enough" attention to "there are two **implicitly** potential behaviours related with Warning C6387.......................... related with Warning C6011 where it took its initial point which corresponds to null pointer dereferencing" as I mentioned above. There are a lot of samples about this behaviour what I mean. In ATL is usually added ATLENSURE or ATLASSERT into the function for extra checking for safety since `_In_` mark does not preventing some kind of behaviours in few situations. I can give you tons of examples from real ATL codes to confirm it if you would not be able to find them ownself - [](https://learn.microsoft.com/en-us/cpp/atl/reference/debugging-and-error-reporting-macros?view=msvc-170#atlassert:~:text=*%20psz)%0A%7B-,ATLENSURE,-(NULL%20!%3D%20psz). I'll give you more precise example to you to catch the real problem stayed beyond of your perspective view:

```
#include <iostream>
#include <Windows.h>

#if !defined(_MSC_VER)
#define __WI_ANALYSIS_ASSUME(_exp)                          ((void)sizeof(!(_exp))) // Validate syntax on non-debug builds
#else
#define __WI_ANALYSIS_ASSUME(_exp)                          __noop(_exp)
#endif
#define __WI_ASSERT_FAIL_ANNOTATION(msg) __annotation(L"Debug", L"AssertFail", msg)
#define WI_ASSERT(condition)                                (__WI_ANALYSIS_ASSUME(condition), ((!(condition)) ? (__WI_ASSERT_FAIL_ANNOTATION(L"" #condition), DbgRaiseAssertionFailure(), FALSE) : TRUE))


inline bool is_extended_length_path(_In_ PCWSTR path)
{
    std::cout << path;
    WI_ASSERT(path);
    return wcsncmp(path, L"\\\\?\\", 4) == 0;
}


int main(int argc, const char* argv[])
{
    WCHAR* path = (WCHAR*)malloc(SIZE_MAX);
    is_extended_length_path(reinterpret_cast<PCWSTR>(path));
    free(path);

    return 0;
}
```

Results:
***Assertion Failuer***
Assertion Failure
***Access Violation***
Access Violation

shalala66:
First, assert will avoid undefined behavior in this scenario. Second, I did not forget, I just depicted some variation of scenario. Third, maybe you don't understand the real case although I explained it as soon as obviously. Because, you didn't give "enough" attention to "there are two **implicitly** potential behaviours related with Warning C6387.......................... related with Warning C6011 where it took its initial point which corresponds to null pointer dereferencing" as I mentioned above. There are a lot of samples about this behaviour what I mean. In ATL is usually added ATLENSURE or ATLASSERT into the function for extra checking for safety since `_In_` mark does not preventing some kind of behaviours in few situations. I can give you tons of examples from real ATL codes to confirm it if you would not be able to find them ownself:
https://learn.microsoft.com/en-us/cpp/atl/reference/debugging-and-error-reporting-macros?view=msvc-170#atlassert:~:text=*%20psz)%0A%7B-,ATLENSURE,-(NULL%20!%3D%20psz

I'll give you more precise example to you to catch the real problem stayed beyond of your perspective view:

```
#include <iostream>
#include <Windows.h>

#if !defined(_MSC_VER)
#define __WI_ANALYSIS_ASSUME(_exp)                          ((void)sizeof(!(_exp))) // Validate syntax on non-debug builds
#else
#define __WI_ANALYSIS_ASSUME(_exp)                          __noop(_exp)
#endif
#define __WI_ASSERT_FAIL_ANNOTATION(msg) __annotation(L"Debug", L"AssertFail", msg)
#define WI_ASSERT(condition)                                (__WI_ANALYSIS_ASSUME(condition), ((!(condition)) ? (__WI_ASSERT_FAIL_ANNOTATION(L"" #condition), DbgRaiseAssertionFailure(), FALSE) : TRUE))


inline bool is_extended_length_path(_In_ PCWSTR path)
{
    std::cout << path;
    WI_ASSERT(path);
    return wcsncmp(path, L"\\\\?\\", 4) == 0;
}


int main(int argc, const char* argv[])
{
    WCHAR* path = (WCHAR*)malloc(SIZE_MAX);
    is_extended_length_path(reinterpret_cast<PCWSTR>(path));
    free(path);

    return 0;
}
```

Results:
**Assertion Failure**
Assertion Failure

**Access Violation**
Access Violation

dunhor:
> First, assert will avoid undefined behavior in this scenario

No, it won't. Asserts are debug mode only checks that are ignorable. They can help you find bugs, sure, but they in no way eliminate undefined behavior.

> I just depicted some variation of scenario

And as I mentioned before, there is a bug in your example. You need to check the result of `malloc` before calling `is_extended_length_path` (which is true for _any_ dynamic memory allocating function that returns null on failure). E.g. your code should look like this:

```c++
int main()
{
    auto path = (wchar_t*)malloc(size);
    if (path)
    {
        // Initialize 'path' in some way

        auto isExtended = wil::is_extended_length_path(path);
        // Do something with 'isExtended'
    }
}
```

shalala66:
No, it will. In this regard, there are examples from sufficiently serious sources, in which case if you are not aware of it or you are not in a position to investigate it yourself. Concerning to asserts you suggested me and I just to describe the real problem used the `WI_ASSERT`. Because in my early comment I used simple if statement. It doesn't matter from the perspective to explain real problem. I knew that what your answer will be approximately and that is why I left the link kindred with exactly this theme on my above comment. `ATLENSURE` throws an exception in Release builds as well as in Debug builds. From your side giving answers like this is absurd as I assumed. Again you did not getting enough attention to my wide explanation on the above. Maybe it's not related with your attention, you just dont understand what is it about. I know where is needed to check the result from malloc. You again did not catch the real case. It's very ridiculous, but you again did not understand that I just give an example of some scenario in which it's will cause a problem in your function. In your early comments you also talked about `_In_` mark declaration and claimed that it's enough helps for detection the warnings, but I proved you wrong on my last comment (see images). Now you are talking about bug in my exampIe. It's really very funny. I dont know how can I explain it to you in other ways. I created this scenario purposefully to show the importance of adding extra checker inside the your function. As I mentioned on my last comment I can give you tons of examples from real ATL source codes coherent with adding extra checking inside the function regardless of how it's defined from outside of the function to confirm it if you would not be able to find them yourself.

dunhor:
> `ATLENSURE` throws an exception in Release builds as well as in Debug builds

Your original PR used `assert` not `ATLENSURE`, which behaves similarly to `WI_ASSERT`. Something more equivalent to `ATLENSURE` in WIL would be `THROW_HR_IF` or `FAIL_FAST_IF`. But again, neither of those are necessary because the function documents itself with a precondition that the argument is not null. Any example that does not properly check the return value of `malloc` is a strawman.

Locking the thread for now as you've shown that you are unreceptive to feedback and things have started to go off the rails.

PR #234, opened 2022-04-23
Add enum_iterator type

asklar:
This change adds an `enum_iterator` type which allows iterating in a natural way over `IEnum`-like COM interfaces, like `IEnumIDList`, `IDiaEnumSymbols`, etc.

bonus track: also fixes a few unrelated typos

PR #398, opened 2023-11-26
IEnumXxx helpers v2

asklar:
This is another [simplified] attempt at the range-based iterator helpers for IEnumXYZ-style interfaces (v1 can be found in #234).
Merging @jonwis's, @oldnewthing's and my approach into this PR.
Some notable deltas:
- copy ctor/operator= need to be exposed
- make_range needs to return something that has begin() and end(), not a pair

### DIFF
--- a/include/wil/com.h
+++ b/include/wil/com.h
@@ -3002,6 +3002,135 @@ namespace wil
 
 #endif // __IObjectWithSite_INTERFACE_DEFINED__
 
+
+// if C++17 or greater
+#if WIL_HAS_CXX_17
+#ifdef WIL_ENABLE_EXCEPTIONS
+namespace details
+{
+    template<typename>
+    struct com_enumerator_next_traits;
+
+    template<typename Itf, typename T, typename Ret>
+    struct com_enumerator_next_traits<Ret(__stdcall Itf::*)(ULONG, T*, ULONG*)>
+    {
+        using Interface = Itf;
+        using Result = T;
+    };
+
+
+    template<typename Interface>
+    struct com_enumerator_traits
+    {
+        using Result = typename com_enumerator_next_traits<decltype(&Interface::Next)>::Result;
+        using smart_result = std::conditional_t<
+            std::is_pointer_v<Result> &&
+            std::is_base_of_v<::IUnknown, std::remove_pointer_t<Result>>, wil::com_ptr<std::remove_pointer_t<Result>>, Result>;
+    };
+}
+
+template<typename T>
+struct is_com_ptr : std::false_type {};
+
+template<typename T>
+struct is_com_ptr<wil::com_ptr<T>> : std::true_type {};
+
+template <typename IEnumType, typename TStoredType = typename details::com_enumerator_traits<IEnumType>::smart_result>
+struct iterator
+{
+    wil::com_ptr<IEnumType> m_enum{};
+    TStoredType m_currentValue{};
+
+    static constexpr bool is_com_ptr_v = is_com_ptr<TStoredType>::value;
+
+    iterator(iterator&&) = default;
+    iterator(iterator const&) = default;
+    iterator& operator=(iterator&&) = default;
+    iterator& operator=(iterator const&) = default;
+
+    iterator(IEnumType* enumPtr) : m_enum(enumPtr)
+    {
+        FetchNext();
+    }
+
+    auto operator->()
+    {
+        return wistd::addressof(m_currentValue);
+    }
+
+    auto& operator*()
+    {
+        return m_currentValue;
+    }
+
+    iterator& operator++()
+    {
+        // If we're already at the end, don't try to advance. Otherwise, use Next to advance.
+        if (m_enum)
+        {
+            FetchNext();
+        }
+
+        return *this;
+    }
+
+    bool operator!=(iterator const& other) const
+    {
+        return !(*this == other);
+    }
+
+    bool operator==(iterator const& other) const
+    {
+        return (m_enum.get() == other.m_enum.get());
+    }
+
+private:
+    void FetchNext()
+    {
+        if (m_enum)
+        {
+            m_currentValue = TStoredType{};
+            auto hr = m_enum->Next(1, &m_currentValue, nullptr);
+            if (hr == S_FALSE)
+            {
+                m_enum = nullptr;
+            }
+            else 
+            {
+                THROW_IF_FAILED_MSG(hr, "Failed to get next");
+            }
+        }
+    }
+};
+
+template<typename IEnumXXX>
+struct iterator_range
+{
+    iterator<IEnumXXX> m_begin;
+
+    iterator_range(IEnumXXX* enumPtr) : m_begin(enumPtr)
+    {
+    }
+
+    WI_NODISCARD auto begin()
+    {
+        return m_begin;
+    }
+
+    WI_NODISCARD constexpr auto end() const noexcept
+    {
+        return iterator<IEnumXXX>(nullptr);
+    }
+};
+
+template<typename IEnumXxx, typename TStoredType = typename wil::details::com_enumerator_traits<IEnumXxx>::smart_result>
+WI_NODISCARD auto make_range(IEnumXxx* enumPtr)
+{
+    return wil::iterator_range<IEnumXxx>(enumPtr);
+}
+#endif // WIL_HAS_CXX_17
+#endif // WIL_ENABLE_EXCEPTIONS
+
 } // wil
 
 #endif
--- a/tests/ComTests.cpp
+++ b/tests/ComTests.cpp
@@ -5,7 +5,10 @@
 #include <wrl/implements.h>
 
 #include "common.h"
-
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+#include <ShObjIdl_core.h>
+#include <ShlObj_core.h>
+#endif
 #include <Bits.h>
 
 using namespace Microsoft::WRL;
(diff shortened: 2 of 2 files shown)


include/wil/com.h
```diff
@@ -3002,6 +3002,135 @@ namespace wil
...
+        using Result = T;
+    };
+
+
+    template<typename Interface>
+    struct com_enumerator_traits
+    {
+        using Result = typename com_enumerator_next_traits<decltype(&Interface::Next)>::Result;
+        using smart_result = std::conditional_t<
+            std::is_pointer_v<Result> &&
```
jonwis, line 3027: Consider moving the conditionals into this line, then the output types onto the next.

include/wil/com.h
```diff
@@ -3002,6 +3002,135 @@ namespace wil
...
+            else 
+            {
+                THROW_IF_FAILED_MSG(hr, "Failed to get next");
+            }
+        }
+    }
+};
+
+template<typename IEnumXXX>
+struct iterator_range
```
jonwis, line 3107: Should this type be declared inside make_range so it's not just hanging around visible here?

Similarly with the iterator<T> - if it's local to make_range then you have to obtain one from make_range...

include/wil/com.h
```diff
@@ -3002,6 +3002,135 @@ namespace wil
...
+        return m_begin;
+    }
+
+    WI_NODISCARD constexpr auto end() const noexcept
+    {
+        return iterator<IEnumXXX>(nullptr);
+    }
+};
+
+template<typename IEnumXxx, typename TStoredType = typename wil::details::com_enumerator_traits<IEnumXxx>::smart_result>
```
jonwis, line 3126: Should TStoredType be removed, or used in the method body?

tests/ComTests.cpp
```diff
@@ -2819,4 +2822,72 @@ TEST_CASE("StreamTests::Saver", "[com][IStream]")
...
+    }
+    {
+        wil::com_ptr<IShellFolder> desktop;
+        REQUIRE_SUCCEEDED(::SHGetDesktopFolder(&desktop));
+        wil::com_ptr<IEnumIDList> enumIDList;
+        REQUIRE_SUCCEEDED(desktop->EnumObjects(nullptr, SHCONTF_NONFOLDERS, &enumIDList));
+        REQUIRE(enumIDList);
+        
+        auto count = 0;
+        for (auto pidl : wil::make_range(enumIDList.get()))
```
jonwis, line 2880: Consider also a test on a mock "zero element" enumerator.

include/wil/com.h
asklar: I'll move the `iterator_range` type inside `make_range`. I'd like to keep the iterator itself visible - esp. as it allows the types in the iterator to be tested, lmk if you disagree.

tests/ComTests.cpp
```diff
@@ -2819,4 +2822,127 @@ TEST_CASE("StreamTests::Saver", "[com][IStream]")
...
+        {
+            REQUIRE(muffin == nullptr);
+            count++;
+            break;
+        }
+        REQUIRE(count == 0);
+        
+    }
+    {
+#if (NTDDI_VERSION >= NTDDI_VISTA)
```
dmachaj, line 2903: This is already in an ifdef cpp17 block.  Can you even compile pre-Vista headers in C++ 17 mode?

asklar: not sure but i don't think we should tie together/assume lang version x SDK version. One check is for one thing, the other is for another, if the combination of C++17 w/pre-Vista is invalid, there's no harm (this check just makes it clear that this API was added in Vista)

jonwis (approved):

asklar:
@jonwis I don't have merge permissions so if you're ok with it, can you press the `<button/>`?

include/wil/com.h
```diff
@@ -3002,6 +3002,135 @@ namespace wil
...
+        using Result = T;
+    };
+
+
+    template<typename Interface>
+    struct com_enumerator_traits
+    {
+        using Result = typename com_enumerator_next_traits<decltype(&Interface::Next)>::Result;
+        // If the result is a COM pointer type (IFoo*), then we use wil::com_ptr<IFoo>. Otherwise, we use IFoo*.
+        using smart_result = std::conditional_t<std::is_pointer_v<Result> && std::is_base_of_v<::IUnknown, std::remove_pointer_t<Result>>,
```
dunhor, line 3034: `wistd::*`

include/wil/com.h
```diff
@@ -3002,6 +3002,135 @@ namespace wil
...
+        using Interface = Itf;
+        using Result = T;
+    };
+
+
+    template<typename Interface>
+    struct com_enumerator_traits
+    {
+        using Result = typename com_enumerator_next_traits<decltype(&Interface::Next)>::Result;
+        // If the result is a COM pointer type (IFoo*), then we use wil::com_ptr<IFoo>. Otherwise, we use IFoo*.
```
dunhor, line 3033:
```suggestion
        // If the result is a COM pointer type (IFoo*), then we use wil::com_ptr<IFoo>. Otherwise, we use the raw type.
```
A little confusing

include/wil/com.h
```diff
@@ -3002,6 +3002,135 @@ namespace wil
...
+        using smart_result = std::conditional_t<std::is_pointer_v<Result> && std::is_base_of_v<::IUnknown, std::remove_pointer_t<Result>>,
+            wil::com_ptr<std::remove_pointer_t<Result>>, Result>;
+    };
+}
+
+template <typename IEnumType, typename TStoredType = typename details::com_enumerator_traits<IEnumType>::smart_result>
+struct iterator
```
dunhor, line 3040:
```suggestion
struct com_enum_iterator
```
Given that this is directly in the `wil` namespace, the name should be more specific

include/wil/com.h
```diff
@@ -3002,6 +3002,135 @@ namespace wil
...
+            {
+                THROW_IF_FAILED_MSG(hr, "Failed to get next");
+            }
+        }
+    }
+};
+
+
+template<typename IEnumXxx, typename TStoredType = typename wil::details::com_enumerator_traits<IEnumXxx>::smart_result>
+WI_NODISCARD auto make_range(IEnumXxx* enumPtr)
```
dunhor, line 3108: This needs either a new name (so it doesn't conflict with other `make_range` functions, or it needs SFINAE

asklar: ah got it, that makes sense. the other make_range overloads are:
- make_range(T begin, T end)
- make_range(T begin, size_t count)

This overload is make_range(T), so it shouldn't conflict as long as there isn't another one-param overload. Right?

dunhor: Isn't there `make_range(IVector<T>*)` (& company)?

Even so, I don't think those would conflict since they are more "specific" and would "win" in overload resolution. However, we should still future-proof where we can. Enabling this overload only when the type has `T::Next` (likely of the desired signature) is certainly doable.

dunhor: Nvm, that's `get_range(IVector<T>`, etc. Second comment still stands (and now I'm kind of sad we have two naming conventions here)

asklar: I'm not sure how you're proposing that we do it while at the same time having a default TStoredType param... did you have a proposal in mind?

dunhor: Is it required to make `TStoredType` a template argument? If not, the function body can deduce this type and it need not be part of the signature. Otherwise, you can do this [the classic SFINAE way](https://godbolt.org/z/881EE5h7x)

asklar: I've moved TStoredType to not be a template argument

include/wil/com.h
```diff
@@ -3002,6 +3002,150 @@ namespace wil
...
+    struct has_next
+    {
+      template <typename U = T>
+      static auto test(int) -> decltype(wistd::declval<U>()->Next(0, nullptr, nullptr), wistd::true_type{});
+
+      template <typename>
+			static auto test(...) -> wistd::false_type;
+
+			static constexpr bool value = decltype(test<T>(0))::value;
+    };
```
dunhor, line 3038:
```suggestion
    template<typename T>
    struct has_next
    {
        template <typename U = T>
        static auto test(int) -> decltype(wistd::declval<U>()->Next(0, nullptr, nullptr), wistd::true_type{});

        template <typename>
        static auto test(...) -> wistd::false_type;

        static constexpr bool value = decltype(test<T>(0))::value;
    };
```
Something wonky happened with the formatting here

include/wil/com.h
```diff
@@ -3002,6 +3002,150 @@ namespace wil
...
+
+    template<typename T>
+    constexpr bool has_next_v = has_next<T>::value;
+
+    template<typename Interface, wistd::enable_if_t<has_next_v<Interface*>, int> = 0>
```
dunhor, line 3043: nit: I don't think the `enable_if` is doing anything here?

dunhor:
A couple additional notes that are not urgent as they can be added later if desired:
1. The iterator type lacks an `iterator_category` typedef (and other typedefs), which may limit its usability in STL or other algorithms. This should probably be an input iterator I think?
2. The iterator type lacks a postfix `operator++(int)` which may also limit its usability in algorithms, and is a requirement for input iterators

include/wil/com.h
asklar: i figured we'd want to sfinae this too? can remove if it's not useful.

include/wil/com.h
asklar: oh weird. I think that was probably ghcp using tabs instead of spaces :(

include/wil/com.h
dunhor: SFINAE on class templates _can_ be useful if you've got a number of definitions fighting each other (I've used the `void_t` trick on a number of occasions for this). In this case, it's not all that useful since the traits type shouldn't ever be used with a non-IEnum* type

include/wil/com.h
```diff
@@ -3002,6 +3002,150 @@ namespace wil
...
+    struct com_enumerator_traits
+    {
+        using Result = typename com_enumerator_next_traits<decltype(&Interface::Next)>::Result;
+        // If the result is a COM pointer type (IFoo*), then we use wil::com_ptr<IFoo>. Otherwise, we use the raw pointer type IFoo*.
+        using smart_result = wistd::conditional_t<std::is_pointer_v<Result> && wistd::is_base_of_v<::IUnknown, wistd::remove_pointer_t<Result>>,
```
dunhor, line 3048:
```suggestion
        using smart_result = wistd::conditional_t<wistd::is_pointer_v<Result> && wistd::is_base_of_v<::IUnknown, wistd::remove_pointer_t<Result>>,
```
Missed this one on prior review

dunhor (approved):

asklar:
@dunhor going to write a couple more tests, will ping back in a sec :)

include/wil/com.h
```diff
@@ -3002,6 +3002,155 @@ namespace wil
...
+            else 
+            {
+                THROW_IF_FAILED_MSG(hr, "Failed to get next");
+            }
+        }
+    }
+};
+
+
+template<typename IEnumXxx, wistd::enable_if_t<wil::details::has_next_v<IEnumXxx*>, int> = 0>
```
oldnewthing, line 3126: We should allow the caller to specify the custom smart type to `make_range`. 

```cpp
for (auto&& pidl : make_range<IEnumIDList, wil::unique_pidl>(penum.get()))
{
   ...
}
```

Otherwise you have a memory leak because we leak all the pidls.

We should require a smart type if the enumerator produces a pointer.

```cpp
for (auto&& pidl : make_range(penum.get())) // should be illegal - leaks the pidls
{
   ...
}
```



Unfortunately, we cannot distinguish between returning structures with nothing worth destructing
```
struct SafeToJustAbandon
{
   int first, last;
};
```

and structures that require cleanup.

```cpp
struct MustCleanMeUpOrIWillLeak
{
    VARIANT v; // if you fail to VariantClear me, you will leak memory!
};
```

include/wil/com.h
```diff
@@ -3002,6 +3002,155 @@ namespace wil
...
+    struct com_enumerator_traits
+    {
+        using Result = typename com_enumerator_next_traits<decltype(&Interface::Next)>::Result;
+        // If the result is a COM pointer type (IFoo*), then we use wil::com_ptr<IFoo>. Otherwise, we use the raw pointer type IFoo*.
+        using smart_result = wistd::conditional_t<wistd::is_pointer_v<Result> && wistd::is_base_of_v<::IUnknown, wistd::remove_pointer_t<Result>>,
+            wil::com_ptr<wistd::remove_pointer_t<Result>>, Result>;
+    };
+}
+
+template <typename IEnumType, typename TStoredType = typename details::com_enumerator_traits<IEnumType>::smart_result>
```
oldnewthing, line 3053: We should use [type deduction from the future](https://devblogs.microsoft.com/oldnewthing/20230609-00/?p=108318) to allow the caller to specify only the TStoredType.

```cpp
// cumberwome
auto enumerator = com_iterator<IEnumIDList, wil::unique_idlist>(penum.get());

// better ergonomics
auto enumerator = com_iterator<wil::unique_idlist>(penum.get());
```

include/wil/com.h
asklar: @oldnewthing Originally this had the TStoredType with a default value of `smart_result`. However, @dunhor mentioned we should sfinae this method, so that requires an optional non-type template param.
So in order to support both I think you'd have to write `make_range<IEnumIDList, int, wil::unique_pidl>` (because putting the TStoredType first causes multiple definition IIRC), see https://github.com/microsoft/wil/pull/398#discussion_r1411281525

dunhor: _Technically_ the SFINAE part doesn't need to be a template parameter (though with an `auto` return type, the list of tricks available to you decreases). The original code also had a problem where an additional compilation error was coming in due to a non-dependent (i.e. not a substitution failure) access of the `T::Next` function pointer.

In Raymond's example, `TStoredType` is deduced "as necessary" and doesn't seem to have any of the original issues: https://godbolt.org/z/GTTaa6b7h

oldnewthing: Wearing my "samples-driven API design" hat, I figure that we want the calling code to look like this:
```cpp
wil::com_ptr<IEnumUnknown> enumUnknown;
wil::com_ptr<IEnumIDList> enumIDList;

// allowed: deduces IEnumUnknown and com_ptr<IUnknown>
auto enumerator = wil::com_enumerator(enumUnknown.get());

// allowed: deduces IEnumUnknown and uses caller-specified com_ptr_nothrow<IUnknown>
auto enumerator = wil::com_enumerator<wil::com_ptr_nothrow<IUnknown>>(enumUnknown.get());

// allowed: deduces IEnumIDList, uses explicitly-provided unique_idlist
auto enumerator = wil::com_enumerator<wil::unique_idlist>(enumIDLIst.get());

// not allowed - need to specify how to free the IDList
auto enumerator = wil::com_enumerator(enumIDLIst.get());
```
And similarly for `make_range`.

This means that we have
```cpp
template<typename TStoredType = void, typename IEnumXxx>
struct com_enumerator
{
    static_assert(!std::is_same_v<typename com_enumerator_traits::Result, typename com_enumerator_traits::smart_result> || !std::is_same_v<TStoredType, void>, "You must specify a type to receive the enumerated objects.")
};
```
possibly with SFINAE stuff added to fail the com_enumerator at deduction time rather than at instantiation.

dunhor: > possibly with SFINAE stuff added to fail the com_enumerator at deduction time rather than at instantiation.

If it's just a struct, there's no real need for SFINAE because there's no overload resolution for structs. It's _occasionally_ useful to SFINAE on specializations, but I can't imagine that being necessary here

include/wil/com.h
asklar: See #404

include/wil/com.h
asklar: following up in #404

PR #401, opened 2023-11-29
Add a .clang-format file to WIL

dunhor:
# NOTICE:

⭐ This is starting out as a draft PR so that it's easier for folks to see the changes/impact of the `.clang-format` file and its options

⭐ The `.clang-format` file has originally been pulled from the Windows OS repo unmodified. While we may end up changing parts of it, this is the exception, not the rule. We are not soliciting feedback based off style preferences

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,223 @@
+﻿---
+Language: Cpp
+# BasedOnStyle: LLVM
+
+AccessModifierOffset: -4
+AlignAfterOpenBracket: AlwaysBreak
+AlignConsecutiveAssignments: false
+AlignEscapedNewlines: DontAlign
+AlignOperands: true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLambdasOnASingleLine: Empty
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: true
+AlwaysBreakTemplateDeclarations: true
+BinPackArguments: false
+BinPackParameters: false
+BraceWrapping:
+  AfterCaseLabel: true
+  AfterClass: true
+  AfterControlStatement: true
+  AfterEnum: true
+  AfterFunction: true
+  AfterNamespace: true
+  AfterStruct: true
+  AfterUnion: true
+  AfterExternBlock: true
+  BeforeCatch: true
+  BeforeElse: true
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+BreakBeforeBinaryOperators: None
+BreakBeforeBraces: Custom
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializers: AfterColon
+BreakStringLiterals: false
+ColumnLimit: 130
+CommentPragmas: '^ IWYU pragma:'
+CompactNamespaces: true
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+DisableFormat: false
+FixNamespaceComments: true
+ForEachMacros: [ foreach, Q_FOREACH, BOOST_FOREACH ]
+IncludeBlocks: Regroup
+IndentCaseLabels: false
+IncludeCategories:
+  - Regex: '^"(stdafx.h|pch.h|precomp.h)"$'
+    Priority: -1
+IndentWidth: 4
+IndentWrappedFunctionNames: false
+KeepEmptyLinesAtTheStartOfBlocks: true
+MacroBlockBegin: '^BEGIN_COM_MAP$|^BEGIN_CONNECTION_POINT_MAP$|^BEGIN_HELPER_NODEMAP$|^BEGIN_MODULE$|^BEGIN_MSG_MAP$|^BEGIN_OBJECT_MAP$|^BEGIN_TEST_CLASS$|^BEGIN_TEST_METHOD$|^BEGIN_TEST_METHOD_PROPERTIES$'
+MacroBlockEnd: '^END_COM_MAP$|^END_CONNECTION_POINT_MAP$|^END_HELPER_NODEMAP$|^END_MODULE$|^END_MSG_MAP$|^END_OBJECT_MAP$|^END_TEST_CLASS$|^END_TEST_METHOD$|^END_TEST_METHOD_PROPERTIES$'
+MaxEmptyLinesToKeep: 1
+NamespaceIndentation: Inner
+ObjCBlockIndentWidth: 2
+ObjCSpaceAfterProperty: false
+ObjCSpaceBeforeProtocolList: true
+PenaltyBreakBeforeFirstCallParameter: 19
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakString: 2000
+PenaltyExcessCharacter: 2
+PenaltyReturnTypeOnItsOwnLine: 1000
+PointerAlignment: Left
+SortIncludes: false
+SpaceAfterCStyleCast: false
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles: Never
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard: Latest
+TabWidth: 4
+UseTab: Never
+
+AttributeMacros: [
+  CALLBACK,
+]
+
+StatementMacros: [
+  _Acquires_exclusive_lock_,
+  _Acquires_lock_,
+  _Acquires_nonreentrant_lock_,
+  _Acquires_shared_lock_,
+  _Analysis_assume_smart_lock_acquired_,
+  _Analysis_assume_smart_lock_released_,
+  _Create_lock_level_,
+  _Detaches_lock_,
+  _Function_class_,
+  _Global_cancel_spin_lock_,
+  _Global_critical_region_,
+  _Global_interlock_,
+  _Global_priority_region_,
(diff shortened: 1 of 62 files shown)


.clang-format
```diff
@@ -0,0 +1,215 @@
...
+IncludeCategories:
+  - Regex: '^"(stdafx.h|pch.h|precomp.h)"$'
+    Priority: -1
+IndentWidth: 4
+IndentWrappedFunctionNames: false
+KeepEmptyLinesAtTheStartOfBlocks: true
+MacroBlockBegin: '^BEGIN_COM_MAP$|^BEGIN_CONNECTION_POINT_MAP$|^BEGIN_HELPER_NODEMAP$|^BEGIN_MODULE$|^BEGIN_MSG_MAP$|^BEGIN_OBJECT_MAP$|^BEGIN_TEST_CLASS$|^BEGIN_TEST_METHOD$|^BEGIN_TEST_METHOD_PROPERTIES$'
+MacroBlockEnd: '^END_COM_MAP$|^END_CONNECTION_POINT_MAP$|^END_HELPER_NODEMAP$|^END_MODULE$|^END_MSG_MAP$|^END_OBJECT_MAP$|^END_TEST_CLASS$|^END_TEST_METHOD$|^END_TEST_METHOD_PROPERTIES$'
+MaxEmptyLinesToKeep: 1
+NamespaceIndentation: Inner
```
dunhor, line 66: Not a huge fan of this & conflicts heavily with existing style. Would prefer `All`

include/wil/common.h
```diff
@@ -92,31 +93,231 @@
...
 #define __WI_HAS_VA_OPT_IMPL(F, T, ...) T
-#define __WI_HAS_VA_OPT_(...) __WI_HAS_VA_OPT_IMPL(__VA_OPT__(0,) 1, 0)
+#define __WI_HAS_VA_OPT_(...) __WI_HAS_VA_OPT_IMPL(__VA_OPT__(0, ) 1, 0)
 /// @endcond
 
 //! Evaluates to '1' when support for '__VA_OPT__' is available, else '0'
 #define WI_HAS_VA_OPT __WI_HAS_VA_OPT_(unused)
 
 /// @cond
-#define __WI_ARGS_COUNT1(A0, A1, A2, A3, A4, A5, A6, A7, A8, A9, A10, A11, A12, A13, A14, A15, A16, A17, A18, A19, A20, A21, A22, A23, A24, A25, A26, A27, A28, A29, \
```
dunhor, line 113: An obvious candidate for `// clang-format off` / `// clang-format on`

include/wil/com.h
```diff
@@ -27,2099 +27,2160 @@
...
-    {
-        template <typename T>
-        class ComPtr;
-    }
+namespace WRL
+{
+    template <typename T>
+    class ComPtr;
 }
+} // namespace Microsoft
```
dmachaj, line 35: I'm not sure how I feel about the comment on the closing brace of each namespace.  For short ones like this it is overkill.

dmachaj:
It is very churny but I like that it is fixing some indents that are clearly wrong.  It is also nice that it auto-wraps comments at a reasonable level (or un-wraps things that were split across lines for no good reason).

.clang-format
dmachaj: Agreed, this one is churning files super aggressively.  Some files are almost 100% diff if you turn off ignore whitespace.  This also disagrees with the preferred style of everything I've ever seen.

dmachaj:
> #ifndef _WRL_IMPLEMENTS_H_

Private copies of WRL should probably not be style checked

---
Refers to: tests/workarounds/wrl/wrl/implements.h:8 in b558095. [](commit_id = b558095ff71b7d5fd8a95be8eb670f6d6dda862a, deletion_comment = False)

include/wil/com.h
dunhor: I've never personally liked it either, but it's relatively minor and easy to ignore that idk if it's worth diverging from the OS `.clang-format` file

dunhor:
> > #ifndef _WRL_IMPLEMENTS_H_
> 
> Private copies of WRL should probably not be style checked
> 
> Refers to: tests/workarounds/wrl/wrl/implements.h:8 in [b558095](https://github.com/microsoft/wil/commit/b558095ff71b7d5fd8a95be8eb670f6d6dda862a). [](commit_id = [b558095](https://github.com/microsoft/wil/commit/b558095ff71b7d5fd8a95be8eb670f6d6dda862a), deletion_comment = False)

Yeah, probably. For now I'm just wanting files to be review-able; I'll definitely go back & fine-tune things later. Though frankly, I can probably delete those files now as the fixes are almost certainly in an SDK by now

dunhor:
> inline IInspectable* com_raw_ptr(T ^ ptr)

Hmm. Idk how, if possible at all, to get proper support for C++/CX pointer syntax

---
Refers to: include/wil/com.h:1237 in b558095. [](commit_id = b558095ff71b7d5fd8a95be8eb670f6d6dda862a, deletion_comment = False)

include/wil/com.h
```diff
@@ -27,2099 +27,2160 @@
...
 
 #ifdef __cplusplus_winrt
 
-    template <typename T>
-    inline IInspectable* com_raw_ptr(T^ ptr)
-    {
-        return reinterpret_cast<IInspectable*>(static_cast<::Platform::Object^>(ptr));
-    }
+template <typename T>
+inline IInspectable* com_raw_ptr(T ^ ptr)
```
dunhor, line 1237: Hmm. Idk how, if possible at all, to get proper support for C++/CX pointer syntax

.clang-format
ChrisGuzak: see [osgwiki](https://www.osgwiki.com/wiki/Cpp_Core_Guidelines#Subjective_Settings) about this one. I initially took your view, but eventually agreed that this is a judgement call and many reasonable folks have chosen Inner and that we would be better of picking one setting.

dunhor: Yeah, I agree about preferring no changes for the sake of consistency unless there's unanimous consensus among us. For this instance, I see two main non-personal preference related reasons to possibly change things here:
1. The churn of this change is dramatically increased, meaning higher risk of merge conflicts with other changes, either in the OS repo or in still-active PRs here
2. WIL header files can be very large, and it's useful to be able to quickly tell if line(s) are inside or outside the `wil` namespace

.clang-format
```diff
@@ -0,0 +1,219 @@
...
+]
+
+TypenameMacros: [
+  IFACEMETHOD,
+  STDMETHOD,
+  STDAPI_,
+]
+
+WhitespaceSensitiveMacros: [
+  __WI_CLANG_DISABLE_WARNING,
```
dunhor, line 182: Note that this has been custom added by me. Otherwise `clang-format` tries to separate `-` characters for Clang warning names

include/wil/result_macros.h
```diff
@@ -583,79 +658,421 @@ WI_ODR_PRAGMA("WIL_FreeMemory", "0")
...
-#define __RETURN_WIN32_FAIL(err, str)                        __WI_SUPPRESS_4127_S do { const DWORD __err = (err); return __R_FN(Return_Win32)(__R_INFO(str) __err); } __WI_SUPPRESS_4127_E while ((void)0, 0)
-#define __RETURN_GLE_FAIL(str)                               return __R_FN(Return_GetLastError)(__R_INFO_ONLY(str))
-#define __RETURN_GLE_FAIL_NOFILE(str)                        return __R_FN(Return_GetLastError)(__R_INFO_NOFILE_ONLY(str))
-#define __RETURN_NTSTATUS(status, str)                       __WI_SUPPRESS_4127_S do { const NTSTATUS __status = (status); if (FAILED_NTSTATUS(__status)) { return __R_FN(Return_NtStatus)(__R_INFO(str) __status); } return S_OK; } __WI_SUPPRESS_4127_E while ((void)0, 0)
-#define __RETURN_NTSTATUS_FAIL(status, str)                  __WI_SUPPRESS_4127_S do { const NTSTATUS __status = (status); return __R_FN(Return_NtStatus)(__R_INFO(str) __status); } __WI_SUPPRESS_4127_E while ((void)0, 0)
+#define __RETURN_HR_MSG(hr, str, fmt, ...) \
```
dunhor, line 667: How do folks feel about these macros getting formatted? I tend to like it, however there are _a lot_ of these types of macros in this file.

FYI - in case it's not clear, I'm asking about wrapping these with `clang-format off` / `clang-format on`

dunhor: Seems to add ~600 lines for these macros

dunhor: `tracelogging.h` nearly doubled in size 😮

include/wil/winrt.h
```diff
@@ -66,1216 +66,1270 @@ namespace std
...
-            typedef const TSmart& reference;
-
-            // for begin()
-            vector_iterator(VectorType* v, unsigned int pos)
-                : m_v(v), m_i(pos)
-            {
-            }
+    template <typename LhsT, typename RhsT>
+    WI_NODISCARD auto operator()(const LhsT& lhs, const RhsT& rhs) const
+        WI_NOEXCEPT->decltype(details::hstring_compare<true, false>::less(lhs, rhs))
```
dunhor, line 385: Interesting behavior: looks like it's treating `WI_NOEXCEPT` as a variable, so it seems to be assuming that `->` is doing pointer access

dunhor: Looks like `clang-format` version 17 added a `Macros` key that allows specifying mapping of definition to expansion, which would presumably fix this. Sadly it looks like VS ships with version 16 currently :(

dunhor: Tested it out and while I can confirm that it fixes cases like these, for some reason it also causes more aggressive line wrapping that doesn't occur without the `Macros` key, nor does it occur with the `noexcept` keyword itself. Seems like a bug in `clang-format`

dunhor: Using `Macros` also fixes the weird formatting of member initializers on a `noexcept` constructor

include/wil/registry.h
```diff
@@ -26,3244 +26,3279 @@
...
-        //
-        //     wil::shared_cotaskmem_string shared_string_value{};
-        //     hr = wil::reg::get_value_nothrow(key, L"string_value_name", shared_string_value);
-        //     // or can specify explicitly reading a string into a wil::shared_cotaskmem_string type
-        //     hr = wil::reg::get_value_string_nothrow(key, L"string_value_name", shared_string_value);
-        //
-
-        /**
-         * \brief Reads a value under a specified key, the registry type based off the templated type passed as data
-
```
dunhor, line 2315: For some reason `clang-format` dislikes this style of doxygen comments that use backslash :/

dunhor: Using `@` syntax doesn't have this problem, but has a different issue: lines don't reflow at all

dunhor: I think it might be best to
1. Use `@` consistently instead of `\` (someone let me know if they are not interchangeable)
2. Manually format the doxygen code blocks

PR #403, opened 2023-12-03
add for_each_window and for_each_thread_window helpers

asklar:
Adds some helpers to make iterating through hwnds a little nicer

### DIFF
--- a/include/wil/windowing.h
+++ b/include/wil/windowing.h
@@ -32,8 +32,7 @@ namespace details
     void CallCallbackNoThrow(TEnumApi&& enumApi, TCallback&& callback) noexcept
     {
         auto enumproc = [](HWND hwnd, LPARAM lParam) -> BOOL {
-            auto pCallbackData = reinterpret_cast<decltype(&callbackData)>(lParam);
-            auto pCallback = pCallbackData->pCallback;
+            auto pCallback = reinterpret_cast<TCallback*>(lParam);
             using result_t = decltype((*pCallback)(hwnd));
 #ifdef __cpp_if_constexpr
             if constexpr (std::is_void_v<result_t>)
@@ -58,7 +57,7 @@ namespace details
             return (*pCallback)(hwnd);
 #endif
         };
-        enumApi(enumproc, reinterpret_cast<LPARAM>(&callbackData));
+        enumApi(enumproc, reinterpret_cast<LPARAM>(&callback));
     }
 
 #ifdef WIL_ENABLE_EXCEPTIONS
--- a/include/wil/windowing.h
+++ b/include/wil/windowing.h
@@ -11,6 +11,10 @@
 #ifndef __WIL_WINDOWING_INCLUDED
 #define __WIL_WINDOWING_INCLUDED
 
+#ifndef WIL_ENABLE_EXCEPTIONS
+#error This header requires exceptions
+#endif
+
 #include <WinUser.h>
 #include <exception>
 
@@ -24,6 +28,41 @@ namespace details
     {
     };
 
+    template<typename TEnumApi, typename TCallback>
+    void CallCallbackNoThrow(TEnumApi&& enumApi, TCallback&& callback) noexcept
+    {
+        auto enumproc = [](HWND hwnd, LPARAM lParam) -> BOOL {
+            auto pCallbackData = reinterpret_cast<decltype(&callbackData)>(lParam);
+            auto pCallback = pCallbackData->pCallback;
+            using result_t = decltype((*pCallback)(hwnd));
+#ifdef __cpp_if_constexpr
+            if constexpr (std::is_void_v<result_t>)
+            {
+                (*pCallback)(hwnd);
+                return TRUE;
+            }
+            else if constexpr (std::is_same_v<result_t, HRESULT>)
+            {
+                // NB: this works for both HRESULT and NTSTATUS as both S_OK and ERROR_SUCCESS are 0
+                return (S_OK == (*pCallback)(hwnd)) ? TRUE : FALSE;
+            }
+            else if constexpr (std::is_same_v<result_t, bool>)
+            {
+                return (*pCallback)(hwnd) ? TRUE : FALSE;
+            }
+            else
+            {
+                static_assert(details::always_false<TCallback>::value, "Callback must return void, bool, or HRESULT");
+            }
+#else
+            return (*pCallback)(hwnd);
+#endif
+        };
+        enumApi(enumproc, reinterpret_cast<LPARAM>(&callbackData));
+    }
+
+
+#ifdef WIL_ENABLE_EXCEPTIONS
     template <typename TEnumApi, typename TCallback>
     void CallCallback(TEnumApi&& enumApi, TCallback&& callback)
     {
@@ -73,18 +112,13 @@ namespace details
             std::rethrow_exception(callbackData.exception);
         }
     }
+#endif
 } // namespace details
 
 template <typename TCallback>
 auto for_each_window_nothrow(TCallback&& callback) noexcept
 {
-    details::CallCallback(&EnumWindows, wistd::forward<TCallback>(callback));
-}
-
-template <typename TCallback>
-auto for_each_window(TCallback&& callback)
-{
-    details::CallCallback(&EnumWindows, wistd::forward<TCallback>(callback));
+    details::CallCallbackNoThrow(&EnumWindows, wistd::forward<TCallback>(callback));
 }
 
 template <typename TCallback>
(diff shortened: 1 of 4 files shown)


include/wil/windowing.h
```diff
@@ -0,0 +1,78 @@
...
+            if constexpr (std::is_void_v<result_t>)
+            {
+                try
+                {
+                    (*pCallback)(data);
+                    return TRUE;
+                }
+                catch (...)
+                {
+                    return FALSE;
```
sylveon, line 36: I'm not sure silently gobbling up the exception is the right thing to do. Maybe let noexcept do the std::terminate.

tests/WindowingTests.cpp
```diff
@@ -0,0 +1,50 @@
...
+        {
+            if (pid == GetWindowThreadProcessId(hwnd, nullptr))
+            {
+                REQUIRE(IsWindow(hwnd));
+            };
+            return true;
+        });
+
+    const auto thread_id = GetCurrentThreadId();
+    wil::for_each_thread_window(thread_id, [](HWND hwnd)
```
jonwis, line 32: Can you add a test for a lambda that is mutable? Like:

```c++
std::vector<HWND> windows;
wil::for_each_window(thread_id, [&windows] (HWND wnd) { windows.push_back(wnd); });
```

And one that takes a mutable lambda:

```c++
wil::for_each_window(thread_id, [windows = std::vector<HWND>{}] (HWND wnd) mutable { windows.push_back(wnd); });
```

include/wil/windowing.h
sylveon: Should be possible with `std::exception_ptr`.

include/wil/windowing.h
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+
+        template<typename TEnumWindows, typename TCallback>
+        auto for_each_window(const TEnumWindows& enumWindows, const TCallback& callback) noexcept
+        {
+            return enumWindows(LambdaEnumWindowsProc<TCallback>, reinterpret_cast<LPARAM>(&callback));
+        }
+
+        using EnumThreadWindowsType = decltype(&EnumThreadWindows);
```
sylveon, line 56: Can you add EnumChildWindow?

include/wil/windowing.h
```diff
@@ -0,0 +1,78 @@
...
+                {
+                    (*pCallback)(data);
+                    return TRUE;
+                }
+                catch (...)
+                {
+                    return FALSE;
+                }
+            }
+            else if constexpr (std::is_same_v<result_t, HRESULT>)
```
dunhor, line 39: This makes me feel a little uneasy, but I'm not sure how legitimate that fear is. `HRESULT` is defined as `long` and `BOOL` is defined as `int` (if you're in `minwindef.h`; if you're in `wtypes.h` it is `long`, oops). That's walking a line that, on the surface, looks thin and it seems feasible that the return type of a lambda gets deduced incorrectly. An explicit return type would fix that (until you're getting the definition from `wtypes.h`), but who does that.

In practice... most `HRESULT` constants are defined so that their type is correctly casted as `long`/`HRESULT`. Additionally, `TRUE`/`FALSE` seem to be defined as just `1` and `0` respectively, meaning they will be `int`s and not `long`, regardless of how `BOOL` is defined. Additionally, if the user uses `true`/`false`, it'll take the "fallback" `else` below and will cast to `BOOL` correctly.

All this is to say, I think this is _probably_ okay, but it does give me pause. I can't think of any non-fabricated scenario that would cause issues (someone using `return 0` instead of `return S_OK` isn't a realistic fear in my eyes). Still, calling it out in case I'm overlooking something that someone else sees.

include/wil/windowing.h
```diff
@@ -0,0 +1,78 @@
...
+namespace wil
+{
+    namespace details
+    {
+        template<typename TCallback>
+        BOOL CALLBACK LambdaEnumWindowsProc(HWND data, LPARAM lParam) noexcept
+        {
+            const TCallback* pCallback = reinterpret_cast<const TCallback*>(lParam);
+            using result_t = decltype((*pCallback)(data));
+#ifdef __cpp_if_constexpr
```
dunhor, line 26:
```suggestion
#if __cpp_if_constexpr
```
or
```c++
#ifdef __cpp_if_constexpr >= 201606L
```
Major nit. Most all guidance or uses I've seen test against a version (or against 0 if you're testing presence at all)

dunhor, line 26:
```suggestion
#ifdef __cpp_if_constexpr
            using result_t = decltype((*pCallback)(data));
```
I believe that should resolve your build failure

tests/WindowingTests.cpp
dunhor, line 32: Question: when running the tests, does the callback get invoked at all? Maybe the console widow, but I'm somewhat doubtful of that. If not, might be good to use the thread id of something that's guaranteed to have windows (like Explorer's desktop maybe?)

include/wil/windowing.h
asklar: I'm not sure what you meant @dunhor.
If we want to check against a version, then `#if __cpp_if_constexpr >= 201606L` would be right. Here I was checking whether it's defined at all, which is equivalent as I'm not relying on post 201606 behavior. 
Saying `#ifdef __cpp_if_constexpr >= 201606L` seems wrong (the >= part should be ignored)

asklar:
@dunhor / @jonwis, any concerns with this iteration or can we merge?

include/wil/windowing.h
```diff
@@ -0,0 +1,164 @@
...
+        if (callbackData.exception)
+        {
+            std::rethrow_exception(callbackData.exception);
+        }
+    }
+#endif
+} // namespace details
+
+template <typename TCallback>
+auto for_each_window_nothrow(TCallback&& callback) noexcept
```
dunhor, line 113: All these have an `auto` return type, but don't actually `return` anything. If intentional, explicitly mark as `void`; gives off the impression that something was forgotten

include/wil/windowing.h
dunhor: Feature test macros, like `__cpp_if_constexpr`, are defined to values that indicate their "level" of support. There's two possible sources of concern:

The first is if new standards expand on the functionality. Let's hypothetically say that the first version of `__cpp_if_constexpr` did not allow braces for `if constexpr` (i.e. forcing it to be a single statement). In that case, the `#ifdef` check would pass for older compilers, but fail compilation when it encounters the `{`.

The second is for pre-release experimental features. E.g. it's theoretically possible for `__cpp_if_constexpr` to be set to some pre-release constant that correlates with a draft of the proposal that uses the syntax `constexpr if` instead of `if constexpr` (just an example; idk if such a draft or implementation exists). In that case, if a user is using an older compiler with the experimental feature enabled, the `#ifdef` check will pass, but compilation will fail because the syntax is incorrect.

Now, both of these are hypothetical scenarios, so I don't think that there's _actually_ a concern here - hence the nit. I'm just advocating for performing explicit version checks as the default.

> Saying `#ifdef __cpp_if_constexpr >= 201606L` seems wrong

Woops. I meant `#if __cpp_if_constexpr >= 201606L`

include/wil/windowing.h
```diff
@@ -0,0 +1,164 @@
...
+{
+namespace details
+{
+    template <typename T>
+    struct always_false : wistd::false_type
+    {
+    };
+
+    template <typename TEnumApi, typename TCallback>
+    void CallCallbackNoThrow(TEnumApi&& enumApi, TCallback&& callback) noexcept
```
dunhor, line 28: These functions are more specific than a generic "invoke callback" - suggest renaming to something like `EnumerateWindows`, `DoEnumWindows`, `PerformWindowEnumeration`, etc.

include/wil/windowing.h
```diff
@@ -0,0 +1,164 @@
...
+    {
+        auto enumproc = [](HWND hwnd, LPARAM lParam) -> BOOL {
+            auto pCallback = reinterpret_cast<TCallback*>(lParam);
+#ifdef __cpp_if_constexpr
+            using result_t = decltype((*pCallback)(hwnd));
+            if constexpr (std::is_void_v<result_t>)
```
dunhor, line 34:
```suggestion
            if constexpr (wistd::is_void_v<result_t>)
```

include/wil/windowing.h
```diff
@@ -0,0 +1,164 @@
...
+            {
+                (*pCallback)(hwnd);
+                return TRUE;
+            }
+            else if constexpr (std::is_same_v<result_t, HRESULT>)
```
dunhor, line 39:
```suggestion
            else if constexpr (wistd::is_same_v<result_t, HRESULT>)
```

include/wil/windowing.h
asklar: thanks, addressed

include/wil/windowing.h
asklar: changed to void

dunhor (approved):

PR #405, opened 2023-12-05
Declare default virtual destructors to quiet [-Wnon-virtual-dtor]

bashi-bazouk:
This change allows compiling ComTests.cpp and ResourceTests.cpp with -Werror -Wnon-virtual-dtor enabled.

### DIFF
--- a/tests/ComTests.cpp
+++ b/tests/ComTests.cpp
@@ -69,6 +69,7 @@ public:
         ReleaseCounter = 0;
         return res;
     }
+    virtual ~IUnknownFake() = default;
 protected:
     static int AddRefCounter;
     static int ReleaseCounter;
@@ -552,6 +553,7 @@ public:
     {
         return 0;
     }
+    virtual ~ExtensionHelper() = default;
 };
 
 TEST_CASE("ComTests::Test_ConstPointer", "[com][com_ptr]")
@@ -597,6 +599,7 @@ TEST_CASE("ComTests::Test_ComPtrWithForwardDeclaration", "[com][com_ptr]")
         {
             return 0;
         }
+        virtual ~MyClass() = default;
     };
 }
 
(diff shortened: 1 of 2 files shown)


dunhor:
I get that these types are just for unit tests, but these are all COM types where the destructor really shouldn't be part of the v-table because the type destroys itself in the `Release` method. If these warnings are causing issue, we should disable them either globally or locally at their definitions

ChrisGuzak (approved):

tests/ComTests.cpp
```diff
@@ -614,6 +617,7 @@ interface __declspec(uuid("ececcc6a-5193-4d14-b38e-ed1460c20a01"))
 IDerivedTest : public ITest
 {
    STDMETHOD_(void, TestDerived)() = 0;
+   virtual ~ITest() = default;
```
jonwis, line 620: This should be `~IDerivedTest` - you can't declare a destructor for a base type.

tests/ComTests.cpp
```diff
@@ -626,18 +630,21 @@ interface __declspec(uuid("ececcc6a-5193-4d14-b38e-ed1460c20a03"))
 IDerivedTestInspectable : public ITestInspectable
 {
    STDMETHOD_(void, TestInspctableDerived)() = 0;
+   virtual ~ITestInspectable() = default;
```
jonwis, line 633: Same as above; s/b `~IDerivedTestInspectable`

jonwis (changes requested):

jonwis:
In a earlier project, we kept finding leaks when someone had the pattern `class CIntermediate : IFooExtendsIUnknown` then `class CDerivedMost : CIntermediate` where the `CDerivedMost` overrode a method on `IFooExtendsIUnknown` and added additional state. The `delete this` in `CIntermediate::Release` didn't trigger `~CDerivedMost` because nobody had a `virtual ~Anything()` in the type hierarchy.  The super easy fix was to just add `virtual ~CIntermediate() { };` and all the leaks got closed.

So I'm sympathetic to the warning.  Should it be firing for types that contain only types with trivial destructors, though? Like the `FakeStream` type has no state to destroy and no members with nontrivial destructors.

So the two that need it in this change are the `IUnknownFake` (an instance with non-pure-virtual methods) and `FakeStream`. The others are just "pure interfaces."

PR #410, opened 2023-12-14
wil::com_task does not propagate thread-local restricted error information across thread boundaries

### LINKED ISSUE #408: wil::com_task does not propagate thread-local restricted error information across thread boundaries
I have a crash that I'm debugging and it has been difficult to debug because the UI thread doesn't have the original stowed exception information.  The first RoOriginateError call on the UI thread is in the `winrt::terminate` handler when we are already going down a fatal code path.  After trying lots of different approaches I finally thought to check for stowed exceptions on every thread in the process and sure enough one of the background threads has the "real" original exception.

The program in question is roughly as follows:
```cpp
wil::com_task<Foo> DoThingAsync()
{
  THROW_HR(E_FAIL); // oopsies
  co_return foo;
}

winrt::fire_and_forget UpdateUI()
{
  try
  {
    auto result = co_await DoThingAsync();
  }
  catch (winrt::hresult_error&)
  { /* handle */ }
}
```

My true bug is that we catch `winrt::hresult_error` but not `wil::ResultException` so the catch block doesn't catch and the exception hits the `winrt::fire_and_forget` noexcept boundary.  That crashes the process.  Now that I understand the problem it is easy enough to fix.

I am filing this issue because it was overly difficult to figure this out based on a crash dump.  I think the situation could be improved if `wil::com_task` were to call [GetRestrictedErrorInfo](https://learn.microsoft.com/en-us/windows/win32/api/roerrorapi/nf-roerrorapi-getrestrictederrorinfo) on the background thread and if it is non-null propagate it to the UI thread with a call to `SetRestrictedErrorInfo`.  The auto-originate would have then seen the original and let it remain active.  Failure analysis would have also pointed me far more directly at the true problem.

dmachaj:
Closes #408 

This change is being made because I was debugging a crash yesterday and felt that the blame was not helpful enough.  I figured out the problem relatively quickly but it took me almost 2 hours to get the debugger to tell me the same thing, which is way too long.  The problem was root caused to the new WIL coroutine library lacking COM restricted error propagation.

The error was thrown on one thread and then crashed the UI thread.  The detailed error origination was still orphaned on the random background thread so it was not available to blame when the crash occurred.  The failfast just had to make do with what it could scrounge up from the rethrown exception.

This set of changes adds a `restricted_error` member to `wil::details::coro::result_holder`.  The restricted error field will hold the `IRestrictedErrorInfo` if an exception is thrown by the coroutine.  When the exception propagates to the resume context the error will be restored to its rightful place on that thread.

This approach is complicated because `wil::task<>` and `wil::com_task<>` are convertible between each other.  The COM task functionality only lights up when COM headers are included before `<wil/coroutine.h>`.  However the `result_holder` is outside this lightup block so extra steps needed to be taken to reconcile these details.

The extra step is that `restricted_error` is saved as a `void*`.  There are a series of function pointers that are able to Retrieve/Restore/Destroy the error information.  Those function pointers are null by default.  If COM is available then they get filled in by functions that can use COM types.  When the pointers are null they are simply not called.

I wrote a new test case that ensures that the COM per-thread error information contains the expected error that was thrown by a background thread.  I also applied this private coroutine.h to my primary project and re-tested the crash from yesterday.  The blame in the debugger is now much more accurate and useful than it was before.


dmachaj:
Things I'd like to add to this PR:
# Another test case where <wil/coroutine.h> is used without COM headers available.
# Write a small sample program that demonstrates the problem and include the before/after debugger blame.

I am sending this without those items because they are (IMO) non-critical and I wanted to begin the review process sooner.

include/wil/coroutine.h
```diff
@@ -646,9 +683,27 @@ namespace wil::details::coro
 #define __WIL_COROUTINE_NON_AGILE_INCLUDED
 #include <ctxtcall.h>
 #include <wil/com.h>
+#include <roerrorapi.h>
```
dunhor, line 686: Is this generally applicable if COM headers are included? I've always associated the restricted error info more with WinRT than with COM. Just curious if it makes sense to factor this out into its own `ifdef` if `roerrorapi.h` has already been included (assuming a unique include guard exists for it).

dunhor (approved):
I like this approach of using the function pointers. Avoids the possible ODR violations that are becoming far too common around here.

include/wil/coroutine.h
dmachaj: I factored it out into a separate bock with suitable include guards but it is disabled in our app because it seems that <roerrorapi.h> is not included by anything else that we use.

That separation would vastly limit the usefulness of this error propagation because none of the VS project templates would include it by default.  But even if it isn't included COM is still providing restricted error information so it really needs to be propagated.

In case it is interesting I see that WRL headers include it so there is definitely precedent for including this header broadly.

dunhor: WRL was mostly created for the Windows Runtime (although you can of course use it to consume/produce non-WinRT types), so not necessarily a precedent setter here.

That said, I don't have many strong feelings here. I think it's ultimately fine either way.

dmachaj:
I wrote a little test program to demonstrate the difference between the before and after.  Here's the test program (which began using the C++/WinRT console app project template).

```cpp
#include <wil/result_originate.h>
#include <wil/winrt.h>
#include <wil/cppwinrt.h>
#include <wil/coroutine.h>

wil::com_task<bool> CrashOnBackground()
{
    THROW_HR(E_APPLICATION_TEMPORARY_LICENSE_ERROR);
    co_return true;
}

winrt::fire_and_forget ForegroundMethod(HANDLE e)
{
    // Signal the incoming event handle when the coroutine has completed.
    auto complete = wil::SetEvent_scope_exit(e);

    co_await CrashOnBackground();
}

int main()
{
    init_apartment(winrt::apartment_type::single_threaded);

    auto done = wil::shared_event(wil::unique_event(wil::EventOptions::ManualReset));
    auto handle = done.get();

    ForegroundMethod(handle);

    DWORD waitResult;
    while ((waitResult = MsgWaitForMultipleObjects(1, &handle, false, INFINITE, QS_ALLEVENTS)) == WAIT_OBJECT_0 + 1)
    {
        MSG msg;
        while (PeekMessage(&msg, nullptr, 0, 0, PM_REMOVE))
        {
            TranslateMessage(&msg);
            DispatchMessage(&msg);
        }
    }
}
```

The debugger blame before this change is as follows:
```
FAULTING_SOURCE_CODE:  
    17:     // Signal the incoming event handle when the coroutine has completed.
    18:     auto complete = wil::SetEvent_scope_exit(e);
    19: 
    20:     co_await CrashOnBackground();
>   21: }
    22: 
    23: int main()
    24: {
    25:     init_apartment(winrt::apartment_type::single_threaded);
    26: 

STACK_TEXT:  
000001bf`b9793e00 00007ffa`38c2030d combase!RoOriginateErrorW+0x4d
000001bf`b9793e08 00007ff6`3002b374 WilTaskRestrictedError!wil::details::RaiseRoOriginateOnWilExceptions+0x214
000001bf`b9793e10 00007ff6`3002a362 WilTaskRestrictedError!wil::details::LogFailure+0x462
000001bf`b9793e18 00007ff6`3001a8ea WilTaskRestrictedError!wil::details::ReportFailure_Return<1>+0xfa
000001bf`b9793e20 00007ff6`30019257 WilTaskRestrictedError!wil::details::ReportFailure_Base<1,0>+0x87
000001bf`b9793e28 00007ff6`300198d2 WilTaskRestrictedError!wil::details::ReportFailure_CaughtExceptionCommon<1>+0x2a2
000001bf`b9793e30 00007ff6`3001959f WilTaskRestrictedError!wil::details::ReportFailure_CaughtException<1>+0xcf
000001bf`b9793e38 00007ff6`30035f24 WilTaskRestrictedError!wil::winrt_to_hresult+0xd4
000001bf`b9793e40 00007ff6`30035220 WilTaskRestrictedError!winrt::to_hresult+0x60
000001bf`b9793e48 00007ff6`30034a36 WilTaskRestrictedError!winrt::terminate+0x56
000001bf`b9793e50 00007ff6`300359a4 WilTaskRestrictedError!std::coroutine_traits<winrt::fire_and_forget,void *>::promise_type::unhandled_exception+0x24
000001bf`b9793e58 00007ff6`3003ea95 WilTaskRestrictedError!`ForegroundMethod$_ResumeCoro$1'::`1'::catch$3+0x65
000001bf`b9793e60 00007ffa`303f1080 VCRUNTIME140_1D!_CallSettingFrame_LookupContinuationIndex+0x20
000001bf`b9793e68 00007ffa`303f3aae VCRUNTIME140_1D!__FrameHandler4::CxxCallCatchBlock+0x1de
000001bf`b9793e70 00007ffa`39eb3c66 ntdll!RcFrameConsolidation+0x6
000001bf`b9793e78 00007ff6`30026ac4 WilTaskRestrictedError!ForegroundMethod$_ResumeCoro$1+0x564
000001bf`b9793e80 00007ff6`30026491 WilTaskRestrictedError!ForegroundMethod$_InitCoro$2+0x121
000001bf`b9793e88 00007ff6`3002720f WilTaskRestrictedError!ForegroundMethod+0xcf
000001bf`b9793e90 00007ff6`300363db WilTaskRestrictedError!main+0x9b
000001bf`b9793e98 00007ff6`30038a49 WilTaskRestrictedError!invoke_main+0x39
000001bf`b9793ea0 00007ff6`3003892e WilTaskRestrictedError!__scrt_common_main_seh+0x12e
000001bf`b9793ea8 00007ff6`300387ee WilTaskRestrictedError!__scrt_common_main+0xe
000001bf`b9793eb0 00007ff6`30038ade WilTaskRestrictedError!mainCRTStartup+0xe
000001bf`b9793eb8 00007ffa`3821257d KERNEL32!BaseThreadInitThunk+0x1d
000001bf`b9793ec0 00007ffa`39e6aa58 ntdll!RtlUserThreadStart+0x28
```

The debugger blame after this change is:
```
FAULTING_SOURCE_CODE:  
     7: 
     8: wil::com_task<bool> CrashOnBackground()
     9: {
    10:     co_await winrt::resume_background();
>   11:     THROW_HR(E_APPLICATION_TEMPORARY_LICENSE_ERROR);
    12:     co_return true;
    13: }
    14: 
    15: winrt::fire_and_forget ForegroundMethod(HANDLE e)
    16: {

STACK_TEXT:  
000002ab`ed9d99c0 00007ffa`38c2030d combase!RoOriginateErrorW+0x4d
000002ab`ed9d99c8 00007ff7`84aab534 WilTaskRestrictedError!wil::details::RaiseRoOriginateOnWilExceptions+0x214
000002ab`ed9d99d0 00007ff7`84aaa522 WilTaskRestrictedError!wil::details::LogFailure+0x462
000002ab`ed9d99d8 00007ff7`84a9a63b WilTaskRestrictedError!wil::details::ReportFailure_NoReturn<0>+0x13b
000002ab`ed9d99e0 00007ff7`84a99517 WilTaskRestrictedError!wil::details::ReportFailure_Base<0,0>+0x87
000002ab`ed9d99e8 00007ff7`84a9a1b9 WilTaskRestrictedError!wil::details::ReportFailure_Hr<0>+0xa9
000002ab`ed9d99f0 00007ff7`84aae02f WilTaskRestrictedError!wil::details::in1diag5::Throw_Hr+0x6f
000002ab`ed9d99f8 00007ff7`84aa519f WilTaskRestrictedError!CrashOnBackground$_ResumeCoro$1+0x3af
000002ab`ed9d9a00 00007ff7`84aa3c01 WilTaskRestrictedError!std::coroutine_handle<void>::operator+0x51
000002ab`ed9d9a08 00007ff7`84ab439f WilTaskRestrictedError!winrt::impl::resume_background_callback+0x3f
000002ab`ed9d9a10 00007ffa`39e5e4b3 ntdll!TppSimplepExecuteCallback+0xa3
000002ab`ed9d9a18 00007ffa`39e45986 ntdll!TppWorkerThread+0x8f6
000002ab`ed9d9a20 00007ffa`3821257d KERNEL32!BaseThreadInitThunk+0x1d
000002ab`ed9d9a28 00007ffa`39e6aa58 ntdll!RtlUserThreadStart+0x28
```

The output after this change is definitely a lot more useful because it points at the source of the problem instead of where it is first observed by the crashing thread.

dunhor:
Looks like the pipeline got disabled yesterday. I've re-enabled it and once I re-remember how to manually run for PRs I'll kick it off. Otherwise pushing new changes should trigger it

dmachaj:
> Looks like the pipeline got disabled yesterday. I've re-enabled it and once I re-remember how to manually run for PRs I'll kick it off. Otherwise pushing new changes should trigger it

Nice timing, then.  I just finished getting my new test case to compile and was pushing anyway.

include/wil/coroutine.h
```diff
@@ -161,6 +161,7 @@
 
 #include <atomic>
 #include <exception>
+#include <utility>
```
dmachaj, line 164: This was a preexisting problem.  This file uses `std::exchange` but did not include `<utility>`.

dunhor:
> > Looks like the pipeline got disabled yesterday. I've re-enabled it and once I re-remember how to manually run for PRs I'll kick it off. Otherwise pushing new changes should trigger it
> 
> Nice timing, then. I just finished getting my new test case to compile and was pushing anyway.

Seems you forgot to include the file `CoroutineTests.cpp` ;)

dmachaj:
> > > Looks like the pipeline got disabled yesterday. I've re-enabled it and once I re-remember how to manually run for PRs I'll kick it off. Otherwise pushing new changes should trigger it
> > 
> > 
> > Nice timing, then. I just finished getting my new test case to compile and was pushing anyway.
> 
> Seems you forgot to include the file `CoroutineTests.cpp` ;)

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
        ../../tests/CoroutineTests.cpp
```

🤦 totally wasted the build happening over my lunch break

include/wil/coroutine.h
```diff
@@ -151,7 +151,7 @@
 #if defined(_RESUMABLE_FUNCTIONS_SUPPORTED)
```
dunhor, line 151: Not your change, and I'm not necessarily going to suggest touching it now, but this feels backwards. I'd think we'd want to check for the standard impl first, then fall back to experimental

dunhor (approved):
Going to wait for this build to succeed and then merge it before merging the clang-format change

include/wil/coroutine.h
dmachaj: That's a good point. @oldnewthing do you recall if there was a reason for this ordering or is it arbitrary?

PR #412, opened 2023-12-15
Add a .clang-format file to WIL

dunhor:
Previous draft PR was here: https://github.com/microsoft/wil/pull/401, which has a few comments. I've created a new one as that one had a bit of experimentation and got a little messy. The biggest topic of conversation from before was the outer-most namespace contents not getting tabbed causing a lot of whitespace changes. For now the plan is to roll with it, however that may make merge conflicts a little tricky in the short term.

**NOTE: The `.clang-format` file has been take from the OS repo with very minimal additions to ensure that some of WIL's macros don't cause weird formatting. This is _not_ a request for comment on personal style preferences. Comments and feedback should be focused on issues with the formatting (e.g. macros being handled incorrectly, etc.).**

For now, all I've done is add the `.clang-format` (with additions mentioned above), add a script that runs `clang-format` over all our code with the version of `clang-format` that ships with Visual Studio (as an attempt to get some consistency), and then run said script. No manual code changes have been made (I'll call out any I end up making). A few things I plan to look into for the future:
* PR validation that either fails if `clang-format` was not run, or just runs `clang-format` and submits the result. For now, we can ask people to run `clang-format` if it's obvious that they didn't, but otherwise just run it ourselves periodically (likely on an OS sync)l
* A script to just format lines that have changed (from my understanding there's some degree of git integration with `clang-format`)

A few additional things I plan to address/look into:
* C++/CX "hat" (`^`) pointers get formatted like they are XOR operations. I plan to do a global search for `__cplusplus_winrt` and wrap them with `clang-format off`/`clang-format on` and fix the pointer formatting. That way the code is still formatted initially, but will require manual intervention in the future
* Exclude the `workarounds` directory from auto-formatting
* Some very long `template <...>` lines seem to not be getting wrapped (though others are for some reason). I plan to investigate those
* The following macros/non-standard compiler extensions seem to sometimes cause issues; I plan to look at these, but not quite sure what/how much I can fix:
  * Use of `__R*` macros in `result_macros.h` sometimes cause line breaks, sometimes not
  * Some function definitions that have `RESULT_NORETURN`/`__declspec(noinline)` sometimes get formatted weirdly
  * Use of SAL annotations seem to be correlated with some inconsistent formatting
  * Macros that specify function names (e.g. something like `#define FUNCTION(ret, name) ...`) sometimes gets formatted weirdly

A few additional notes/questions:
* `WI_NOEXCEPT` sometimes causes interesting formatting side effects. In particular, on constructors that contain initializers and functions with trailing return types (e.g. `auto fn() WI_NOEXCEPT -> int`) where the `->` gets formatted like a pointer access. There's a way to tell `clang-format` that this is a macro for `noexcept` which fixes the formatting, however it's relatively new and the version of `clang-format` that ships with Visual Studio does not include it.
* There's a decent amount of macro churn in both `result_macros.h` and `tracelogging.h` that doesn't necessarily make the code easier to read/understand. Are there any places where we want to disable formatting in those files?

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -17,8 +17,8 @@
 #include "wistd_memory.h"
 
 #pragma warning(push)
-#pragma warning(disable:26135 26110)    // Missing locking annotation, Caller failing to hold lock
-#pragma warning(disable:4714)           // __forceinline not honored
+#pragma warning(disable : 26135 26110) // Missing locking annotation, Caller failing to hold lock
+#pragma warning(disable : 4714)        // __forceinline not honored
 
 #ifndef __WIL_RESOURCE
 #define __WIL_RESOURCE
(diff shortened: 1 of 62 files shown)


ChrisGuzak (approved):
I don't have strong opinions about the questions you asked in the PR description. do what you think is best.

include/wil/resource.h
```diff
@@ -27,7326 +27,7575 @@
...
-            }
+// unique_any provides a template alias for easily building a unique_any_t from a unique_storage class with the given
+// template parameters for resource_policy.
+
+template <
+    typename pointer,                                      // The handle type
+    typename close_fn_t,                                   // The handle close function type
+    close_fn_t close_fn,                                   //      * and function pointer
+    typename pointer_access = details::pointer_access_all, // all, noaddress or none to control pointer method access
+    typename pointer_storage = pointer,          // The type used to store the handle (usually the same as the handle itself)
```
dunhor, line 521: Despite `AlignTrailingComments: true`, I can't seem to figure out a way to keep the comments aligned. Not urgent or pressing; I just feel like it should work

dunhor:
> I don't have strong opinions about the questions you asked in the PR description. do what you think is best.

I found some annotations that were missing. Otherwise, couldn't figure out any non-hacky way to get the rest to work. The `Macros` vector seems to work decently well, but isn't available until version 17, which is higher than the version that ships with VS. I'll delay those for now. I'll leave everything else as-is. We can go back and re/un-format if we get too annoyed with it later

PR #414, opened 2023-12-19
Fixing the name field when enumerating registry keys and values

### LINKED ISSUE #413: registry enumeration has a bug with the key name field size and m_name_length getting out of sync.
This shows a bug in the intermediate representations of the iterator. Note that passing by value works, by creating a copy and re-syncing the std::wstring size. 

it seems when the size of the value name gets smaller is when this bug occurs.

it seems `m_name_length` is for the bstr case and I don't have much sympathy for code that needs the value name in that format. can we simplify this and get rid of m_name_length or not use a std::wstring in this case to avoid the duplication and incoherence it creates?

```cpp
        // failing test. post this issue on github
        TEST_METHOD(VerifyRegistryEnumeratorWorksWithByRefParameter)
        {
            PCWSTR subKey = LR"(BugRepro)";
            auto extensionsIds = wil::reg::create_unique_key(HKEY_CURRENT_USER_LOCAL_SETTINGS, subKey, wil::reg::key_access::readwrite);
            wil::reg::set_value_string(extensionsIds.get(), L"{00021401-0000-0000-C000-000000000046}", L"");
            wil::reg::set_value_string(extensionsIds.get(), L"{66742402-F9B9-11D1-A202-0000F81FEDEE}", L"");
            wil::reg::set_value_string(extensionsIds.get(), L"Windows.Foundation.Uri", L"");
            wil::reg::set_value_string(extensionsIds.get(), L"Windows.Foundation.PropertyValue", L"");
            wil::reg::set_value_string(extensionsIds.get(), L"Wrong", L"{F8F24682-0E3B-4BE3-B6F5-1EC33AFE207D}");

            auto begin = wil::reg::value_iterator{ extensionsIds.get() };
            auto end = wil::reg::value_iterator{};

            std::for_each(begin, end, [&](auto nameAndType) // this works
            {
                cpp_unit::Logger::WriteMessage(nameAndType.name.c_str());
                auto size = wcslen(nameAndType.name.c_str());
                cpp_unit::Assert::IsTrue(size == nameAndType.name.size());
            });

            std::for_each(begin, end, [&](auto&& nameAndType) // this fails with the size assert below
            {
                cpp_unit::Logger::WriteMessage(nameAndType.name.c_str());
                auto size = wcslen(nameAndType.name.c_str());
                cpp_unit::Assert::IsTrue(size == nameAndType.name.size()); // failing
            });
        }
```

keith-horton:
When using std::wstring as the name type when enumerating registry names and values, we need to fix the internal consistency of the std::wstring object + have it correctly sync with the name_size property in the parent object.

Fixes #413

### DIFF
--- a/include/wil/registry_helpers.h
+++ b/include/wil/registry_helpers.h
@@ -1293,6 +1293,8 @@ namespace reg
     {
         constexpr uint32_t iterator_end_offset = 0xffffffff;
         constexpr size_t iterator_default_buffer_length = 32;
+        constexpr size_t iterator_max_keyname_length = 255;
+        constexpr size_t iterator_max_valuename_length = 16383;
 
         // function overloads to allow *_enumerator objects to be constructed from all 3 types of HKEY representatives
         inline HKEY get_hkey(HKEY h) WI_NOEXCEPT
@@ -1557,8 +1559,20 @@ namespace reg
                 }
                 if (error == ERROR_MORE_DATA)
                 {
-                    // resize by iterator_default_buffer_length and try again
-                    string_length += ::wil::reg::reg_iterator_details::iterator_default_buffer_length;
+                    // if our default wasn't big enough, resize to either half max or max length and try again
+                    if (string_length < reg_iterator_details::iterator_max_keyname_length / 2)
+                    {
+                        string_length = reg_iterator_details::iterator_max_keyname_length / 2;
+                    }
+                    else if (string_length < reg_iterator_details::iterator_max_keyname_length + 1)
+                    {
+                        string_length = reg_iterator_details::iterator_max_keyname_length + 1; // plus one for null
+                    }
+                    else
+                    {
+                        // if we're already at max length, we can't grow anymore, so we'll just return the error
+                        break;
+                    }
                     continue;
                 }
                 // any other error will fail
@@ -1680,8 +1694,18 @@ namespace reg
                 }
                 if (error == ERROR_MORE_DATA)
                 {
-                    // resize by iterator_default_buffer_length and try again
-                    string_length += ::wil::reg::reg_iterator_details::iterator_default_buffer_length;
+                    if (string_length == ::wil::reg::reg_iterator_details::iterator_max_valuename_length + 1)
+                    {
+                        // this is the max size, so we can't grow anymore, so we'll just return the error
+                        break;
+                    }
+
+                    // resize and try again - growing exponentially up to the max
+                    string_length *= 2;
+                    if (string_length > ::wil::reg::reg_iterator_details::iterator_max_valuename_length + 1)
+                    {
+                        string_length = ::wil::reg::reg_iterator_details::iterator_max_valuename_length + 1;
+                    }
                     continue;
                 }
 
--- a/include/wil/registry_helpers.h
+++ b/include/wil/registry_helpers.h
@@ -28,1081 +28,1084 @@
 
 namespace wil
 {
-namespace reg
-{
-    /**
-     * @brief Helper function to translate registry return values if the value was not found
-     * @param hr HRESULT to test from registry APIs
-     * @return boolean if the HRESULT indicates the registry value was not found
-     */
-    constexpr bool is_registry_not_found(HRESULT hr) WI_NOEXCEPT
+    namespace reg
     {
-        return (hr == __HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND)) || (hr == __HRESULT_FROM_WIN32(ERROR_PATH_NOT_FOUND));
-    }
-
-    /**
-     * @brief Helper function to translate registry return values if the buffer was too small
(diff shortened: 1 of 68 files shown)


include/wil/registry_helpers.h
```diff
@@ -28,1081 +28,1084 @@
...
-        // Open key for reading.
-        read,
-
-        // Open key for reading and writing. Equivalent to KEY_ALL_ACCESS.
-        readwrite,
-    };
-
-    /// @cond
-    namespace reg_view_details
+    namespace reg
```
dunhor, line 31: Please run `clang-format` on this change

dunhor:
Fixes #413

dunhor:
This should add the test Chris described in #413 for current & future validation

include/wil/registry_helpers.h
```diff
@@ -1617,7 +1644,11 @@ namespace reg
                 if (error == ERROR_SUCCESS)
                 {
                     // some types, like std::wstring, cannot have embedded nulls
-                    ::wil::reg::reg_view_details::reg_value_type_info::trim_buffer(name);
+                    // if we trimmed anything, we must also get the new sizej
```
dunhor, line 1647:
```suggestion
                    // if we trimmed anything, we must also get the new size
```

include/wil/registry_helpers.h
```diff
@@ -1367,6 +1370,27 @@ namespace reg
...
+                {
+                    return 0;
+                }
+                return new_length;
+            }
+
+            // continue to use the existing buffer since the requested length is less than or equals to the current length
+            clear_name(name, current_length);
+            return current_length;
+        }
```
dunhor, line 1392:
```suggestion
#if defined(WIL_ENABLE_EXCEPTIONS) && defined(_STRING_)
        // failure returns zero
        inline size_t resize_name(::std::wstring& name, size_t current_length, size_t new_length) WI_NOEXCEPT
        {
            // because std::wstring encodes the size within its object, we must resize whenever the 2 are not equal
            // not just when the new_length is greater than the current_length -- which is correct for types that don't encode their size
            if (new_length != current_length)
            {
                // resize_buffer takes size in bytes
                if (FAILED(::wil::reg::reg_view_details::reg_value_type_info::resize_buffer(
                        name, static_cast<DWORD>(new_length * sizeof(wchar_t)))))
                {
                    return 0;
                }
                return new_length;
            }

            // continue to use the existing buffer since the requested length is less than or equals to the current length
            clear_name(name, current_length);
            return current_length;
        }
#endif
```
Or move it up to a pre-existing guard. This should fix your build failure

ChrisGuzak:
out-side of the scope of this change... 
- I see the default value of a key is not considered value as `is_valid()` for the `std::wstring` case uses `!empty()`
- growing the buffer size only by `iterator_default_buffer_length` (16) does not work well if the names are large. this should use an exponential growth strategy.
- it is a bummer we have to allocate the value names. value names are 16k or so, but key names < 255 (or so).
- using a bstr as a temporary representation is very odd. LocalAlloc or other allocators are more conventional. 
- most clients are better off with a **fixed sized** "max value name" constant. 
  - this could be a template param for testing and letting callers pay for what they need. 
  - we could have another "dynamic" mode, also a template param with `if constexpr` enabled code that does the allocations for those that must have it.

keith-horton:
> This should add the test Chris described in #413 for current & future validation

Yep, added a set of new tests that this helped uncover.

keith-horton:
> out-side of the scope of this change...
> 
> * using `BSTR` as the key name is odd.
>   
>   * **LocalSize could be used to determine the size, avoiding the need for the separate size book-keeping needed because of `BSTR` that is the cause of this bug.**
>   * LocalAlloc or other allocators are more conventional. our guidelines say, "don't use BSTRs in new code".
> * it is a bummer we have to allocate the value names. value names are 16k or so, but key names < 255 (or so).
> * most clients are better off with a **fixed sized** "max value name" constant.
>   
>   * this could be a template param for testing and letting callers pay for what they need.
>   * we could have another "dynamic" mode, also a template param with `if constexpr` enabled code that does the allocations for those that must have it.
> * The default value of a key (that with an empty string for a name) may have incorrect behavior as `is_valid()` for the `std::wstring` case uses `!empty()` (code below)
> * growing the buffer size only by `iterator_default_buffer_length` (16) does not work well if the names are large. an exponential growth strategy seems better.
> 
> What do you think about the idea of switching to using `LocalAlloc` and `LocalSize` to get rid of `m_current_length`?
> 
> ```c++
>             inline bool is_valid(const ::std::wstring& name) WI_NOEXCEPT
>             {
>                 return !name.empty();
>             }
> ```

Lots of good feedback.

I've addressed some of it in this most recent push. Will be thinking about some of the other feedback.

ChrisGuzak (approved):
great response to my comments Keith! much improved.

PR #415, opened 2023-12-19
Fix & Parallelize the CI pipelines

dunhor:
There's a few things here:
1. Some time ago, likely around when I needed to switch to running tests in the architecture-specific VS command prompt, we effectively stopped running the x64 tests because the pipeline has a copy/paste error where it runs the _x86_ vcvars script. This has been fixed here.
2. I've updated the pipeline to upgrade the version of LLVM installed to 17. This is because the build machines already have version 16 installed, so we've been running validation on an old version. The reason I explicitly specified a version is so that the pipeline doesn't randomly start failing for folks in the future for unrelated reasons.
3. With the switch to Clang version 17, the ASan tests have started failing for x64. I've noticed this locally for quite some time, however I haven't dug too deep into it because the CI seems to have still been working. Disabling them for now until I have adequate time to look into it.
4. I have parallelized the pipeline so that each individual architecture+compiler+build type runs in parallel. This should cut our CI validation times from like ~2.5 hours to <20 min or so. cc @dmachaj who will be happy about this.

### DIFF
--- a/scripts/azure-pipelines.yml
+++ b/scripts/azure-pipelines.yml
@@ -51,6 +51,7 @@ jobs:
       if %ERRORLEVEL% NEQ 0 goto :eof
       echo ##vso[task.setvariable variable=PATH]%PATH%;C:\Program Files\LLVM\bin
     displayName: 'Install Clang'
+    condition: eq($(compiler), 'clang')
 
   - script: |
       call scripts\call-vcvars.cmd $(arch)
@@ -64,5 +65,6 @@ jobs:
 
   - script: |
       call scripts\call-vcvars.cmd $(arch)
+      set ASAN_WIN_CONTINUE_ON_INTERCEPTION_FAILURE=1
       call scripts\runtests.cmd ~[LocalOnly]
-    displayName: 'Run $(compiler) $(arch)$(build-type) Tests'
+    displayName: 'Test $(compiler) $(arch)$(build-type)'
--- a/scripts/azure-pipelines.yml
+++ b/scripts/azure-pipelines.yml
@@ -16,11 +16,11 @@ jobs:
     matrix:
       ${{each compiler in variables.compileres}}:
         ${{each arch in variables.architectures}}:
-          ${{each build-type in variables.buildTypes}}:
-            ${{compiler}}-${{arch}}-${{build-type}}:
+          ${{each buildType in variables.buildTypes}}:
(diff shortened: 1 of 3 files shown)


scripts/azure-pipelines.yml
```diff
@@ -7,50 +7,64 @@ jobs:
 - job: BuildAndTest
   timeoutInMinutes: 360
 
+  strategy:
```
dunhor, line 10: If anyone knows how to do this in a nicer way, let me know. E.g. it's a lot of copy/paste to add a new "value" and a LOT more copy paste to add a new "variable".

dmachaj: Can you do a `matrix` like cppwinrt?  https://github.com/microsoft/cppwinrt/blob/2511bf7fcb54a64c1e596a950788e2ee30cd1e4c/.pipelines/build.yml#L29

dmachaj: It appears that matrix supports nesting: https://learn.microsoft.com/en-us/azure/devops/pipelines/yaml-schema/jobs-job-strategy?view=azure-pipelines

dmachaj:
~20 minutes instead of 2.5 hours is a __huge__ improvement.  Thanks so much for implementing this.

scripts/azure-pipelines.yml
dunhor: Missed the comment about `each` on my first skim through. Not too happy with the `split`, however the docs say that variables must be strings, so arrays don't work :/

dmachaj (approved):

dmachaj:
Also nice - this will eliminate the risk of the "only run x86 tests" bug that you fixed in this PR.  If you build x64 but try to run x86 tests then it will fail because that build agent won't have the x86 tests to run.

dunhor:
> If you build x64 but try to run x86 tests then it will fail because that build agent won't have the x86 tests to run.

Not quite; it'll just silently pass. I actually had to modify the pipeline to delete x86 output a while ago because we were exhausting disk space (something else this change will help with), so we're currently in a state where running the x64 tests just pass after 0-ish seconds elapsed.

That said, the script to run the proper vcvars script should make it a lot harder to regress this since it'll be harder to introduce a similar copy/paste error.

ChrisGuzak (approved):

PR #416, opened 2023-12-21
Add PR validation for formatting

dunhor:
This change:
* Adds a script to run `git clang-format` to format only changes against the master branch to reduce noise in different `clang-format` versions
* Runs this script in CI, failing if code hasn't been formatted (I've added a formatting violation to test for now)
* Also added the `format` target so that it's easier to run `clang-format` over all sources

An example pipeline run that failed on this check: https://dev.azure.com/msft-wil/Windows%20Implementation%20Library/_build/results?buildId=1045&view=results

One thing we might want to discuss/vote on: instead of failing the CI, might it make sense to just commit the results of `clang-format` into the target branch? Personally, I tend to lean towards _not_ automating pushing changes, however it would save us a lot of "please run clang-format" comments on PRs...

### DIFF
--- a/scripts/format-changes.cmd
+++ b/scripts/format-changes.cmd
@@ -1,0 +1,19 @@
+@echo off
+setlocal
+setlocal EnableDelayedExpansion
+
+set PROJECT_ROOT=%~dp0\..
+
+set BRANCH=%1
+if "%BRANCH%"=="" (
+    echo ERROR: Missing commit/branch argument. If this is a fork of the microsoft/wil repo, you likely want to specify
+    echo        'upstream/master'. If this is not a fork, you likely want to specify 'origin/master'. Examples:
+    echo.
+    echo            format-changes.cmd origin/master
+    echo            format-changes.cmd upstream/master
+    exit /b 1
+)
+
+pushd %PROJECT_ROOT% > NUL
+git clang-format %BRANCH% -- include/wil/*.h tests/*.h tests/*.cpp
+popd > NUL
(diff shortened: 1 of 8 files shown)


scripts/format-changes.cmd
```diff
@@ -0,0 +1,22 @@
...
+setlocal EnableDelayedExpansion
+
+set PROJECT_ROOT=%~dp0\..
+
+set BRANCH=%1
+if "%BRANCH%"=="" (
+    echo ERROR: Missing commit/branch argument. If this is a fork of the microsoft/wil repo, you likely want to specify
+    echo        'upstream/master'. If this is not a fork, you likely want to specify 'origin/master'. Examples:
+    echo.
+    echo            format-changes.cmd origin/master
```
dunhor, line 12: On forks, I know that this is problematic if you've never run `git fetch upstream`, however the failure should hopefully be relatively obvious.

For both forks and non-forks this could potentially be problematic if you've run `git fetch upstream`/`git fetch origin`, but haven't merged with your local branch. Or similarly, if you've merged with master on a different machine and pulled only those changes without also fetching changes from master. I believe that doing so will end up diffing more than just the changes you've made in your branch. I don't believe that would necessarily be problematic (formatting more code shouldn't really affect much), but it might with mismatching `clang-format` versions. This should not be an issue for the CI check.

ChrisGuzak (approved):
don't feel strongly about "auto-format in the CI".

ChrisGuzak:
you mentioned `git clang-format`, how does that work?

dunhor:
> you mentioned `git clang-format`, how does that work?

I realized that may not have been entirely clear, so I'm updating the readme with some info

scripts/format-changes.cmd
dunhor: Hopefully the readme is enough to diagnose any possible misuses

PR #421, opened 2024-01-02
Consume Detours for function mocking

### LINKED ISSUE #393: CoRegisterMallocSpy causes unit tests to fail periodically
Discovered in #359, tests that use or validate helpers for CoRegisterMallocSpy cause downstream breaks for correct working code.

This tracks following one of the solutions proposed by @dunhor in https://github.com/microsoft/wil/issues/359#issuecomment-1786455431

> There's a crash in the tests caused by a bug in COM that's caused by a combination of using `IMallocSpy` and not cleaning up C++/WinRT's `CoIncrementMTAUsage`. Raymond and I have chatted about this a bit offline (as it turns out, this is not the first test to fail because of it), and I've floated a few ideas:
>
> 1. Simplest thing is to manually call `CoInitialize` as I've suggested above. This bypasses >?>C++/WinRT's logic & we don't hit the issue because we've properly cleaned up COM before process exit
> 2. Another simple thing to do is to disable the test, but obviously not ideal
> 3. Another idea I've floated is to tag the tests that use `IMallocSpy` specially and then perform two passes of test invocation: one that excludes the `IMallocSpy`-using tests and one that only runs those tests
>
> I'm personally a fan of option 1: it's simple and effective. Option 3 would help us avoid these issues moving forward, however.

cc @oldnewthing as well

dunhor:
A little background - internal to the Windows repo there's a library for mocking functions for the purpose of writing tests. As of this writing, this library has not yet been made public. The original tests for WIL (before being made open source) relied on this library for a number of things. When porting the tests to this repo, the tests that rely on this library either needed to be modified to use alternative means for testing the same functionality or omitted entirely. This led to a number of hacks to try and achieve the same functionality, some of which have historically caused issues.

This change is an attempt to replace the use of that library with something that uses the Detours library to achieve something similar (it's noteworthy that the internal library _also_ uses Detours). This code isn't intended to be as feature rich as the internal library and is targeted at providing just what's necessary for our tests. In particular, there are two types used for mocking:
* `detoured_thread_function` - detours a function, however it only invokes the callback if the call came in on the thread that created the object. This is to avoid unintended consequences of some other thread interfering with the test - e.g. because of running tests in parallel, random chance, etc.
* `detoured_global_function` - detours a function that calls the function object regardless of which thread called the detoured function. This is intended as a last resort option when the thread making the call cannot be predicted - e.g. from a thread pool thread, etc.

Both of these types are designed to be able to be constructed with a lambda since testing function pointers + globals is often a mess. Additionally, both are designed such that calls are non-reentrant on the thread that called them. This is nice because it allows the detour to just invoke the function it's detouring by name as opposed to needing to do something like reference the object holding it, however this also means that some scenarios may not work 100% as expected (e.g. your detour => detoured function => other function... => detoured function).

I've updated relevant tests to use these types as opposed to the work arounds we've been using & added some tests that were never copied in the first place. Now, this change isn't without some negatives:

1. In order to use Detours, the tests now have a dependency on `vcpkg`. This is an extra setup step - albeit one most people will only need to do once - however it does increase the "barrier to entry" slightly. Aside - I also took the opportunity to switch to consuming Catch2 via vcpkg while I was at it. _NOTE: I'm unsure if this may have consequences for code consuming WIL from vcpkg_.
2. I bumped the minimum C++ standard version that the tests build against to C++17 because it was a huge pain to write the detoured function types without inline static variables. I don't think this will be a major issue - C++17 has been out for quite some time now and the Windows build switched to C++17 a while ago, however it's noteworthy that we no longer have build coverage.

Ultimately, these tradeoffs seem worth it to me; just calling them out to see what others think.

Fixes #393


include/wil/com_apartment_variable.h
```diff
@@ -217,11 +217,13 @@ namespace details
         // get current value or throw if no value has been set
         std::any& get_existing()
         {
-            if (auto any = get_if())
+            auto any = get_if();
+            if (!any)
             {
-                return *any;
+                THROW_HR(E_NOT_SET);
```
dunhor, line 223: Note that changes like this are a consequence of defining `RESULT_NORETURN` to nothing. I debated trying to change it so that only the failfast functions are marked with `RESULT_NORETURN` since those are the only ones that we detour, however this is simple enough

tests/StlTests.cpp
```diff
@@ -50,6 +51,42 @@ TEST_CASE("StlTests::TestSecureAllocator", "[stl][secure_allocator]")
     }
 }
 
+// MockMethod<void(std::allocator<T>::*)(T*, size_t)>
```
dunhor, line 54: I was debating adding support for mocking member functions, but never got around to it. I don't think it would be _too_ hard, but we already have decent enough validation of the `secure_deleter` type that I'm going to punt it for now.

dunhor:
Note that an alternative to adding a dependency on vcpkg is to instead check in & build Detours ourselves. Personally, I'm not a huge fan, but...

ChrisGuzak (approved):
wow... great work Duncan.

vcpkg.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": [
+    "detours",
+    "catch2"
```
dunhor, line 4: I'm going to delay doing this for now, but we might want to consider specifying explicit versions so we don't unexpectedly break. E.g. the current version of `catch2` is what they call "v3" which had some breaking changes I needed to fix...

include/wil/registry.h
```diff
@@ -2157,12 +2157,12 @@ namespace reg
             return {value};
```
jonwis, line 2157: While you're here, care to make this a `return std::move(value)` ? IIRC, `std::optional<T>` has a ctor for `T&&`.

tests/CoroutineTests.cpp
```diff
@@ -3,7 +3,9 @@
 // this project.  The best we can do is to define WIN32_LEAN_AND_MEAN to prevent windows.h
 // from including <ole2.h> (which in turn includes <combasepi.h>).
 //
-// We also avoid including "common.h" from the current directory because it pulls in COM too.
+// We also avoid including "common.h" from the current directory because it pulls in COM too.#include "pch.h"
```
jonwis, line 6: Comment got mangled? Has trailing `#include`

vcpkg.json
jonwis: I'm a fan of explicit version dependencies, at least to major/minor level.

jonwis (approved):

tests/CppWinRTAuthoringTests.cpp
```diff
@@ -226,73 +228,57 @@ TEST_CASE("CppWinRTAuthoringTests::EventsAndCppWinRt", "[property]")
...
 
-// This test seems to have issues on the CI machines, so enabling only when run locally
-TEST_CASE("CppWinRTAuthoringTests::NotifyPropertyChanged", "[property][LocalOnly]")
+TEST_CASE("CppWinRTAuthoringTests::NotifyPropertyChanged", "[property]")
 {
 #if defined(WIL_ENABLE_EXCEPTIONS)
-    auto uninit = wil::RoInitialize_failfast(RO_INIT_MULTITHREADED);
+    auto uninit = wil::RoInitialize_failfast(RO_INIT_SINGLETHREADED);
     // We need to initialize the XAML core in order to instantiate a PropertyChangedEventArgs.
-    // Do all the work on a separate DispatcherQueue thread so we can shut it down cleanly
```
asklar, line 235: bad merge? looks like it might be missing raymond's change to drain the queue?

tests/wiTest.cpp
```diff
@@ -1768,11 +1780,21 @@ TEST_CASE("WindowsInternalTests::HandleWrappers", "[resource][unique_any]")
...
 
     {
-        auto verify = MakeSecureDeleterMallocSpy();
-        REQUIRE_SUCCEEDED(::CoRegisterMallocSpy(verify.Get()));
-        auto removeSpy = wil::scope_exit([&] {
-            ::CoRevokeMallocSpy();
-        });
+        witest::detoured_thread_function<&::LocalFree> detour;
+        REQUIRE_SUCCEEDED(detour.reset([](HLOCAL p) -> HLOCAL {
+            HLOCAL h = nullptr;
```
asklar, line 1785: `wil::unique_hlocal`?

tests/wiTest.cpp
```diff
@@ -1689,11 +1688,24 @@ TEST_CASE("WindowsInternalTests::HandleWrappers", "[resource][unique_any]")
...
-        auto verify = MakeSecureDeleterMallocSpy();
-        REQUIRE_SUCCEEDED(::CoRegisterMallocSpy(verify.Get()));
-        auto removeSpy = wil::scope_exit([&] {
-            ::CoRevokeMallocSpy();
-        });
+        witest::detoured_thread_function<&::CoTaskMemFree> detour;
+        REQUIRE_SUCCEEDED(detour.reset([](void* p) {
+            if (p != nullptr)
+            {
+                IMalloc* pMalloc = nullptr;
```
asklar, line 1695: `wil::com_ptr<IMalloc>`?

tests/CppWinRTAuthoringTests.cpp
dunhor: This intentionally reverted that change as it's no longer necessary since we're no longer using `IMallocSpy`. This also means that this test no longer sporadically fails and can therefore be enabled to run on CI machines.

tests/CoroutineTests.cpp
dunhor: What on earth... Looks like I copy/pasted twice and didn't realize :/

include/wil/registry.h
dunhor: Looks like this is a lesson on why brace construction is not always a good idea. `return value;` works as you'd expect & avoids the copy since `value` is an x-value. `return {value};` does _not_ avoid the copy since it's not an x-value in this context.

Example: https://godbolt.org/z/hxd4EfKWe

tests/wiTest.cpp
dunhor: Not sure that helps much here. `h` is the result of the call to `LocalFree` and is returned to the caller. We also couldn't use it for wrapping `p` since we need the result of the call to `LocalFree` to return to the caller.

tests/wiTest.cpp
dunhor: Fair. Probably shouldn't have blindly copy pasted code without giving it a once-over

tests/CppWinRTAuthoringTests.cpp
asklar: IIUC the xaml manager Close is async so technically you still need to drain the queue (it might happen to not crash, but it's a bug to not drain the queue)

dunhor: Fair point. I guess I can keep the dispatcher queue & thread, but remove the wait

asklar (approved):

vcpkg.json
dunhor: Okay, I _think_ I understand a bit more about how vcpkg handles versioning. From what I can gather, it doesn't really follow the format that something like npm uses and I'm unsure if it's possible to declare a package to use a specific major version (e.g. force using `3.*`). That said, you can specify an explicit baseline - which I've added now - and from my understanding, that should use the package versions listed in that baseline. So, while we're not specifying explicit versions for each package, we should still get consistency and reliability if I'm understanding things correctly.

tests/CppWinRTAuthoringTests.cpp
dunhor: Hmm. For some reason even doing it this way causes issues when running on the CI machines. Need to take a deeper look at it...

dunhor: Oh my. The DispatcherQueue seems to intentionally crash if its DispatcherQueueController "parent" gets destroyed (destructor runs) when shutdown starts, and looks like it will unintentionally crash if the controller gets destroyed when shutdown completes. My guess is that the CI machines are hitting the latter scenario as it's reporting an AV, not a fail fast exception. From what I can gather, this should be safe if we wait for `ShutdownQueueAsync` to complete before completing the test.

dunhor: Well, it wasn't very smart of me to look at the _current_ implementation when the CI machines are running Windows Server 2022. From what I can gather from the build number, this correlates to fe_release. I created a VM with the same build (albeit Desktop, not Server) and was able to repro this under TTD. Looks like `InputHost.dll` is getting unloaded before the thread pool tries to invoke a callback into that dll on thread termination. I'm unsure if this is something that has been fixed since then. I've never hit this locally, so maybe it has, however it's tricky to say since the code in question doesn't exist anymore (though it's likely it's just been moved elsewhere). That said, the operation being scheduled is a `Release` call, so hopefully if it _has_ been moved, wherever it was moved to is using something like WRL and has proper module ref count tracking.

In the end, it looks like Raymond's change to wait for thread termination is probably the most reasonable way to work around this bug on both CI machines and older builds. I've increased the timeout so that hopefully the potato CI machines don't hit as many false positives.

asklar: ooof. well, xaml islands is fun like that. 
@jessecol does this ring any bells?

asklar:
hooray :)

dunhor:
[celebrate]     Duncan Horn reacted to your message:
________________________________

tests/CppWinRTAuthoringTests.cpp
JesseCol: Hi!  It doesn't ring any bells, I've found that usually closing all the WindowsXamlManagers/DeskopWindowXamlSources and draining message queues on the threads using Xaml is enough for a clean shutdown.  Sadly, I can't recall what objects live in InputHost.dll or speculate intelligently about why it's getting unloaded.

dunhor: The function that registers the TP Wait is `InputHost!WRL2ContextSession::RegisterThreadExitCallback`. The CI machines seem to be running Windows Server 2022 build 10.0.20348, which AFAICT is a build number that correlates to `fe_release`, so pretty old; predates Windows 11. Wouldn't surprise me if this has been fixed - either intentionally or unintentionally because that code no longer seems to exist - and this hasn't really been seen/reported.

PR #428, opened 2024-01-23
Adding macro RETURN_IF_FAILED_SUPPRESS_TELEMETRY_IF_EXPECTED

### LINKED ISSUE #425: Return failed HRESULT, log failure but don't send telemetry
For the result macros, when one sets a telemetry fallback provider with SetResultTelemetryFallback() and a logging callback with SetResultLoggingCallback(), we always log and send telemetry for HRESULT failures in the RETURN_ macros.
The _EXPECTED macros allows one to skip the reporting for some expected HRESULTs, but that suppresses both logging and telemetry.

Is there a way to suppress only telemetry for expected scenarios, but still get logging?

The scenario is when you don't want the telemetry to have expected errors, but they still help in debugging when looking at a local log.

araujoarthur0:
Closes #425.

For the result macros, when one sets a telemetry fallback provider with SetResultTelemetryFallback() and a logging callback with SetResultLoggingCallback(), we always log and send telemetry for HRESULT failures in the RETURN_ macros.
The _EXPECTED macros allows one to skip the reporting for some expected HRESULTs, but that suppresses both logging and telemetry. There was no alternative if one wanted to always log failures but just suppress telemetry in case the HRESULT was expected.

The new macro `RETURN_IF_FAILED_SUPPRESS_TELEMETRY_IF_EXPECTED` returns an HRESULT if it failed, always logging those failures, but sends a suppressed telemetry callback if the HRESULT matches one of the expected HRESULT values.
The way this is done is leveraging the existing enum `FailureFlags::SuppressTelemetry`. This enum makes the telemetry callback be [called](https://github.com/microsoft/wil/blob/master/include/wil/result.h#L1069) with `alreadyReported` as `true`.

There is a new parameter `FailureFlags` to the internal ReturnHr methods in result_macros.h.

New tests were added to check the feature behavior.

### DIFF
--- a/include/wil/result_macros.h
+++ b/include/wil/result_macros.h
@@ -757,6 +757,14 @@ WI_ODR_PRAGMA("WIL_FreeMemory", "0")
         return __hr; \
     } \
     __WI_SUPPRESS_4127_E while ((void)0, 0)
+#define __RETURN_HR_FAIL_SUPPRESS_TELEMETRY(hr, str) \
+    __WI_SUPPRESS_4127_S do \
+    { \
+        const HRESULT __hr = (hr); \
+        __R_FN(Return_HrSuppressTelemetry)(__R_INFO(str) __hr); \
+        return __hr; \
+    } \
+    __WI_SUPPRESS_4127_E while ((void)0, 0)
 #define __RETURN_HR_FAIL_NOFILE(hr, str) \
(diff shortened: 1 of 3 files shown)


araujoarthur0:

ChrisGuzak (approved):

dunhor (approved):

PR #430, opened 2024-01-26
Fix comments in filesystem.h

marc2k3:
Fixes return value comment for try_open_file.

Also, some missing braces on other structured bindings have been corrected.


ChrisGuzak (approved):
thanks for the fixes

Issue #436, opened 2024-02-06
Feature request: Add functions to enumerate Toolhelp32Snapshot


HO-COOH:
Found myself often needed to enumerate modules/threads of a process, something you do with `CreateToolhelp32Snapshot`. Its current usage is a bit chunky with needing to first call `xxxFirst` followed with `xxxNext`. Everytime I just roll up a helper function to make the code more readable. As a result, a generic helper function for it would be great.

PR #438, opened 2024-02-20
Add Toolhelp32Snapshot helpers

HO-COOH:
As described in #436


include/wil/toolhelp32.h
```diff
@@ -0,0 +1,133 @@
+#include "resource.h"
```
dunhor, line 1: Why is `resource.h` included before include guard?

include/wil/toolhelp32.h
```diff
@@ -0,0 +1,133 @@
...
+    }
+} // namespace details
+
+template <typename TCallback>
+void for_each_process(TCallback&& callback)
+{
+    PROCESSENTRY32 entry{};
+    entry.dwSize = sizeof(entry);
+    details::do_enum_snapshot(
+        unique_handle{CreateToolhelp32Snapshot(TH32CS_SNAPPROCESS, 0)}.get(),
```
dunhor, line 48: All of these functions - `CreateToolhelp32Snapshot`, `Process32First`, etc. - can fail. Ideally the API would preserve & report failures to the caller

include/wil/toolhelp32.h
```diff
@@ -0,0 +1,133 @@
...
+        entry,
+        &Process32First,
+        &Process32Next,
+        wistd::forward<TCallback>(callback));
+}
+
+template <typename TCallback>
+void for_each_thread(TCallback&& callback)
```
dunhor, line 56: This name is confusing. On its own, I would expect it to enumerate the threads in the _current_ process, however that is not what this is doing.

include/wil/toolhelp32.h
```diff
@@ -0,0 +1,133 @@
...
+    details::do_enum_snapshot(
+        unique_handle{CreateToolhelp32Snapshot(TH32CS_SNAPTHREAD, 0)}.get(),
+        entry,
+        &Thread32First,
+        &Thread32Next,
+        wistd::forward<TCallback>(callback));
+}
+
+template <typename TCallback>
+void for_each_module(TCallback&& callback, bool include32For64Bit = false)
```
dunhor, line 69: Having function arguments after callback arguments is typically very awkward to call. To get around the issue of not being able to specify default arguments that way, you can overload the function.

include/wil/toolhelp32.h
```diff
@@ -0,0 +1,133 @@
...
+    details::do_enum_snapshot(
+        unique_handle{CreateToolhelp32Snapshot(TH32CS_SNAPHEAPLIST, 0)}.get(),
+        entry,
+        &Heap32ListFirst,
+        &Heap32ListNext,
+        wistd::forward<TCallback>(callback));
+}
+
+template <typename TCallback>
+void for_each_heap(TCallback&& callback, ULONG_PTR heapId, DWORD pid = GetCurrentProcessId())
```
dunhor, line 95: Same comment about function arguments after the callback argument

include/wil/toolhelp32.h
```diff
@@ -0,0 +1,133 @@
...
+                    return false;
+                }(),
+                "Callback must return void or bool");
+        }
+        enumResult = Heap32Next(&entry);
+    } while (enumResult);
+}
+
+template <typename TCallback>
+void for_each_heap(TCallback&& callback, HEAPLIST32 const& heapList, DWORD pid = GetCurrentProcessId())
```
dunhor, line 127: Same function argument ordering comment here

include/wil/toolhelp32.h
HO-COOH: @dunhor Agreed. Would early returning `false` in case of anything fails (something like this) be sufficient?
```cpp
template<...>
bool for_each_process(...)
{
    unique_handle h{Create...()};
    if (!h) return false;
    //...
}
```

include/wil/toolhelp32.h
HO-COOH: - `for_each_thread_global` for all system threads
- `for_each_thread_of_process(callback)` for current thread
- `for_each_thread_of_process(pid, callback)` for threads of specified process ?

include/wil/toolhelp32.h
```diff
@@ -0,0 +1,133 @@
+#include "resource.h"
+#ifndef __WIL_TOOLHELP32_INCLUDED
+#define __WIL_TOOLHELP32_INCLUDED
+#include <TlHelp32.h>
+#include <processthreadsapi.h>
+namespace wil
```
dunhor, line 6:
```suggestion
//*********************************************************
//
//    Copyright (c) Microsoft. All rights reserved.
//    This code is licensed under the MIT License.
//    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF
//    ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
//    TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
//    PARTICULAR PURPOSE AND NONINFRINGEMENT.
//
//*********************************************************
#ifndef __WIL_TOOLHELP32_INCLUDED
#define __WIL_TOOLHELP32_INCLUDED

#include <TlHelp32.h>
#include <processthreadsapi.h>

#include "resource.h"

namespace wil
```

include/wil/toolhelp32.h
dunhor: WIL is pretty heavily standardized on using `HRESULT` as the error type communicated back to callers. The general pattern is to have three functions: a base `_nothrow` function, a `_failfast` version that fails fast on error, and an exception throwing function. So in this case, you'd basically have the three functions:
```c++
HRESULT for_each_process_nothrow(...) { ... }
void for_each_process_failfast(...) { FAIL_FAST_IF_FAILED(for_each_process_nothrow(...); }
#if WIL_ENABLE_EXCEPTION
void for_each_process(...) { THROW_IF_FAILED(for_each_process_nothrow(...); }
#endif
```
Given that your function accepts an arbitrary lambda from the user that might throw, `for_each_process_nothrow` probably shouldn't be made `noexcept`, making the function name a little misleading, but IMO that's probably fine.

include/wil/toolhelp32.h
dunhor: > Not creative enough for names

Me too... Only suggestions I can come up with at the moment are `for_each_system_thread` and `for_each_process_thread` that has an overload with a PID (otherwise assumes current process)

PR #440, opened 2024-02-20
Add helpers for implementing Out-Of-Proc COM Server with cppwinrt only (without WRL)

### LINKED ISSUE #439: Helpers for implementing Out Of Proc COM server in cppwinrt, without WRL
Context:
https://github.com/microsoft/cppwinrt/issues/601

This comment summarize the request in a succinct way:
image

For brevity, here are the original requirement as described in the above thread:
1. Simple object registration
2. Automatic outstanding object ref counting
3. Synchronized COM shutdown and event when all objects have disconnected.

As analyzed by kenny, (2) is already handled by cppwinrt. Some extra work is needed for (1) and (3).

I have created a [sample repo](https://github.com/roxk/NotifiableModuleLock) that demonstrates that both (1) and (3) is possible with just cppwinrt, with the help of some small helpers:

|ask|helper|
|--|--|
| simple object registration | register function with variadic template |
| synchronized COM shutdown | notifiable module lock |

You can read the sample repo for more details. In short, here is how it works:
1. cppwinrt exposes a configuration WINRT_CUSTOM_MODULE_LOCK that allows users to hook into the module lock mechanism
2. Write a custom module lock which, (a) notifies when module count reaches 0, and (b) provides customization point for the notification
3. cppwinrt also provides winrt::no_module_lock so that class factory implementation can be omitted from module object count
4. Write variadic functions that take in authored type and register class factory recursively

This issue asks that these helpers be added to wil for mass adoption so that cppwinrt-only Out Of Proc COM server becomes a reality. Once these helpers get into wil, `IExplorerCommand` and `IWidgetProvider` sample should be updated so that they are cppwinrt-only. Say goodbye to WRL for good.

---

#### Notifiable module lock
```
namespace wil
{
    template <typename Func>
    struct notifiable_module_lock
    {
        // Other ref count details omitted...
        static void set_notifier(Func& func)
        {
            notifier = func;
        }
    private:
        static inline Func notifier;
    };
}
```

#### Object registration
```
// NOTE: this has to be in another header file, or in the same file guarded by some MACRO
// This is because the module lock has to be defined _before_ including winrt/base.h
// but object registration support requires the use of winrt::implements, which means it has to be
// defined _after_ including winrt/base.h
namespace wil::details
{
  template <typename T>
  struct Factory : winrt::implements<Factory<T>, IClassFactory, winrt::no_module_lock> {}
}

namespace wil
{
// RAII COM server registration revoker
struct com_server_revoker
{
    std::vector<DWORD> registrations;
    ~com_server_revoker()
    {
        for (auto&& registration : registrations)
        {
            winrt::check_hresult(CoRevokeClassObject(registration));
        }
    }
};

template <typename T, typename... Rest>
[[nodiscard]] com_server_revoker register_com_server();
}
```

#### Usage
```
// pch.h
#pragma once
// Include the module lock. Filename TBD
#include <wil/cppwinrt_notifiable_module_lock.h>
// Enable cppwinrt custom module lock support by defining this MACRO
#define WINRT_CUSTOM_MODULE_LOCK
// Define the custom module lock
namespace winrt
{
  inline auto& get_module_lock()
  {
    static wil::notifiable_module_lock<void(*)()> lock.
    return lock;
  }
}
// Include winrt _after_ configuring module lock
#include <unkwn.h>
#include <winrt/base.h>
// Include com registration helper
#include <wil/cppwinrt_register_com_server.h>

// main.cpp
wil::unique_event _comExitEvent;

void notifier()
{
  _comExitEvent.SetEvent();
}

int main()
{
  _comExitEvent.create();
  wil::notifiable_module_lock<void(*)()>::set_notifier(notifier);
  winrt::init_apartment();
  auto revoker = wil::register_com_server<MyClass>();
  _comExitEvent.wait();
  return 0;
}
```

### Open Question

1. Should wil include a default implementation of the boilerplate of defining the `WINRT_CUSTOM_MODULE_LOCK` MACRO and actually defining the custom module lock in `winrt` namespace? I have tested this idea and found that wil needs to provide another API to configure the module lock's notifier
2. Should notifiable module lock and object registration be in the same header file? I personally prefer separating them into two headers, but I'm aware wil like to group functionality into one header file and use macro to light up functionality

roxk:
Fixes #439

- [x] ~~Figure out how to run test. Test should involve spawning client that create instance in another process and the server should be able to reach after com exit event's wait~~ Found that create_instance with `CLSCTX_LOCAL_SERVER` in the same process/in another thread of the same process also works. Questions to reviewers: Does it guarantee out-of-process creation would also work?
- [x] Add test for failure mode
  - [x] `com_register_server`


include/wil/cppwinrt_notifiable_module_lock.h
```diff
@@ -0,0 +1,103 @@
...
+
+#ifndef __WIL_CPPWINRT_NOTIFIABLE_MODULE_LOCK_DEFINED
+#define __WIL_CPPWINRT_NOTIFIABLE_MODULE_LOCK_DEFINED
+
+#if _MSVC_LANG >= 201703L
+
+#include <atomic>
+#include <cstdint>
+
+// Ligh up cppwinrt custom module lock
```
dunhor, line 22:
```suggestion
// Light up cppwinrt custom module lock
```

include/wil/cppwinrt_notifiable_module_lock.h
```diff
@@ -0,0 +1,103 @@
...
+//    PARTICULAR PURPOSE AND NONINFRINGEMENT.
+//
+//*********************************************************
+//! @file
+//! Utilities for implementing OOP COM server with cppwinrt only
+
+#ifndef __WIL_CPPWINRT_NOTIFIABLE_MODULE_LOCK_DEFINED
+#define __WIL_CPPWINRT_NOTIFIABLE_MODULE_LOCK_DEFINED
+
+#if _MSVC_LANG >= 201703L
```
dunhor, line 17: Seems unnecessary. Isn't C++/WinRT's minimum required standard _at least_ C++17 (if not higher now)?

include/wil/cppwinrt_notifiable_module_lock.h
```diff
@@ -0,0 +1,103 @@
...
+
+#include <atomic>
+#include <cstdint>
+
+// Ligh up cppwinrt custom module lock
+#define WINRT_CUSTOM_MODULE_LOCK
```
dunhor, line 23: What is the effect if this header is included after `<cppwinrt/base.h>`? What is the effect if some translation units include this header, but others don't?

For the first question, an `#ifdef` w/ `#error` would likely add value if that silently fails. I'm not sure we can do anything about the second; C++/WinRT would likely need to be the one with an ODR check.

include/wil/cppwinrt_notifiable_module_lock.h
```diff
@@ -0,0 +1,103 @@
...
+        notifier = func;
+    }
+
+private:
+    std::atomic<int32_t> m_count;
+    static inline Func notifier;
+};
+} // namespace wil
+
+#ifndef WIL_CPPWINRT_COM_SERVER_CUSTOM_MODULE_LOCK
```
dunhor, line 80: If you remove the template argument from the above, this doesn't seem like it would be necessary. Nobody would ever need to customize the type of the callback object type, so there's no downside to providing these implementations; including the header is already enough.

include/wil/cppwinrt_notifiable_module_lock.h
dunhor: Maybe I'm answering my own question, but hopefully the linker would complain about the two different versions of `winrt::get_module_lock`. Hopefully...

include/wil/cppwinrt_notifiable_module_lock.h
```diff
@@ -0,0 +1,103 @@
...
+
+    operator uint32_t() const noexcept
+    {
+        return m_count;
+    }
+
+    static void set_notifier(Func func)
+    {
+        notifier = func;
+    }
```
dunhor, line 72:
```suggestion
    template <typename Callback>
    static void set_notifier(Callback&& func)
    {
        notifier = std::forward<Callback>(func);
    }
```

include/wil/cppwinrt_notifiable_module_lock.h
dunhor: Same comment on *com_server file

sylveon: Yes

include/wil/cppwinrt_register_com_server.h
```diff
@@ -0,0 +1,90 @@
...
+
+#if _MSVC_LANG >= 201703L
+
+#include <winrt/base.h>
+#include <vector>
+
+namespace wil::details
+{
+template <typename T>
+struct Factory : winrt::implements<Factory<T>, IClassFactory, winrt::no_module_lock>
```
dunhor, line 25:
```suggestion
struct CppWinRTClassFactory : winrt::implements<Factory<T>, IClassFactory, winrt::no_module_lock>
```

include/wil/cppwinrt_register_com_server.h
```diff
@@ -0,0 +1,90 @@
...
+    ~com_server_revoker()
+    {
+        revoke();
+    }
+    void revoke()
+    {
+        for (auto&& registration : registrations)
+        {
+            winrt::check_hresult(CoRevokeClassObject(registration));
+        }
```
dunhor, line 74:
```suggestion
        }
        registrations.clear();
```
Or make `revoke` private if nobody is supposed to call it

include/wil/cppwinrt_register_com_server.h
```diff
@@ -0,0 +1,90 @@
...
+
+template <>
+void register_com_server<void>(std::vector<DWORD>&)
+{
+}
+} // namespace wil::details
+
+namespace wil
+{
+struct com_server_revoker
```
dunhor, line 62: Delete copy constructor & copy assignment operator

include/wil/cppwinrt_register_com_server.h
```diff
@@ -0,0 +1,90 @@
...
+    }
+};
+
+template <typename T, typename... Rest>
+[[nodiscard]] com_server_revoker register_com_server()
+{
+    std::vector<DWORD> registrations;
+    registrations.reserve(sizeof...(Rest) + 1);
+    details::register_com_server<T, Rest...>(registrations);
+    return {std::move(registrations)};
```
dunhor, line 84:
```suggestion
    return registrations;
```
It's an x-value

include/wil/cppwinrt_register_com_server.h
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+};
+
+template <typename T, typename... Rest>
+[[nodiscard]] com_server_revoker register_com_server()
+{
+    std::vector<DWORD> registrations;
+    registrations.reserve(sizeof...(Rest) + 1);
+    details::register_com_server<T, Rest...>(registrations);
```
dunhor, line 83: If this throws in the middle of registration, you'll leak all previous registrations

include/wil/cppwinrt_register_com_server.h
```diff
@@ -0,0 +1,90 @@
...
+    {
+        return S_OK;
+    }
+};
+
+template <typename T = void, typename... Rest>
+void register_com_server(std::vector<DWORD>& registrations)
+{
+    auto& back = registrations.emplace_back();
+    winrt::check_hresult(
```
dunhor, line 49: If this throws, you'll have a bogus value at the end of the vector. Note that the value will be `0` - I'm unsure if the system treats that as an invalid value that's okay to blindly revoke or not

include/wil/cppwinrt_notifiable_module_lock.h
```diff
@@ -0,0 +1,103 @@
...
+#include <atomic>
+#include <cstdint>
+
+// Ligh up cppwinrt custom module lock
+#define WINRT_CUSTOM_MODULE_LOCK
+
+namespace wil
+{
+// Adopted from cppwinrt
+template <typename Func>
```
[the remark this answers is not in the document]
dunhor: Comment now that I see your test - I don't think much value is added by allowing functions with a non-`void` return type. Simplifying here seems much better overall.

tests/CppWinRTComServerTests.cpp
```diff
@@ -0,0 +1,102 @@
+#include "pch.h"
+
+#include "common.h"
+#undef GetCurrentTime
+#if _MSVC_LANG >= 201703L
```
dunhor, line 5:
```suggestion
```
We control the compilation here

tests/CppWinRTComServerCustomModuleLockTests.cpp
```diff
@@ -0,0 +1,65 @@
+#include "pch.h"
```
dunhor, line 1: This file & corresponding test executable can be deleted if we de-templatize the notifier bype

include/wil/cppwinrt_notifiable_module_lock.h
roxk: This is to guard against running test on non-C++/17 mode. I saw other tests including this macro 🧐

include/wil/cppwinrt_notifiable_module_lock.h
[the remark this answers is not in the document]
roxk: This is just following the footstep of `WRL`'s, `Module`, which allows the callback to be generic. I already omitted the class method variant. I personally don't mind forcing the callback to be just `void callback()`, but I'm not sure if that would break any use case. Might need to ping other WRL users for insight here 🙏

include/wil/cppwinrt_notifiable_module_lock.h
roxk: Right, wanted to use universal reference but forgot. Will update or out right remove once we decided on whether generic notifiable module lock is necessary.

include/wil/cppwinrt_register_com_server.h
roxk: In my original design it was in the dtor, but I reckon there might be corner cases where users would want to revoke earlier/manually, just for the sake of API completeness. I thought of using a bool to make `Revoke` work only once but clearing registration tokens is better.

include/wil/cppwinrt_register_com_server.h
roxk: Does WRL's [Module::Registers](https://learn.microsoft.com/en-us/cpp/cppcx/wrl/module-class?view=msvc-170#registerobjects) fail-fast for any error, or just skip classes that failed to register? I'd like to imitate the behavior here as the ask is "WRL but in cppwinrt speak".

include/wil/cppwinrt_notifiable_module_lock.h
sylveon: Suggest adding `#ifdef WINRT_BASE_H` as a more explicit guard/error against this case. Unfortunately I don't think it can catch everything (for example TUs which include `winrt/base.h` but never `wil/cppwinrt_notifiable_module_lock.h`), but hopefully people making use of this functionality define `WINRT_CUSTOM_MODULE_LOCK` at the project level.

What if we simply ommitted defining `WINRT_CUSTOM_MODULE_LOCK` to encourage users to define it at the project level?

include/wil/cppwinrt_notifiable_module_lock.h
dunhor: This header file (and subsequent test file(s)) are standalone and intended strictly for C++/WinRT. It's therefore reasonable to assume necessary levels of compiler support. If, say, you wanted to include this header from something like `resource.h`, then you would need guards, however even then it's probably more appropriate to guard the inclusion.

include/wil/cppwinrt_notifiable_module_lock.h
dunhor: > If you define another module lock after including `winrt/base.h`, the macro itself has no effect as the whole `winrt/base.h` is guarded by `#pragma once`.

Sorry, maybe I wasn't clear. From what I can gather, this header must be included _before_ any C++/WinRT headers and must be included in any translation unit that also includes any C++/WinRT headers. I wasn't trying to imply that including `<cppwinrt/base.h>` multiple times was somehow problematic or desired.

> Subsequent redefinition of `winrt::get_module_lock` get various kind of compile/linker error.

I would definitely expect that including `<cppwinrt/base.h>` before this header would result in compilation errors due to the conflicting definitions of `cppwinrt::get_module_lock` in the same TU. It's probably a nasty non-obvious compilation error that's likely not simple to diagnose, but at least that should fail consistently. Do you also get a failure from the linker error if another TU includes C++/WinRT, but not this header? You'd still have two conflicting definitions, however it's my understanding that the linker is not required to emit a diagnostic here (though if `link.exe` currently does that makes me feel a little better).

> What if we simply ommitted defining `WINRT_CUSTOM_MODULE_LOCK` to encourage users to define it at the project level?

Pending whether or not the linker fails for this scenario, that might be an option. We'd want to add something like:
```c++
#ifndef WINRT_CUSTOM_MODULE_LOCK
#error You must define 'WINRT_CUSTOM_MODULE_LOCK' at the project level
#endif
```
Though it would still be great if C++/WinRT had its own ODR check here

roxk: > What if we simply ommitted defining WINRT_CUSTOM_MODULE_LOCK to encourage users to define it at the project level?

That works. Current implementation is merely a "least resistance path" that does everything for you. I did think of making it opt-in. That is to say, define another macro, say, `WIL_USE_DEFAULT_NOTIFIABLE_MODULE_LOCK` which wire everything up.

I'm fine with both opt-out or opt-in. But perhaps reviewers/MS folks would prefer opt-in?

include/wil/cppwinrt_notifiable_module_lock.h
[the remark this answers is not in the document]
roxk: I think I can reference how wil's `scoped_exit` handle this case.

include/wil/cppwinrt_register_com_server.h
roxk: Updated. Follow WRL to undo all registered classes upon any failures. Class revocation errors are ignored as per WRL.

include/wil/cppwinrt_notifiable_module_lock.h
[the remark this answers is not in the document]
roxk: Even `scope_exit` uses template🧐I bet the intention is to allow efficient storage of all callable - is `scope_exit` is avoiding std::function intentionally?

roxk:
I found re-registering the same class doesn't trigger any error. Is there any trivial ways to make `CoRegisterClassObject` throw error without using e.g. invalid combination of flags? The doc says `CoRegisterClassObject` would only ever return invalid arguments, out of memory, or unexpected. How to trigger unexpected in a test?

include/wil/cppwinrt_notifiable_module_lock.h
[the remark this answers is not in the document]
sylveon: scope_exit is more performance sensitive, it is used as a cheap wrapper for RAII where no other suitable one exists. As such it can show up in tight loops and I would imagine ends up inlined most of the time. Allocating a function and preventing inlining via type erasure could be a problem.

This is in contrast to our server here, which allocates a std::function once at startup (which is already an expensive operation) and then only calls it during shutdown. Not very performance sensitive.

Arguably, there are no concerns about code size bloat since you should probably only have 1 server in a single exe, but maintainability is a concern and templates generally result in suboptimal IDE experience.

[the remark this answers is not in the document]
dunhor: I agree with everything @sylveon says but will add one more thing: complexity. Not just the complexity of definition & maintenance, but the complexity in use. If I wanted to use something like a function object (which cannot be casted to a function pointer), I would need to dig around to figure out what additional compile definitions I need, what needs to be defined, etc. This is made even worse if I want to use a stateful lambda (with a non-empty capture) which not only also cannot be casted to a function pointer, but I also now have the additional complexity of trying to reference the lambda type from inside of my `get_module_lock` function. I could, of course, just use `notifiable_module_lock<std::function<void()>>`, but I see no reason we shouldn't just make this the default experience.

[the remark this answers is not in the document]
roxk: > if I want to use a stateful lambda (with a non-empty capture) which not only also cannot be casted to a function pointer, but I also now have the additional complexity of trying to reference the lambda type from inside of my get_module_lock function

Technically you can go `decltype(myLambda)` but I agree it's not a very pleasant API. WRL go over the board in doing placement new and allocating the function storage for the generic type to hide the generic specification - I don't think we need that here either. Would go with `std::function` then.

include/wil/cppwinrt_notifiable_module_lock.h
roxk: Although in practice I'd expect the `notifiable_module_lock` header to be included in `pch.h` and thus it's very rare that there'd be ODR issues, I added the ODR guard just to be safe. Note that API wise, this means the least resistance path needs to be enabled with defining `WINRT_CUSTOM_MODULE_LOCK` + including `notifiable_module_lock` header. For the advanced scenario where user want to use notifiable module lock and defining his own lock, he'd need to also define `WIL_CPPWINRT_COM_SERVER_CUSTOM_MODULE_LOCK`, as shown in the test.

roxk: I have also added the following guard
```
#ifndef WINRT_BASE_H
#error You must include this header before including any winrt header
#endif
```

include/wil/cppwinrt_notifiable_module_lock.h
roxk: Comment is removed, will mark as resolved

tests/CppWinRTComServerCustomModuleLockTests.cpp
roxk: Changed the test to test `WIL_CPPWINRT_COM_SERVER_CUSTOM_MODULE_LOCK`, in case user need to further define custom type for the lock

include/wil/cppwinrt_notifiable_module_lock.h
```diff
@@ -0,0 +1,96 @@
...
+//    ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+//    TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+//    PARTICULAR PURPOSE AND NONINFRINGEMENT.
+//
+//*********************************************************
+//! @file
+//! Utilities for implementing OOP COM server with cppwinrt only
+
+#ifndef __WIL_CPPWINRT_NOTIFIABLE_MODULE_LOCK_DEFINED
+#define __WIL_CPPWINRT_NOTIFIABLE_MODULE_LOCK_DEFINED
```
dunhor, line 15:
```suggestion
#ifndef __WIL_CPPWINRT_NOTIFIABLE_MODULE_LOCK_INCLUDED
#define __WIL_CPPWINRT_NOTIFIABLE_MODULE_LOCK_INCLUDED
```
Just to match the other files

include/wil/cppwinrt_register_com_server.h
```diff
@@ -0,0 +1,107 @@
...
+//    ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+//    TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+//    PARTICULAR PURPOSE AND NONINFRINGEMENT.
+//
+//*********************************************************
+//! @file
+//! Utilities for making managing COM server easier
+
+#ifndef __WIL_CPPWINRT_REGISTER_COM_SERVER_DEFINED
+#define __WIL_CPPWINRT_REGISTER_COM_SERVER_DEFINED
```
dunhor, line 15:
```suggestion
#ifndef __WIL_CPPWINRT_REGISTER_COM_SERVER_INCLUDED
#define __WIL_CPPWINRT_REGISTER_COM_SERVER_INCLUDED
```
Just to match the other files

include/wil/cppwinrt_register_com_server.h
```diff
@@ -0,0 +1,107 @@
...
+    }
+};
+
+template <typename T = void, typename... Rest>
+void register_com_server(std::vector<DWORD>& registrations)
+{
+    DWORD registration{};
+    winrt::check_hresult(CoRegisterClassObject(
+        winrt::guid_of<T>(), winrt::make<CppWinRTClassFactory<T>>().get(), CLSCTX_LOCAL_SERVER, REGCLS_MULTIPLEUSE, &registration));
+    registrations.push_back(registration);
```
dunhor, line 49: Maybe getting too nit-picky, but if this throws, you'll leak the registration. Do we have a pre-existing RAII type for this in WIL? Something like `unique_coclass_cookie`?

dunhor: Actually, with that, you wouldn't even need the `com_server_revoker` type. Just return a `std::vector<wil::unique_coclass_cookie>` and you get effectively the same thing. Revoking is just a call to `.clear()`

tests/CMakeLists.txt
```diff
@@ -92,17 +92,19 @@ link_libraries(${DETOURS_LIBRARY} Catch2::Catch2WithMain)
 
 add_subdirectory(app)
 add_subdirectory(cpplatest)
+add_subdirectory(cppwinrt-com-server)
+add_subdirectory(cppwinrt-com-server-custom-module-lock)
```
dunhor, line 96: You should modify `scripts/runtests.cmd` to ensure that these tests run locally for people and on CI machines.

include/wil/cppwinrt_register_com_server.h
roxk: I think push back is no-throw if a previous call to `reserve` succeeds and the number of element to push < capacity?

dunhor: Gotcha. I missed the `reserve` call because it was in a different location. Worth a comment here.

tests/CMakeLists.txt
roxk: Updated. Ran `runtests.cmd` locally and all `cppwinrt-com-server*` passed.

include/wil/cppwinrt_register_com_server.h
roxk: @dunhor Added comment. I find I like the idea of `unique_coclass_cookie` though, should I add it in another PR or? I'm just afraid `unique_coclass_cookie` alone would take another round of like 20+ comments😅

dunhor: > should I add it in another PR or?

Looking now since I didn't initially, looks like `unique_com_class_object_cookie` already exists and is probably what you want here

roxk: Updated to return `std::vector<wil::unique_com_class_object_cookie>` instead of `com_revoker`.

include/wil/cppwinrt_notifiable_module_lock.h
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    template <typename Func>
+    static void set_notifier(Func&& func)
+    {
+        notifier = std::forward<Func>(func);
+    }
+
+private:
+    std::atomic<int32_t> m_count;
```
dunhor, line 78:
```suggestion
    std::atomic<int32_t> m_count{0};
```
It's been a while, but last time I used `std::atomic`, the default constructor doesn't initialize to zero. I think what you currently have is working because it's a `static` variable, which should get zero initialized.

dunhor: Looks like the construction is trivial until C++20, at which point it gets initialized with `T()`. Does C++/WinRT still support C++17?

include/wil/cppwinrt_register_com_server.h
```diff
@@ -0,0 +1,83 @@
...
+    HRESULT __stdcall CreateInstance(IUnknown* outer, GUID const& iid, void** result) noexcept final
+    {
+        *result = nullptr;
+
+        if (outer)
+        {
+            return CLASS_E_NOAGGREGATION;
+        }
+
+        return winrt::make_self<T>().as(iid, result);
```
dunhor, line 35: Should this have function-level `try...CATCH_RETURN()`?

include/wil/cppwinrt_register_com_server.h
```diff
@@ -0,0 +1,83 @@
...
+{
+    std::vector<wil::unique_com_class_object_cookie> registrations;
+    registrations.reserve(sizeof...(Rest) + 1);
+    try
+    {
+        details::register_com_server<T, Rest...>(registrations);
+        // C++17 doesn't provide guaranteed copy ellision, but the copy should be elided nonetheless.
+        return registrations;
+    }
+    catch (...)
```
dunhor, line 74: It's not necessarily harmful, but the `try...catch` is no longer necessary. The `std::vector`'s destructor should clean up properly

dunhor:
> I found re-registering the same class doesn't trigger any error. Is there any trivial ways to make `CoRegisterClassObject` throw error without using e.g. invalid combination of flags? The doc says `CoRegisterClassObject` would only ever return invalid arguments, out of memory, or unexpected. How to trigger unexpected in a test?

You can [detour the function](https://github.com/microsoft/wil/blob/5c6a7ba43e79edd71dcd1ce4c2d21b0610183c86/tests/mocking.h#L331) to return whatever you wish. E.g.

```c++
witest::detoured_thread_function<&::CoRegisterClassObject> detour([](REFCLSID, LPUNKNOWN, DWORD, DWORD, LPDWORD cookie) {
    *cookie = 0;
    return E_UNEXPECTED;
});
```

include/wil/cppwinrt_notifiable_module_lock.h
roxk: C++/WinRT requires C++17. Think it's best to initialize to 0 nonetheless.

roxk: Btw this is a direct copy of the lock implementation in cppwinrt. You might want to discuss internally whether that should be fixed/file an issue there.

include/wil/cppwinrt_register_com_server.h
roxk: Added. Also handled `winrt::hresult_error` as that caused `CATCH_RETURN` to fail fast.

include/wil/cppwinrt_register_com_server.h
roxk: Removed. Neat.

include/wil/cppwinrt_register_com_server.h
dunhor: Oh right, that's probably because `wil/cppwinrt.h` isn't getting included. This should be fine.

tests/CppWinRTComServerTests.cpp
```diff
@@ -0,0 +1,137 @@
...
+        {
+            REQUIRE(e.code() == E_ACCESSDENIED);
+        }
+    }
+}
+
+winrt::Windows::Foundation::IAsyncAction create_instance_after_5_s()
+{
+    using namespace std::chrono_literals;
+    co_await winrt::resume_after(5s);
```
dunhor, line 107: What benefit does waiting here have? 5 seconds is quite a while to delay test execution for. From what I can gather, there's not really much benefit over doing something simple like `co_await winrt::resume_background();`

roxk: Just simulating a real world usage where the server is created for a while before an instance is created. I guess what you are saying is this is not something worth testing for? I don't know whether there had been edge cases where bugs only manifest in this case (as opposed to the case where the executable is spawned per activation request), so it might indeed be redundant.

include/wil/cppwinrt_register_com_server.h
roxk: Updated to use `wil/cppwinrt.h` :D

roxk:
> You can [detour the function](https://github.com/microsoft/wil/blob/5c6a7ba43e79edd71dcd1ce4c2d21b0610183c86/tests/mocking.h#L331) to return whatever you wish

Added test to test `register_com_server` failure mode.

(this witest detour API is so handy I feel like it should be part of wil instead of `witest`)

dunhor:
> > You can [detour the function](https://github.com/microsoft/wil/blob/5c6a7ba43e79edd71dcd1ce4c2d21b0610183c86/tests/mocking.h#L331) to return whatever you wish
> 
> Added test to test `register_com_server` failure mode.
> 
> (this witest detour API is so handy I feel like it should be part of wil instead of `witest`)

Note that code in `mocking.h` is something I quickly hacked together over the winter holidays during some spare down time. It was inspired by/meant to resemble a lightweight version of an internal library we have for function mocking for tests. The internal WIL unit tests that existed before this repo used this library for a number of tests, and when I was porting those tests here, I had to hack together some alternate solutions, several of which have caused issues over the years, prompting me to create that header. The hope is that one day that library can be open sourced. That would be much better than this hacked together, minimally tested version that was mostly only designed to fit our testing needs. I actually looked into trying this, however at the time rolling my own seemed like the less time-consuming option.

include/wil/cppwinrt_notifiable_module_lock.h
```diff
@@ -0,0 +1,96 @@
...
+
+    template <typename Func>
+    static void set_notifier(Func&& func)
+    {
+        notifier = std::forward<Func>(func);
+    }
+
+private:
+    std::atomic<int32_t> m_count{0};
+    static inline std::function<void()> notifier;
```
dunhor, line 79: Just noticed this; feels weird that this is a `static` variable, especially when `m_count` is not. I get that it's so that `set_notifier` can be a static member, but you can solve that problem in other ways. E.g. have a `static` `instance` method that `get_module_lock` also calls. E.g.
```c++
static notifiable_module_lock& instance() noexcept
{
    static notifiable_module_lock lock;
    return lock;
}
```
So then the usage becomes:
```c++
wil::notifiable_module_lock::instance().set_notifier(fn);
```
You could technically also have `set_notifier` remain `static`, but call `instance` to set the member.

tests/CppWinRTComServerTests.cpp
dunhor: In this case, it seems like you're more testing order of operations:
1. Register server
2. Bump module ref count w/ coroutine
3. Bump module ref count with `CoCreateInstance`
4. Drop both module ref counts
5. Main thread completes operation. 

The only real synchronization point is between 4 & 5, but that's already being handled by your event wait. The sleep doesn't _really_ help test much more than to effectively ensure that the wait for 5 happens before 3. You can get the same effect with additional synchronization. E.g. something maybe like:
```c++
TEST_CASE("CppWinRTComServerTests::NotifierAndRegistration", "[cppwinrt_com_server]")
{
    wil::unique_event moduleEvent(wil::EventOptions::None);
    wil::unique_event coroutineRunning(wil::EventOptions::None);
    wil::unique_event coroutineContinue(wil::EventOptions::None);

    wil::notifiable_module_lock::set_notifier([&] {
        moduleEvent.SetEvent();
    });

    winrt::init_apartment();

    auto revoker = wil::register_com_server<MyServer>();

    [&]() {
        coroutineRunning.SetEvent();

        coroutineContinue.wait();
        auto instance = create_my_server_instance();
        REQUIRE(winrt::get_module_lock() == 2);
    }();

    coroutineRunning.wait();
    REQUIRE(winrt::get_module_lock() == 1); // Coroutine bumped count

    coroutineContinue.SetEvent();
    moduleEvent.wait();

    REQUIRE(winrt::get_module_lock() == 0);
}
```

include/wil/cppwinrt_notifiable_module_lock.h
roxk: Right. I'm more inclined to make the whole class an explicit singleton. Right now it is a combination of singleton (the notifier) and non-singleton (the count). It wasn't a problem in cppwirnt even though there is supposed to be only 1 module lock because it was in the `impl` namespace and users are not supposed to directly use it

I feel like making the ctor private, and only provide the static `instance()` is the right way to go. But I'd have to figure out whether it still allows users to provide custom module lock composed of `wil::notifiable_module_lock` in some way.

tests/CppWinRTComServerTests.cpp
roxk:
```
coroutineContinue.SetEvent();
moduleEvent.wait();
```
Wouldn't it technically means `moduleEvent.wait()` can execute after whatever is after `coroutineContinue.wait()`?

include/wil/cppwinrt_notifiable_module_lock.h
roxk: End up making ctor protected as I found customizing a singleton unpleasant. You basically have to implement the module lock again and forward all method calls to the singleton.

API wise I think singleton makes sense, but I'm not sure if it is acceptable by wil's standard. E.g. memory concern if the static instance is not used (although I bet if the static instance is not used, compiler is probably going to trim it).

tests/CppWinRTComServerTests.cpp
dunhor: That's going to be true, even with the sleep (just unlikely). But that's also not entirely bad. You could also technically validate that the even is not set before calling `coroutineContinue.SetEvent()`

include/wil/cppwinrt_notifiable_module_lock.h
dunhor: I think it feels a little awkward to have a singleton class with an `instance` method be designed such that users can derive from it. You could consider splitting it into two types: a "base type" that just has all the methods, data members, etc. that does not have an `instance` method, and an "impl" type that derives from this base type and exposes just an `instance` method.

> E.g. memory concern if the static instance is not used

Function statics are not initialized until the first time they are accessed. I'd also like to believe that the linker is smart enough to prune unused memory if the function is never called. That said, the variable will not exist unless this header is included, and this header wouldn't ever be included unless the consumer _wants_ that variable to exist.

dunhor (approved):
I still think it would be worth splitting the module lock into a base class and impl class so that consumers aren't deriving from a singleton type. That said, I'm still not super sold on the usefulness of `WIL_CPPWINRT_COM_SERVER_CUSTOM_MODULE_LOCK`, so I'm not sure many people would really encounter this.

Otherwise, from an API perspective, this looks good to me. Thank you for working through all the feedback. Functionality wise, things look good, too, however my expertise with COM is relatively limited. I'll give some time to see if someone like @jonwis, @ChrisGuzak, or @oldnewthing have anything to say.

roxk:
> That said, I'm still not super sold on the usefulness of WIL_CPPWINRT_COM_SERVER_CUSTOM_MODULE_LOCK, so I'm not sure many people would really encounter this.

Indeed. One can argue that if they are implementing a custom module lock anyways, they could add the `set_notifier` function themselves.

In the meantime, I would go ahead and split the lock impl to a base and an impl class.

ysc3839:
I think you should add `REGCLS_SUSPENDED` flag in `CoRegisterClassObject`, and call `CoResumeClassObjects` at the end:
https://github.com/microsoft/wil/blob/eb9eb355d6de431311d7f5ea3590c72d53741ccc/include/wil/cppwinrt_register_com_server.h#L47-L55

Just like what WRL does:
https://github.com/microsoft/win32metadata/blob/527806d20d83d3abd43d16cd3fa8795d8deba343/generation/WinSDK/RecompiledIdlHeaders/winrt/wrl/module.h#L386-L395

PR #447, opened 2024-05-28
added -> operator to unique_wdf_object_reference

bevancollins:


rahul-ramadas:
I'm not quite sure I see the utility of adding this. Overloading `->` makes sense when what is returned is itself a pointer, so that you can apply indirection on it to access its fields/methods. But in this case, what is returned is a WDF object type like `WDFREQUEST`, which doesn't have any fields/methods. Can you please exemplify the use-case with some sample code?

Note that this is also the reason why `unique_any` also does not have an overloaded `->`. `unique_any` is meant for "handle" types, not pointer types.

bevancollins:
this is what I'd like to do:

    wil::unique_wdf_object_reference<PDRIVER_OBJECT> device;
    IoGetDeviceObjectPointer(&name, FILE_WRITE_DATA, &file_object, &device);
    irp = IoAllocateIrp(device->StackSize, FALSE);

bevancollins:
oops, sorry I screwed up.

PR #448, opened 2024-06-24
TraceLogging.h preprocessor conformance

dunhor:
@duhowett hit an issue where the `TraceLogging.h` header gave issues when compiling with `/Zc:preprocessor`. Turns out all of the issues were related to using an empty `__VA_ARGS__` after a comma, which passes a single empty argument. The fix is to use the `##__VA_ARGS__` compiler extension that all major compilers support now, which will remove the trailing comma if `__VA_ARGS__` is empty.

NOTE: It would be more conformant to use `__VA_OPT__` for this purpose, _however_ we can't yet assume >= C++20 and forking all of these code paths would be excessive.

Additional note: Looks like we can finally enable `-fno-ms-compatibility` with Clang, which revealed a couple more issues that have been fixed.

### DIFF
--- a/include/wil/Tracelogging.h
+++ b/include/wil/Tracelogging.h
@@ -6602,13 +6602,13 @@ namespace details
 #else
             m_fireEventDelay = __TRACELOGGING_TEST_HOOK_API_TELEMETRY_EVENT_DELAY_MS;
 #endif
-            m_fireEventThreadPoolTimer.reset(::CreateThreadpoolTimer(
-                [](PTP_CALLBACK_INSTANCE, PVOID, PTP_TIMER) {
-                    FireEvent();
-                },
-                nullptr,
-                nullptr));
+            m_fireEventThreadPoolTimer.reset(::CreateThreadpoolTimer(&FireEventCallback, nullptr, nullptr));
             ScheduleFireEventCallback();
+        }
+
+        static void __stdcall FireEventCallback(PTP_CALLBACK_INSTANCE, PVOID, PTP_TIMER)
+        {
+            FireEvent();
         }
 
         ~ApiTelemetryLogger() WI_NOEXCEPT override
--- a/include/wil/Tracelogging.h
+++ b/include/wil/Tracelogging.h
@@ -743,7 +743,7 @@ public:
     __pragma(warning(push)) __pragma(warning(disable : 4127)) do \
     { \
         _tlgActivityDecl(activity) \
-            TraceLoggingWriteActivity(TraceLoggingType::Provider(), (name), _tlgActivityRef(activity).Id(), NULL, __VA_ARGS__); \
+            TraceLoggingWriteActivity(TraceLoggingType::Provider(), (name), _tlgActivityRef(activity).Id(), NULL, ##__VA_ARGS__); \
     } \
     while (0) \
     __pragma(warning(pop)) /// @endcond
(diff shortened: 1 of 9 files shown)


include/wil/Tracelogging.h
```diff
@@ -1299,15 +1299,17 @@ class ActivityBase : public details::IFailureCallback
...
     __pragma(warning(push)) __pragma(warning(disable : 4127)) do \
     { \
-        _tlgActivityDecl(activity) static const UINT64 _tlgActivity_Keyword = _tlgActivityRef(activity).Keyword; \
-        static const UINT8 _tlgActivity_Level = _tlgActivityRef(activity).Level; \
-        static const UINT64 _tlgActivityPrivacyTag = _tlgActivityRef(activity).PrivacyTag; \
+        _tlgActivityDecl(activity); \
+        using _tlg_Activity_t = wistd::remove_reference_t<decltype(activity)>; \
+        static constexpr const UINT64 _tlgActivity_Keyword = _tlg_Activity_t::Keyword; \
+        static constexpr const UINT8 _tlgActivity_Level = _tlg_Activity_t::Level; \
+        static constexpr const UINT64 _tlgActivityPrivacyTag = _tlg_Activity_t::PrivacyTag; \
```
dunhor, line 1306: This is probably the riskiest part of this change, however I'd like to believe that any place where `(*this).Name` works, `decltype(*this)::Name` should also work. Any non-constant should trigger an error in the `static_assert`, even for MSVC

include/wil/Tracelogging.h
```diff
@@ -6597,15 +6602,15 @@ namespace details
 #else
             m_fireEventDelay = __TRACELOGGING_TEST_HOOK_API_TELEMETRY_EVENT_DELAY_MS;
 #endif
-            m_fireEventThreadPoolTimer.reset(::CreateThreadpoolTimer(
-                [](PTP_CALLBACK_INSTANCE, PVOID, PTP_TIMER) {
-                    FireEvent();
-                },
-                nullptr,
-                nullptr));
+            m_fireEventThreadPoolTimer.reset(::CreateThreadpoolTimer(&FireEventCallback, nullptr, nullptr));
```
DHowett, line 6605: Based on this and the EnumWindows changes, I'm guessing that clang (or someone?) doesn't like the no-capture lambdas being passed as win32 API callbacks due to the calling convention?

I thought that that had been fixed as of https://github.com/llvm/llvm-project/commit/ec809e4cfe0be08b1bab7ba7e6ff905b9430571a, but I legitimately can't get it to work in Compiler Explorer...

DHowett: Oh okay, I did it in a really stupid way. You don't need to do this, but for discussion purposes:

https://godbolt.org/z/1q9YcYndr

```c++
extern "C" void CallbackWithStdcall(int(__stdcall * ParamType)());
extern "C" void CallbackWithCdecl(int(__cdecl * ParamType)());

extern "C" int GetExternalValue();

#ifndef _MSC_VER
#define LAMBDA_CALLING_CONVENTION __attribute__((stdcall))
#else
#define LAMBDA_CALLING_CONVENTION
#endif

void bah() {
    CallbackWithStdcall([]() LAMBDA_CALLING_CONVENTION { return GetExternalValue(); });
    CallbackWithCdecl([]{ return GetExternalValue(); });
}
```

this compiles "properly" under clang and MSVC with the target triple set appropriately

DHowett (approved):
I have no power here, but the changes look correct to me. :)

include/wil/Tracelogging.h
dunhor: That was the first thing I tried, however it caused the compiler to ICE in at least one circumstance...

ChrisGuzak (approved):

Issue #117, opened 2020-02-17
MinGW support?


Berrysoft:
WIL library is not compatible with MinGW(-W64) now, mostly because of gcc, which doesn't recognize some non-standard syntax. I also tried clang for mingw-w64, but it gave different errors.

test file:
``` cpp
#include <wil/result.h>

int main()
{
    return 0;
}
```

g++:
It is too long, and is attached as file.
[gcc.log](https://github.com/microsoft/wil/files/4211771/gcc.log)

clang++:
```
In file included from test.cpp:1:
In file included from C:\msys64\mingw64\include\wil/result.h:16:
In file included from C:\msys64\mingw64\include\wil/result_macros.h:17:
C:\msys64\mingw64\include\wil/common.h:261:1: error: C++ requires a type specifier for all declarations
WI_ODR_PRAGMA("WIL_KERNEL_MODE", "0")
^
C:\msys64\mingw64\include\wil/common.h:256:37: note: expanded from macro 'WI_ODR_PRAGMA'
#define WI_ODR_PRAGMA(NAME, TOKEN)  __pragma(detect_mismatch("ODR_violation_" NAME "_mismatch", TOKEN))
                                    ^
C:\msys64\mingw64\include\wil/common.h:261:1: error: use of undeclared identifier 'detect_mismatch'
C:\msys64\mingw64\include\wil/common.h:256:46: note: expanded from macro 'WI_ODR_PRAGMA'
#define WI_ODR_PRAGMA(NAME, TOKEN)  __pragma(detect_mismatch("ODR_violation_" NAME "_mismatch", TOKEN))
                                             ^
C:\msys64\mingw64\include\wil/common.h:470:6: error: linker pragma must include g_header_init variation
    #error linker pragma must include g_header_init variation
     ^
C:\msys64\mingw64\include\wil/common.h:261:38: error: expected ';' after top level declarator
WI_ODR_PRAGMA("WIL_KERNEL_MODE", "0")
                                     ^
In file included from test.cpp:1:
In file included from C:\msys64\mingw64\include\wil/result.h:16:
C:\msys64\mingw64\include\wil/result_macros.h:93:1: error: C++ requires a type specifier for all declarations
WI_ODR_PRAGMA("WIL_AllocateMemory", "1")
^
C:\msys64\mingw64\include\wil/common.h:256:37: note: expanded from macro 'WI_ODR_PRAGMA'
[...]

PR #454, opened 2024-07-25
LLVM MinGW compatibility improvements

m417z:
I tried to use WIL with [LLVM MinGW](https://github.com/mstorsjo/llvm-mingw). After the changes in this PR and with the compiler options and definitions below, I'm able to compile the main WIL functionality.

Related issue: https://github.com/microsoft/wil/issues/117.

Compiler options:
```
-fms-extensions
```

Definitions added before including WIL:
```cpp
#define _Frees_ptr_
#define _Frees_ptr_opt_
#define _Post_z_
#define _Pre_maybenull_
#define _Pre_opt_valid_
#define _Pre_valid_
#define _Ret_opt_bytecap_(size)
#define _Translates_last_error_to_HRESULT_
#define _Translates_NTSTATUS_to_HRESULT_(status)
#define _Translates_Win32_to_HRESULT_(err)
#define InterlockedDecrementNoFence InterlockedDecrement
#define InterlockedIncrementNoFence InterlockedIncrement
#define WIL_NO_SLIM_EVENT
```

### DIFF
--- a/include/wil/result_macros.h
+++ b/include/wil/result_macros.h
@@ -2851,7 +2851,7 @@ namespace details
     // NOTE: The following two functions are unfortunate copies of strsafe.h functions that have been copied to reduce the friction associated with using
     // Result.h and ResultException.h in a build that does not have WINAPI_PARTITION_DESKTOP defined (where these are conditionally enabled).
 
-    static STRSAFEAPI WilStringLengthWorkerA(
+    inline HRESULT WilStringLengthWorkerA(
         _In_reads_or_z_(cchMax) STRSAFE_PCNZCH psz,
         _In_ _In_range_(<=, STRSAFE_MAX_CCH) size_t cchMax,
         _Out_opt_ _Deref_out_range_(<, cchMax) _Deref_out_range_(<=, _String_length_(psz)) size_t* pcchLength)
@@ -2883,7 +2883,7 @@ namespace details
     }
 
     _Must_inspect_result_
-    STRSAFEAPI StringCchLengthA(
+    inline HRESULT StringCchLengthA(
         _In_reads_or_z_(cchMax) STRSAFE_PCNZCH psz,
         _In_ _In_range_(1, STRSAFE_MAX_CCH) size_t cchMax,
         _Out_opt_ _Deref_out_range_(<, cchMax) _Deref_out_range_(<=, _String_length_(psz)) size_t* pcchLength)
@@ -2905,7 +2905,7 @@ namespace details
     }
 #pragma warning(pop)
 
-    _Post_satisfies_(cchDest > 0 && cchDest <= cchMax) static STRSAFEAPI
+    _Post_satisfies_(cchDest > 0 && cchDest <= cchMax) inline HRESULT
         WilStringValidateDestA(_In_reads_opt_(cchDest) STRSAFE_PCNZCH /*pszDest*/, _In_ size_t cchDest, _In_ const size_t cchMax)
     {
         HRESULT hr = S_OK;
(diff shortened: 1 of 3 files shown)


include/wil/result_macros.h
```diff
@@ -2851,8 +2851,8 @@ namespace details
     // NOTE: The following two functions are unfortunate copies of strsafe.h functions that have been copied to reduce the friction associated with using
     // Result.h and ResultException.h in a build that does not have WINAPI_PARTITION_DESKTOP defined (where these are conditionally enabled).
 
-    static STRSAFEAPI WilStringLengthWorkerA(
-        _In_reads_or_z_(cchMax) STRSAFE_PCNZCH psz,
+    inline HRESULT WilStringLengthWorkerA(
```
dunhor, line 2854: Note this isn't quite a 1:1 change since `STRSAFEAPI` is defined as `__stdcall`. Since it's an inline definition, I don't think it should matter, though

dunhor: This could potentially be problematic if anyone is taking a function pointer to one of the functions, _however_
1. All instances appear to be in the `details` namespace, so no 3rd party code should be referencing these functions
2. I did a search on our internal code and saw no evidence of this potentially being done

So this should be fine. Do you know the reason why `STRSAFEAPI` caused issues in the first place in case I overlooked something and `__stdcall` needs to be added back?

m417z: I assumed it's because of the extern c, not the calling convention, but I didn't dive into it. I can check whether leaving `__stdcall` will work, but I thought it's better to just be consistent with the other functions, and potentially improve the performance as the calling convention can be optimized the compiler now. My thought was, as you said, that no external code should depend oh the functions under `details`.

dunhor: > but I thought it's better to just be consistent with the other functions

Yeah, I think that's the best course of action at the moment. If issues arise, I may need to re-introduce `__stdcall`, so I was just wondering if that would start breaking things again (and e.g. if we'd need to use something like a macro instead). But that's unlikely

dunhor (approved):

include/wil/result_macros.h
m417z: OK, so for reference, with `static STRSAFEAPI` (expands to `static _STRSAFE_EXTERN_C HRESULT WINAPI`):
```
In file included from file.cpp:61:
In file included from Compiler/include/wil/resource.h:15:
Compiler/include/wil/result_macros.h:2854:12: error: cannot combine with previous 'static' declaration specifier
 2854 |     static STRSAFEAPI  WilStringLengthWorkerA(
      |            ^
Compiler/include/strsafe.h:85:20: note: expanded from macro 'STRSAFEAPI'
   85 | #define STRSAFEAPI _STRSAFE_EXTERN_C __inline HRESULT WINAPI
      |                    ^
Compiler/include/strsafe.h:71:27: note: expanded from macro '_STRSAFE_EXTERN_C'
   71 | #define _STRSAFE_EXTERN_C extern "C"
      |                           ^
In file included from file.cpp:61:
In file included from Compiler/include/wil/resource.h:15:
Compiler/include/wil/result_macros.h:2854:12: error: expected unqualified-id
Compiler/include/strsafe.h:85:20: note: expanded from macro 'STRSAFEAPI'
   85 | #define STRSAFEAPI _STRSAFE_EXTERN_C __inline HRESULT WINAPI
      |                    ^
Compiler/include/strsafe.h:71:34: note: expanded from macro '_STRSAFE_EXTERN_C'
   71 | #define _STRSAFE_EXTERN_C extern "C"
      |                                  ^
In file included from file.cpp:61:
In file included from Compiler/include/wil/resource.h:15:
Compiler/include/wil/result_macros.h:2898:18: error: use of undeclared identifier 'WilStringLengthWorkerA'; did you mean 'StringLengthWorkerA'?
 2898 |             hr = WilStringLengthWorkerA(psz, cchMax, pcchLength);
      |                  ^
Compiler/include/strsafe.h:1734:12: note: 'StringLengthWorkerA' declared here
 1734 | STRSAFEAPI StringLengthWorkerA(STRSAFE_LPCSTR psz,size_t cchMax,size_t *pcchLength) {
      |            ^
3 errors generated.
```
Removing `_STRSAFE_EXTERN_C` (i.e. with `static HRESULT WINAPI`) works.

dunhor:
`CppWinRTAuthoringTests::NotifyPropertyChanged` timed out which is a known issue

PR #459, opened 2024-08-07
Feature: Add an RAII wrapper to support RPC timeouts when making blocking COM calls #458

### LINKED ISSUE #458: Feature: Add an RAII wrapper to support RPC timeouts when making blocking COM calls
See #459 for the details on this

dmachaj:
Closes #458

We have a project where we want to make blocking COM calls to another COM server process, but we want protection from the calls taking an infinite amount of time.  There is an existing, but very old, wrapper that provides this functionality.  Instead of using the old wrapper I am porting it to WIL so that we can use it in a different repo than the original and to share it more broadly because it is useful.

Add a new `wil::rpc_timeout` class to <wil/com.h> along with a few test cases for it.  This class uses a combination of `CoEnableCallCancellation`, `CoDisableCallCancellation`, and `CoCancelCall` to provide the COM functionality.  `CreateThreadpoolTimer` is used to provide a timed callback that will not block on the same thread as the blocking COM call.

This object is intended to live on the stack and have a short lifetime.  It is single-use and applies to all RPC calls made while it is alive.

New test cases in ComTests.cpp.

### DIFF
--- a/include/wil/com.h
+++ b/include/wil/com.h
@@ -3325,6 +3325,66 @@ WI_NODISCARD auto make_range(IEnumXxx* enumPtr)
 #endif // WIL_HAS_CXX_17
 #endif // WIL_ENABLE_EXCEPTIONS
 
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_SYSTEM)
+#ifdef __WIL_WIN32_HELPERS_INCLUDED
+class rpc_timeout
+{
+public:
+    rpc_timeout(DWORD timeoutInMilliseconds):
+        m_threadId(GetCurrentThreadId())
+    {
+        m_cancelEnablementResult = CoEnableCallCancellation(nullptr);
+        if (SUCCEEDED(m_cancelEnablementResult))
+        {
+            m_timer.reset(CreateThreadpoolTimer(&rpc_timeout::timer_callback, this, nullptr));
+            LOG_LAST_ERROR_IF_NULL(m_timer.get());
+            if (m_timer)
+            {
+                FILETIME ft = filetime::get_system_time();
+                filetime::add(ft, filetime::convert_msec_to_100ns(timeoutInMilliseconds));
+                SetThreadpoolTimer(m_timer.get(), &ft, 0, 0);
+            }
+        }
+    }
+
+    ~rpc_timeout()
+    {
+        if (SUCCEEDED(m_cancelEnablementResult))
+        {
+            CoDisableCallCancellation(nullptr);
+        }
+
+        m_timer.reset();
+    }
+
+    bool timed_out() const
+    {
+        return m_timedOut;
+    }
+
+private:
+    // Disable use of new as this class should only be declared on the stack, never the heap.
+    void *operator new(size_t) = delete;
+    void *operator new[](size_t) = delete;
+
+    static void timer_callback(PTP_CALLBACK_INSTANCE /*instance*/, PVOID context, PTP_TIMER /*timer*/)
+    {
+        // The timer is waited upon during destruction so it is safe to rely on the this pointer in context.
+        rpc_timeout* self = static_cast<rpc_timeout*>(context);
+        if (SUCCEEDED(CoCancelCall(self->m_threadId, 0)))
+        {
+            self->m_timedOut = true;
+        }
+    }
+
+    wil::unique_threadpool_timer_nocancel m_timer;
+    HRESULT m_cancelEnablementResult{};
+    DWORD m_threadId{};
+    bool m_timedOut{};
+};
+#endif // __WIL_WIN32_HELPERS_INCLUDED
+#endif // WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_SYSTEM)
+
 } // namespace wil
 
 #endif
--- a/include/wil/com.h
+++ b/include/wil/com.h
@@ -3344,8 +3344,7 @@ Include wil/win32_helpers.h before wil/com.h to use this.
 class rpc_timeout
 {
 public:
-    rpc_timeout(DWORD timeoutInMilliseconds):
-        m_threadId(GetCurrentThreadId())
+    rpc_timeout(DWORD timeoutInMilliseconds) : m_threadId(GetCurrentThreadId())
     {
         m_cancelEnablementResult = CoEnableCallCancellation(nullptr);
         if (SUCCEEDED(m_cancelEnablementResult))
@@ -3378,8 +3377,8 @@ public:
 
 private:
     // Disable use of new as this class should only be declared on the stack, never the heap.
-    void *operator new(size_t) = delete;
-    void *operator new[](size_t) = delete;
+    void* operator new(size_t) = delete;
+    void* operator new[](size_t) = delete;
 
     static void timer_callback(PTP_CALLBACK_INSTANCE /*instance*/, PVOID context, PTP_TIMER /*timer*/)
     {
--- a/include/wil/com.h
+++ b/include/wil/com.h
@@ -3327,6 +3327,20 @@ WI_NODISCARD auto make_range(IEnumXxx* enumPtr)
 
 #if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_SYSTEM)
 #ifdef __WIL_WIN32_HELPERS_INCLUDED
+
+/** RAII support for making cross-apartment (or cross process) COM calls with a timeout applied to them.
+ * When this is active any timed out calls will fail with an RPC error code such as RPC_E_CALL_CANCELED.
+ * This is a shared timeout that applies to all calls made on the current thread for the lifetime of
+ * the wil::rpc_timeout object.
+~~~
+{
+  auto timeout = wil::rpc_timeout(5000);
+  remote_object->BlockingCOMCall();
+  remote_object->AnotherBlockingCOMCall();
+}
+~~~
+Include wil/win32_helpers.h before wil/com.h to use this.
+*/
 class rpc_timeout
 {
 public:
--- a/include/wil/com.h
+++ b/include/wil/com.h
@@ -3331,28 +3331,28 @@ WI_NODISCARD auto make_range(IEnumXxx* enumPtr)
 /** RAII support for making cross-apartment (or cross process) COM calls with a timeout applied to them.
  * When this is active any timed out calls will fail with an RPC error code such as RPC_E_CALL_CANCELED.
  * This is a shared timeout that applies to all calls made on the current thread for the lifetime of
- * the wil::rpc_timeout object.
+ * the wil::com_timeout object.
  * A periodic timer is used to cancel calls that have been blocked too long.  If multiple blocking calls
  * are made, and multiple are timing out, then there may be a total delay of (timeoutInMilliseconds * N)
  * where N is the number of calls.
 ~~~
 {
-  auto timeout = wil::rpc_timeout(5000);
+  auto timeout = wil::com_timeout(5000);
   remote_object->BlockingCOMCall();
   remote_object->AnotherBlockingCOMCall();
 }
 ~~~
 Include wil/win32_helpers.h before wil/com.h to use this.
 */
-class rpc_timeout
+class com_timeout
 {
 public:
-    rpc_timeout(DWORD timeoutInMilliseconds) : m_threadId(GetCurrentThreadId())
+    com_timeout(DWORD timeoutInMilliseconds) : m_threadId(GetCurrentThreadId())
     {
         m_cancelEnablementResult = CoEnableCallCancellation(nullptr);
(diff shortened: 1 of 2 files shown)


include/wil/com.h
```diff
@@ -3325,6 +3325,80 @@ WI_NODISCARD auto make_range(IEnumXxx* enumPtr)
...
+    {
+        // The timer is waited upon during destruction so it is safe to rely on the this pointer in context.
+        rpc_timeout* self = static_cast<rpc_timeout*>(context);
+        if (SUCCEEDED(CoCancelCall(self->m_threadId, 0)))
+        {
+            self->m_timedOut = true;
+        }
+    }
+
+    wil::unique_threadpool_timer_nocancel m_timer;
```
jonwis, line 3393: Is the `nocancel` version the one you want? What happens if the timer fires right after the COM call returns to you but before the dtor finishes here?

jonwis (approved):

include/wil/com.h
dmachaj: The destructor would block until the callback is completed, preventing the `this` pointer from being used after destruction.  The callback doesn't do significant work so it shouldn't block for long.

include/wil/com.h
```diff
@@ -3325,6 +3325,79 @@ WI_NODISCARD auto make_range(IEnumXxx* enumPtr)
...
+~~~
+{
+  auto timeout = wil::rpc_timeout(5000);
+  remote_object->BlockingCOMCall();
+  remote_object->AnotherBlockingCOMCall();
+}
+~~~
+Include wil/win32_helpers.h before wil/com.h to use this.
+*/
+class rpc_timeout
```
brialmsft, line 3344: >rpc_timeout [](http://example.com/codeflow?start=6&length=11)

I've always been skeptical of this name, given that this is absolutely COM-level cancelation, and could theoretically be used in inproc scenarios which don't involve the RPC runtime at all.

include/wil/com.h
```diff
@@ -3325,6 +3325,79 @@ WI_NODISCARD auto make_range(IEnumXxx* enumPtr)
...
+
+/** RAII support for making cross-apartment (or cross process) COM calls with a timeout applied to them.
+ * When this is active any timed out calls will fail with an RPC error code such as RPC_E_CALL_CANCELED.
+ * This is a shared timeout that applies to all calls made on the current thread for the lifetime of
+ * the wil::rpc_timeout object.
+~~~
+{
+  auto timeout = wil::rpc_timeout(5000);
+  remote_object->BlockingCOMCall();
+  remote_object->AnotherBlockingCOMCall();
```
brialmsft, line 3339: This accidentally illustrates one of the subtle gotchas of this sort of timeout: it causes the cancelation of a single topmost call, and if you don't bail out after that 1 failure and make a different hanging call (or if cancelation happens to land while you're between calls, e.g. in a loop), then this will not save you.

Presumably in this comment, remote_object is a C++/WinRT-style wrapper that will throw if BlockingCOMCall returns RPC_E_CALL_CANCELED. But if somebody naively does this same thing without the throwing wrapper, then if both calls hang, the timeout will help with the first and won't apply to the second.

include/wil/com.h
dmachaj: Feel free to suggest a better name.  I don't mind renaming to something more accurate.

include/wil/com.h
dmachaj: I realized this after sending the PR and was not able to finish fixing it up before I left for the day.  The previous internal wrapper used a periodic timer to catch additional calls if the first call did not early-return.  I am bringing that back to ensure that it covers any number of calls.  However, when I added the additional test case for that I encountered a problem that I haven't been able to address just yet.

include/wil/com.h
dmachaj: The problem that I've been having is that `CoCancelCall` is failing with `RPC_E_NO_CONTEXT` which seems to mean that this truly does require RPC and does not work inproc across apartments.

dmachaj: If the primary test thread is changed from MTA to STA then the RPC no context error goes away and CoCancelCall returns S_OK.  That does not seem to cause the thread to unblock, though.

brialmsft: > Feel free to suggest a better name

com_timeout?

> That does not seem to cause the thread to unblock, though.

Yeah, I was doing some testing of my own against that last night and saw the same thing. Cancelation state gets set on the call, but that doesn't result in the client thread being able to return unilaterally. If I allow the callee STA to pump messages after the cancelation, the caller does end up getting RPC_E_CALL_CANCELED without invoking the server method, but yeah it was still blocked on the server to return.

Broadly speaking, issuing a cancelation doesn't guarantee that the cancelation will be seen/respected within any specific timeframe, so letter-of-the-law this is fine. But this sort of gap definitely limits the usefulness of canceling inproc calls.

I'll also note that one can use CoSetCancelObject to allow custom cancelation even for calls that don't pass 7through the COM runtime (as another example of non-RPC cancelation).

include/wil/com.h
```diff
@@ -3325,6 +3325,79 @@ WI_NODISCARD auto make_range(IEnumXxx* enumPtr)
 #endif // WIL_HAS_CXX_17
 #endif // WIL_ENABLE_EXCEPTIONS
 
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_SYSTEM)
+#ifdef __WIL_WIN32_HELPERS_INCLUDED
```
dunhor, line 3329: Is there any potential issue with just including `win32_helpers.h` in this file? `com.h` is not set up to be includable more than once, making patterns like this very brittle and prone to include order issues.

dmachaj: I'm not sure.  I was hoping you and other experts could advise on that.  The filetime helpers are definitely useful in this case so I'd prefer to use them instead of doing ULARGE_INTEGER math manually.

include/wil/com.h
dunhor: > or if cancelation happens to land while you're between calls

The single failing call seems easy enough for consumers to handle themselves (and is arguably good COM hygiene), but this part is a little scary. I second the persistence of the periodic timer

include/wil/com.h
dunhor: Unless there are good, specific concerns, I'd say just include the header. If problems arise, the number of failures will determine how we might go about mitigating the issue (e.g. something like an opt-out for few failures). Not that I anticipate that, though. Risking problems is better than introducing a ticking time bomb of include order dependencies.

dmachaj: That makes sense.  I'll make that change in a future iteration.

include/wil/com.h
```diff
@@ -3325,6 +3325,79 @@ WI_NODISCARD auto make_range(IEnumXxx* enumPtr)
...
+        // The timer is waited upon during destruction so it is safe to rely on the this pointer in context.
+        rpc_timeout* self = static_cast<rpc_timeout*>(context);
+        if (SUCCEEDED(CoCancelCall(self->m_threadId, 0)))
+        {
+            self->m_timedOut = true;
+        }
+    }
+
+    wil::unique_threadpool_timer_nocancel m_timer;
+    HRESULT m_cancelEnablementResult{};
```
dunhor, line 3394: There are several places for failure. A few things to consier:
1. Add an `operator bool` (or some sort of `HRESULT` getter) so that the caller can check if initialization succeeded.
2. Change this to a more generic `m_initResult` that will also change to the result of `GetLastError` when e.g. threadpool functions fail
3. Make non-throwing, failfast, and throwing versions of this type (advanced; not necessarily necessary)

include/wil/com.h
dunhor: One thing to keep in mind - although I don't necessarily think it's an issue here - is that members construct in-order before the constructor runs and destruct in reverse order after the destructor runs. Effectively, this means that the callback could be running while the rest of this object has been destructed (since this is the first member). You're not doing anything crazy like setting `m_threadId` to zero and writing to `m_timedOut` should be no-op, however this also means that it's possible for `CoDisableCallCancellation` to be called before `CoCancelCall` is called in the callback. I'd hope that COM handles this gracefully and no-ops/returns failure for `CoCancelCall` and so no harm is done; just something to consider.

dmachaj: That's a great point.  Reordering fields should eliminate that risk.  I'll do that and add a comment describing it.

include/wil/com.h
dmachaj: I keep going back and forth on this in my head and I haven't been able to converge on a firm opinion.  The old version that I'm porting from was "best effort" in terms of errors in the constructor.  This version continues that approach.  However, I also understand the desire to have stronger error detection if this is part of WIL.

dunhor: I'm _okay_ with it being best effort so long as it's documented as such. That said, I'm sure some consumers would want to know about failure. Perhaps something like the following would suffice?
```c++
operator bool() noexcept
{
    return static_cast<bool>(m_timer);
}
```

dmachaj: I ended up going with option 1 and 3 both.  The nothrow variant can use the bool operator to determine if it is active.  There are throwing and failfast options that will do their thing if construction fails.

dunhor (approved):

dunhor:
I noticed the `CppWinRTComServerTests::NotifierAndRegistration` failure in other pipeline runs; currently looking at it. This is not a new failure

dunhor (approved):

tests/ComTests.cpp
```diff
@@ -3058,4 +3060,175 @@ TEST_CASE("COMEnumerator", "[com][enumerator]")
...
+const winrt::guid CLSID_RPCTimeoutTestServer{L"CED2F47C-200B-476F-BFDC-D0D79B052AC6"};
+HANDLE g_hangHandle{nullptr};
+HANDLE g_doneHangingHandle{nullptr};
+
+TEST_CASE("com_timeout", "[com][com_timeout]")
+{
+    auto init = wil::CoInitializeEx_failfast();
+
+    // These test cases require a COM server that will use RPC for communication so that we can exercise
+    // RPC cancellation.  Additionally, this server needs to support the ability to control if it hangs
```
brialmsft, line 3080: Inproc calls don't go through RPC. COM does use NDR (implemented by rpcrt4.dll) to translate the call into request/response messages. But NDR is generally considered distinct from the RPC runtime. And in the inproc case, we don't use the RPC runtime to transport messages between client and server.

Consider "These test cases require calling a COM server via proxy, so that we can exercise cancellation through the COM runtime."

Normally wouldn't quibble, but comments in public testcases are a bit documentationy.

tests/ComTests.cpp
```diff
@@ -3058,4 +3060,175 @@ TEST_CASE("COMEnumerator", "[com][enumerator]")
...
+    // or returns quickly.  The following class provides that functionality, using some global events to
+    // provide deterministic ordering of operations.
+    //
+    // Normally a server in the same process would not use RPC.  By marking this as non_agile and running
+    // it on an STA thread we are able to get RPC involved enough to test cancellation.
```
brialmsft, line 3085: Per previous comment, consider:
"To ensure the call goes through a proxy, the object must be non-agile and the call must cross between apartments (MTA -> STA in this case)".

tests/ComTests.cpp
```diff
@@ -3058,4 +3060,175 @@ TEST_CASE("COMEnumerator", "[com][enumerator]")
...
+        // This thread must be STA to pull RPC in as mediator between threads.
+        auto init = wil::CoInitializeEx_failfast(COINIT_APARTMENTTHREADED);
+
+        DWORD registration{};
+        REQUIRE_SUCCEEDED(CoRegisterClassObject(
+            winrt::guid{CLSID_RPCTimeoutTestServer},
+            winrt::make<wil::details::CppWinRTClassFactory<RPCTimeoutTestServer>>().get(),
+            CLSCTX_LOCAL_SERVER,
+            REGCLS_MULTIPLEUSE,
+            &registration));
```
brialmsft, line 3125: Would arguably be more straightforward to use agile_ref to create the proxy, rather than activation:
* Add an agile_ref<IStringable> to the MTA scope, ref-captured by the lambda
* On the STA thread, use make_agile to wrap the object in the agile_ref. This replaces CoRegisterClassObject.
* On the MTA thread, use <agileRef>.get() to unwrap the agile_ref. The result is a proxy connected to the STA object. This replaces the CoCreateInstance.

tests/ComTests.cpp
dmachaj: Copied the suggestion verbatim.

tests/ComTests.cpp
dmachaj: Copied the suggestion verbatim.

tests/ComTests.cpp
dmachaj: I was gonna say "that doesn't sound much simpler" but then I tried coding it up and... yeah its simpler.  It also saves me from having that extremely ugly CoCreate polling loop.

brialmsft (approved):

PR #462, opened 2024-08-09
Please ignore: testing why tests sometimes timeout

dunhor:


tests/CppWinRTComServerTests.cpp
```diff
@@ -170,6 +170,7 @@ TEST_CASE("CppWinRTComServerTests::NotifierAndRegistration", "[cppwinrt_com_serv
     coroutineContinue.SetEvent();
     moduleEvent.wait();
 
+    asyncAction.wait_for(winrt::Windows::Foundation::TimeSpan{INFINITE}); // Ensure exception pointer has time to be set
```
dmachaj, line 173: FWIW infinite waits aren't ideal when you have a problem with tests hanging and then timing out 😉

dunhor: Perhaps, but when there are lambda captures by reference it's not necessarily better. Anyway, this is for investigating the crash not the timeout.

dunhor: Although this is apparently hanging now 🤔

dunhor: Okay, apparently just holding a pointer to the `IAsyncAction` means that it'll never complete? Maybe this is just a side effect of using coroutines with lambdas?

dmachaj: I don't think holding a pointer to an async would have any ability to cause it to hang.  Although in general coroutine+lambda is a dangerous combination.  See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rcoro-capture for additional information.

dunhor: The capture is safe here since we wait infinitely for the coroutine to complete. I'm experimenting with holding a reference to the lambda so it doesn't destruct though. That said, idk why holding the pointer causes it not to complete; seems like a C++/WinRT bug?

dunhor:
Of course the pipeline no longer hits the timeout when I'm trying to diagnose it :/

PR #463, opened 2024-08-14
Improve test reliability

### LINKED ISSUE #461: BuildAndTest timeout should be shorter than 6 hours
Runs seem to take ~30 minutes so a 60-minute timeout would provide 2x overhead just in case.  The five additional hours of runtime is wasteful when something stops making progress.

See: https://github.com/microsoft/wil/blob/f0c6a81c0c9a4b23b6801f40554b8bec425a83b4/scripts/azure-pipelines.yml#L39

dunhor:
I originally started this to look into why tests were timing out so frequently. In this adventure, I stumbled across & fixed a number of other spurious failures, however the original timeout issue no longer repros... I'll keep an eye out to see if it comes back.

Also reduces total test timeout to one hour, which fixes #461

### DIFF
--- a/tests/CppWinRTComServerTests.cpp
+++ b/tests/CppWinRTComServerTests.cpp
@@ -50,6 +50,9 @@ TEST_CASE("CppWinRTComServerTests::DefaultNotifiableModuleLock", "[cppwinrt_com_
     _comExit.create();
 
     wil::notifiable_module_lock::instance().set_notifier(notifier);
+    auto resetOnExit = wil::scope_exit([&] {
+        wil::notifiable_module_lock::instance().set_notifier(nullptr);
+    });
 
     winrt::init_apartment();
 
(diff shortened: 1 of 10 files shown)


tests/CppWinRTComServerTests.cpp
```diff
@@ -129,32 +132,49 @@ TEST_CASE("CppWinRTComServerTests::AnyRegisterFailureClearAllRegistrations", "[c
...
 
     winrt::init_apartment();
 
     auto revoker = wil::register_com_server<MyServer>();
 
-    [&]() -> winrt::Windows::Foundation::IAsyncAction {
-        co_await winrt::resume_background();
-        coroutineRunning.SetEvent();
+    std::exception_ptr coroutineException;
+    auto asyncLambda = [&]() -> winrt::Windows::Foundation::IAsyncAction {
```
dmachaj, line 151: _Consider_ commenting that the lambda needs a variable so that the captures live until the wait that the coroutine is done executing.  I almost commented that "you don't need this variable".  That would be true of a non-coroutine lambda.  But it is very much not true here.  I think this subtlety is worth a small comment.

tests/CppWinRTComServerTests.cpp
```diff
@@ -129,32 +132,49 @@ TEST_CASE("CppWinRTComServerTests::AnyRegisterFailureClearAllRegistrations", "[c
...
 
     coroutineRunning.wait();
     REQUIRE(winrt::get_module_lock() == 1); // Coroutine bumped count
 
     coroutineContinue.SetEvent();
     moduleEvent.wait();
 
+    if (coroutineException)
+    {
+        std::rethrow_exception(coroutineException);
```
dmachaj, line 176: Out of curiosity why not just REQUIRE that the corotuneException is empty?  Does the rethrow make catch2 print something useful?

dmachaj (approved):

PR #465, opened 2024-08-14
Update cppwinrt used by WIL tests to the latest released version

### LINKED ISSUE #464: WIL is testing against an outdated version of cppwinrt

dmachaj:
Closes #464

It is nice to keep up to date on dependencies.  Most consumers of WIL have likely moved to a newer version of cppwinrt, so the testing of WIL itself should also move to a newer version.

Update the version defined in cmake from 2.0.221121.5 to 2.0.240405.15 (about a year and a half newer).

1. `scripts\init_all.cmd`
2. `scripts\build_all.cmd`
3. `scripts\runtests.cmd`

No breaks, no test failures.  (Validated x64 of msvc+clang)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ if (NOT DEFINED WIL_BUILD_VERSION)
 endif()
 
 if (NOT DEFINED CPPWINRT_VERSION)
-    set(CPPWINRT_VERSION "2.0.221121.5")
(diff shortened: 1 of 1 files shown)


dmachaj:
Yesterday GitHub had a big outage so the pipeline was never scheduled.  I added one space and then undid it as two commits (that add up to a NOOP) and pushed to trigger a fresh pipeline run.

dunhor (approved):

PR #467, opened 2024-08-21
Define SAL annotations as empty macros in CMake `WIL::WIL` target for MinGW and Clang

sharadhr:
Relevant to #117 and #454 (the latter of which I feel is incomplete). In that PR users had to `#define` away SAL annotations used by WIL as macros with empty definitions. 

When consumers (e.g. [Azure SDK for C++](https://github.com/Azure/azure-sdk-for-cpp)) use WIL with CMake and compile with Clang targeting MinGW, they will have to continuously re-define these macros which is repetitive and unintuitive.

This PR resolves this, especially in the CMake case. 

Now, consumers may simply use `target_link_libraries(my_target PRIVATE WIL::WIL), and when CMake builds, these macros will be appended to every command-line invocation of the consumer's `my_target`, which considerably simplifies matters. 

The function-like macros are added with `target_compile_options` because `target_compile_definitions` does [not support function-like macros](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Help/prop_tgt/COMPILE_DEFINITIONS.rst?ref_type=heads&plain=1#L8).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,27 @@ file(GLOB_RECURSE HEADER_FILES "${PROJECT_SOURCE_DIR}/include/${PROJECT_NAME}/*.
 add_library(${PROJECT_NAME} INTERFACE)
 add_library(${PROJECT_NAME}::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
 
+# If MinGW and Clang, define SAL annotations as macros with empty definitions.
+if(MINGW AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
+    target_compile_definitions(${PROJECT_NAME} INTERFACE
+        _Frees_ptr_=
+        _Frees_ptr_opt_=
+        _Post_z_=
+        _Pre_maybenull_=
+        _Pre_opt_valid_=
+        _Pre_valid_=
+        _Translates_last_error_to_HRESULT_=
+        InterlockedDecrementNoFence=InterlockedDecrement
+        InterlockedIncrementNoFence=InterlockedIncrement
+        WIL_NO_SLIM_EVENT=
+    )
+    target_compile_options(${PROJECT_NAME} INTERFACE
+        "-D_Ret_opt_bytecap_(size)="
+        "-D_Translates_NTSTATUS_to_HRESULT_(status)="
+        "-D_Translates_Win32_to_HRESULT_(err)="
+    )
+endif()
+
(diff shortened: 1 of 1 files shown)


dunhor:
While I don't see anything necessarily worrisome here, I also don't have much experience with MinGW (and no experience with MinGW with WIL), so I'm hoping others with more experience can comment here. From past issues/PRs, it seems @m417z or @Berrysoft might be able to provide feedback on this approach.

dunhor:
> the right solution here is to have these definitions included into MinGW

That sounds like it's probably the right approach to me. @sharadhr any issues with pursuing the Windows SDK specific changes in MinGW itself? That would just leave `WIL_NO_SLIM_EVENT` and `-fms-extensions`. And even then, I'm curious if the slim event inclusion could be deduced in-code with preprocessor checks.

m417z:
Perhaps `WIL_NO_SLIM_EVENT` can become unnecessary if `ReadAcquire`, `WriteRelease` are added into MinGW. Interestingly, [they are defined somewhere](https://github.com/mirror/mingw-w64/blob/eff726c461e09f35eeaed125a3570fa5f807f02b/mingw-w64-tools/widl/include/winnt.h#L6703) (for a specific tool?), but not in the main include files.

dunhor:
> Alternatively, WIL can be modified to use `std::atomic`. That can also help get rid of `InterlockedDecrementNoFence` and `InterlockedIncrementNoFence`

It could maybe be conditionally updated to do so, however WIL cannot in general assume access to the STL

sharadhr:
> That sounds like it's probably the right approach to me. @sharadhr any issues with pursuing the Windows SDK specific changes in MinGW itself?

Good point. I could submit a request to the mailing list. I have already [submitted some](https://sourceforge.net/p/mingw-w64/mailman/message/58788295/), and they were merged in [not too long after](https://sourceforge.net/p/mingw-w64/mailman/message/58789551/). 
___

In all honesty, this entire effort is to get the [Azure SDK for C++](https://github.com/Azure/azure-sdk-for-cpp) compiling on MinGW, and that entire stack exposed just how incomplete MinGW is. For instance, `<webservices.h>` is not available, either.

Biswa96:
Could you provide steps to reproduce any compiler error with mingw-w64 toolchain? It would help to add the required macros in mingw-w64. Please reply to this email https://sourceforge.net/p/mingw-w64/mailman/message/58813213/

PR #469, opened 2024-08-24
Add WI_MaskFlag/WI_RaiseFlag inline macros

axelriet:
Adding inline bitwise mask/raise macros.

- WI_MaskFlag(var, flag)
- WI_MaskAllFlags(var, flag)
- WI_RaiseFlag(var, flag)
- WI_RaiseAllFlags(var, flag)

The new macros are meant for inline usage without side effects on the variable. Examples:

- SetFileAttributes(Path, WI_MaskFlag(Attr, FILE_ATTRIBUTE_READONLY));  // Mask flag inline without modifying the variable
- SetFileAttributes(Path, WI_RaiseFlag(Attr, FILE_ATTRIBUTE_READONLY)); // Raise flag inline without modifying the variable

### DIFF
--- a/include/wil/common.h
+++ b/include/wil/common.h
@@ -414,22 +414,27 @@ check fails as opposed to the invalid parameter handler that the STL invokes. Th
 //!
 //! Common example usage (manipulation of flag variables):
 //! ~~~~
-//! WI_SetFlag(m_flags, MyFlags::Foo);                              // Set a single flag in the given variable
-//! WI_SetAllFlags(m_flags, MyFlags::Foo | MyFlags::Bar);           // Set one or more flags
-//! WI_ClearFlagIf(m_flags, MyFlags::Bar, isBarClosed);             // Conditionally clear a single flag based upon a bool
-//! WI_ClearAllFlags(m_flags, MyFlags::Foo | MyFlags::Bar);         // Clear one or more flags from the given variable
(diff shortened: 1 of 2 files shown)


oldnewthing:
"MaskFlag" is ambiguous. Am I masking out the flag, or masking in the flag? In bitmap graphics, a mask is something that has set bits to represent the bits you want to keep, so my initial guess was that MaskFlag is "remove all flags except the ones in the mask".

"RaiseFlag" sounds like it's raising an exception.

How about `WI_WithFlag` and `WI_WithoutFlag`?

```cpp
SetFileAttributes(Path, WI_WithFlag(Attr, FILE_ATTRIBUTE_READONLY)) // Turn on the readonly attribute
SetFileAttributes(Path, WI_WithoutFlag(Attr, FILE_ATTRIBUTE_READONLY)) // Turn off the readonly attribute
```

Maybe we should also break the `WI_ClearFlag` family of macros so that they cannot be used as expressions.

```cpp
#define WI_ClearAllFlags(var, flags) do { (var) &= ~(flags)); } while ((void)0, 0)
```

That way if somebody tries
```
SetFileAttributes(Path, WI_ClearFlag(Attr, FILE_ATTRIBUTE_READONLY)) // Turn off the readonly attribute
```
they get a build break instead of a mystery variable mutation.

PR #474, opened 2024-09-18
<wil/safecast.h> should have a helper that can cast larger with zero-extension of signed types

### LINKED ISSUE #473: <wil/safecast.h> should have a helper that can cast larger with zero-extension of signed types
I am trying to work through some suppressed compiler warnings in one of our code bases.  Many of the suppressed warnings are [C4312](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4312?view=msvc-170).  This warning is saying that if a signed integer is cast upwards into a pointer type it may become unsafe if the original value was negative.    The cast will fill in the upper bits with ones, which would change the pointer to some address in the kernel address space.

This new helper aims to help with that by casting upwards with 0s as the upper bits.  That will avoid changing the value to something unsafe if it really is treated as a pointer.

(PR to follow shortly)

dmachaj:
Closes #473

I am trying to work through some suppressed compiler warnings in one of our code bases.  Many of the suppressed warnings are [C4312](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4312?view=msvc-170).  This warning is saying that if a signed integer is cast upwards into a pointer type it may become unsafe if the original value was negative.    The cast will fill in the upper bits with ones, which would change the pointer to some address in the kernel address space.

This new helper aims to help with that by casting upwards with 0s as the upper bits.  That will avoid changing the value to something unsafe if it really is treated as a pointer.

Add a new method (`wil::safe_zero_extending_cast`) that takes an integer type and returns an unsigned integer value that has the upper bits set to zero in all cases.  A naive `static_cast` would have zero in the upper bits for positive values, and ones in the upper bits for negative values.  Using this safe cast the extension will always be zeros.

Added some new test cases.  They all pass.  I ran init-all/build-all/test-all locally.  I also ran clang-format.

### DIFF
--- a/include/wil/safecast.h
+++ b/include/wil/safecast.h
@@ -15,6 +15,7 @@
 
 #include "result_macros.h"
 #include <intsafe.h>
+#include <stdint.h>
 #include "wistd_config.h"
 #include "wistd_type_traits.h"
 
@@ -43,6 +44,12 @@ namespace details
     // operation only supports conversions between integral types.
     template <typename NewT, typename OldT>
     constexpr bool both_integral_v = wistd::is_integral<NewT>::value && wistd::is_integral<OldT>::value;
+
+    // Helper template to determine that the cast from OldT to NewT is going to sign extend the
+    // value.  This is only true when the size of NewT is larger than OldT and OldT is signed.
+    template <typename NewT, typename OldT>
+    constexpr bool is_sign_extending_cast_v =
+        (sizeof(NewT) >= sizeof(OldT)) && both_integral_v<NewT, OldT> && wistd::is_signed_v<OldT>;
 
     // Note on native wchar_t (__wchar_t):
     //      Intsafe.h does not currently handle native wchar_t. When compiling with /Zc:wchar_t-, this is fine as wchar_t is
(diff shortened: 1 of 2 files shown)


include/wil/safecast.h
```diff
@@ -374,6 +381,43 @@ HRESULT safe_cast_nothrow(const OldT var, NewT* newTResult)
...
+// useful when the resulting value is cast to a pointer type as it prevents the upper bits from being fill
+// which would adjust the pointed-to address.
+//
+// For example:
+//      wil::safe_zero_extending_cast<ULONG_PTR>(-1)
+// will return 0x00000000`FFFFFFFF on a 64-bit system.
+template <typename NewT, typename OldT, wistd::enable_if_t<details::is_sign_extending_cast_v<NewT, OldT>, int> = 0>
+NewT safe_zero_extending_cast(const OldT var)
+{
+    if constexpr (sizeof(OldT) == sizeof(NewT))
```
jonwis, line 396: Can we pull in `type_traits` and say:

```c++
return static_cast<NewT>(static_cast<std::make_unsigned_t<OldT>>(var)));
```

dmachaj: My understanding is that WIL avoids STL dependencies outside of a few select headers such as <wil/stl.h>.  I'm unsure of the implications of pulling in a dependency like that.

dunhor: Right, we probably don't want to add _new_ STL dependencies, at least not without validating with a full OS build first. FWIW `wistd::make_unsigned` _does_ already exist

dmachaj: Nice, that did the trick without having to pull in any new dependencies.  Thank you for the offline help getting it to compile.

include/wil/safecast.h
```diff
@@ -15,6 +15,7 @@
 
 #include "result_macros.h"
 #include <intsafe.h>
+#include <stdint.h>
```
dunhor, line 18: One thing to note is that `stdint.h` and `intsafe.h` have some conflicting macro definitions that have caused issues in the past. Since you no longer seem to require these types after the recent updates, it's probably safest to remove this include.

dmachaj: Wow, that was some excellent foresight.  My OS build hit some very bizarre breaks in components I didn't directly touch and I think this is why.  This comment saved me a whole lot of head scratching.  Thank you.

dunhor: ObamaPewGIF

jonwis (approved):

PR #477, opened 2024-09-25
Move around HEAP API wrappers to a more correct spot, and add ANSI string support for heap strings

GabeDeBacker:
WIL currently has support for Unicode (wide) strings leveraging HeapAlloc, HeapFree, but not ANSI strings.

In our use of WIL and heap strings the need for ANSI arose when performing actions such as parsing JSON.

This change moves the heap wrappers into a more appropriate spot based on the inclusion of "heapapi.h" (which is where HeapAlloc, etc. are defined) and adds ANSI string support.

Unit tests were also written to test heap strings (which did not exist previously).

All tests have passed.

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -3710,25 +3710,6 @@ WI_NODISCARD inline secure_zero_memory_scope_exit SecureZeroMemory_scope_exit(_I
     return SecureZeroMemory_scope_exit(static_cast<void*>(initializedString), wcslen(initializedString) * sizeof(initializedString[0]));
 }
 
-/// @cond
-namespace details
-{
-    inline void __stdcall FreeProcessHeap(_Pre_opt_valid_ _Frees_ptr_opt_ void* p)
-    {
-        ::HeapFree(::GetProcessHeap(), 0, p);
-    }
-} // namespace details
-/// @endcond
-
-struct process_heap_deleter
-{
-    template <typename T>
-    void operator()(_Pre_valid_ _Frees_ptr_ T* p) const
-    {
-        details::FreeProcessHeap(p);
-    }
-};
-
 struct virtualalloc_deleter
 {
     template <typename T>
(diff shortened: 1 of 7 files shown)


include/wil/resource.h
```diff
@@ -4061,6 +4023,120 @@ typedef weak_any<shared_hfind_change> weak_hfind_change;
...
+
+    struct heap_allocator
+    {
+        static _Ret_opt_bytecap_(size) void* allocate(size_t size) WI_NOEXCEPT
+        {
+            return ::HeapAlloc(::GetProcessHeap(), HEAP_ZERO_MEMORY, size);
+        }
+    };
+
+    struct process_heap_deleter
```
dunhor, line 4048: This was previously not in the `details` namespace & is probably not safe to move.

dunhor:
Looks like Clang has started triggering new warnings; I'll need to take a look at that.

Additionally, looks like this change needs to be formatted. Please run `clang-format`. The simplest is to run `scripts\run-clang-format.cmd` from a VS command window

include/wil/resource.h
DrusTheAxe: Add to details and a #if __WIL_DEPRECATED_... wrapper around the copy outside details ?

include/wil/resource.h
```diff
@@ -4061,6 +4023,120 @@ typedef weak_any<shared_hfind_change> weak_hfind_change;
...
+#ifndef WIL_NO_ANSI_STRINGS
+    template <>
+    struct string_allocator<unique_process_heap_ansistring> : heap_allocator
+    {
+    };
+#endif
+} // namespace details
+/// @endcond
+
+inline auto make_process_heap_string_nothrow(
```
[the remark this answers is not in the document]
dunhor: This code is just copied from elsewhere in the file and the definition/pattern is already set in stone. That said, the only failure condition is allocation failure, detected by null pointer

include/wil/resource.h
dunhor: There are many legitimate uses of the type; deprecating is not correct either.

include/wil/resource.h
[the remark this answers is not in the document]
GabeDeBacker: Correct. This is the exact same pattern as cotaskmem string versions

include/wil/resource.h
GabeDeBacker: I put this back.

GabeDeBacker:
The COM tests are failing, which I've verified have nothing to do with my change, but I am going to investigate anyway.

dmachaj:
> The COM tests are failing, which I've verified have nothing to do with my change, but I am going to investigate anyway.

That failure is in a test case that I added recently(ish).  That is supposed to pump messages until an event is signaled.  Instead of getting signaled the wait failed with `RPC_S_CALLPENDING`.  I don't recall seeing that particular error code during development of the test coverage.  7 of 8 test passes succeeded and only 1 failed so this seems like some sort of race condition.

GabeDeBacker:
@dmachaj  - The failures are occurring because the thread that runs the COM server was only waiting for 10 seconds before it exited instead of waiting indefinitely for the comServer event to be signaled, so depending on how long it took to actually execute things, the COM server thread would exit and fail the REQUIRE(CoWait) with pending. I'm getting it fixed up, no worries.

dmachaj:
> @dmachaj - The failures are occurring because the thread that runs the COM server was only waiting for 10 seconds before it exited instead of waiting indefinitely for the comServer event to be signaled, so depending on how long it took to actually execute things, the COM server thread would exit and fail the REQUIRE(CoWait) with pending. I'm getting it fixed up, no worries.

Having a timeout on the wait is intentional.  Infinite waits in test code are bad because they blow out the entire test pass if they are not signaled.  (That said, its plausible that this is simply slowness and a longer wait would eliminate false positives).

GabeDeBacker:
@dmachaj  - It isn't really infinite if you make sure the event that ends the tread is always signaled :)

dunhor (approved):

PR #480, opened 2024-10-29
Update to use feature test macros & other similar changes

dunhor:
This has the following changes:

* Where possible, prefer the use of feature test macros (as opposed to C++ version checks, etc.) as this is the more supported and reliable way for determining feature support. The only real legitimate remaining use of C++ version checks is for guarding includes (which should also include a `__has_include` check), which might warn when compiling with a standard version less than when the header was introduced.
* Remove reliances in code on MSVC specific version macros (e.g. `_HAS_CXX##`) to remove the reliance on the MSVC-specific define and the reliance on the necessary header from being included first.
* Introduced a new `WI_HAS_INCLUDE` macro to avoid the need to surround all `__has_include` checks with an `#ifdef` check. This macro takes both a header name and a fallback for when `__has_include` is not defined.
* Remove guards in tests for C++17 features as that's now the minimum version we compile the tests with.

### DIFF
--- a/include/wil/com.h
+++ b/include/wil/com.h
@@ -19,10 +19,10 @@
 #include "win32_helpers.h"
 #include "resource.h" // last to ensure _COMBASEAPI_H_ protected definitions are available
 
-#if __has_include(<tuple>)
+#if WI_HAS_INCLUDE(<tuple>, 1) // Tuple is C++11... assume available
 #include <tuple>
 #endif
-#if __has_include(<type_traits>)
+#if WI_HAS_INCLUDE(<type_traits>, 1) // Type traits is old... assume available
 #include <type_traits>
 #endif
 
@@ -2113,7 +2113,7 @@ wil::com_ptr_nothrow<Interface> CoGetClassObjectNoThrow(DWORD dwClsContext = CLS
     return CoGetClassObjectNoThrow<Interface>(__uuidof(Class), dwClsContext);
 }
 
-#if __cpp_lib_apply && __has_include(<type_traits>)
+#if __cpp_lib_apply && WI_HAS_INCLUDE(<type_traits>, 1)
 /// @cond
 namespace details
 {
@@ -2247,7 +2247,7 @@ auto try_com_multi_query(IUnknown* obj)
 }
 #endif
 
-#endif // __cpp_lib_apply && __has_include(<type_traits>)
+#endif // __cpp_lib_apply && WI_HAS_INCLUDE(<type_traits>, 1)
 
 #pragma endregion
 
@@ -3136,8 +3136,7 @@ void for_each_site(_In_opt_ IUnknown* siteInput, TLambda&& callback)
 
 #endif // __IObjectWithSite_INTERFACE_DEFINED__
 
-// if C++17 or greater
-#if WIL_HAS_CXX_17
+#if __cpp_deduction_guides >= 201703L
 #ifdef WIL_ENABLE_EXCEPTIONS
 /// @cond
 namespace details
@@ -3323,7 +3322,7 @@ WI_NODISCARD auto make_range(IEnumXxx* enumPtr)
     return iterator_range(enumPtr);
 }
 
-#endif // WIL_HAS_CXX_17
+#endif // __cpp_deduction_guides >= 201703L
 #endif // WIL_ENABLE_EXCEPTIONS
 
 #if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_SYSTEM)
--- a/include/wil/common.h
+++ b/include/wil/common.h
@@ -372,19 +372,19 @@ check fails as opposed to the invalid parameter handler that the STL invokes. Th
 #endif
 
 /// @cond
-#if (__cplusplus >= 201703) || (_MSVC_LANG >= 201703)
-#define WIL_HAS_CXX_17 1
-#else
-#define WIL_HAS_CXX_17 0
-#endif
-
 // Until we'll have C++17 enabled in our code base, we're falling back to SAL
 #define WI_NODISCARD __WI_LIBCPP_NODISCARD_ATTRIBUTE
-/// @endcond
 
-/// @cond
 #define __R_ENABLE_IF_IS_CLASS(ptrType) wistd::enable_if_t<wistd::is_class<ptrType>::value, void*> = nullptr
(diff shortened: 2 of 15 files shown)


include/wil/win32_helpers.h
```diff
@@ -41,10 +41,10 @@
 #include "wistd_type_traits.h"
 
 /// @cond
-#if _HAS_CXX20 && defined(_STRING_VIEW_) && defined(_COMPARE_)
+#if (__WI_LIBCPP_STD_VER >= 20) && defined(_STRING_VIEW_) && defined(_COMPARE_)
```
dunhor, line 44: This needs some cleanup to ideally include `<compare>` when available and usable and kick the C++ standard version check, but that's for a future change.

include/wil/windowing.h
```diff
@@ -14,6 +14,8 @@
 #include <WinUser.h>
 #include <exception>
 
+#include "common.h"
```
dunhor, line 17: This was missing before & is needed for the `WIL_HAS_EXCEPTIONS` checks

include/wil/common.h
```diff
@@ -372,19 +372,19 @@ check fails as opposed to the invalid parameter handler that the STL invokes. Th
 #endif
 
 /// @cond
-#if (__cplusplus >= 201703) || (_MSVC_LANG >= 201703)
-#define WIL_HAS_CXX_17 1
```
dmachaj, line 376: Should the cpp17 check have an `#error` if false?  It would be nice for people who make that mistake to get a really clear/obvious error instead of a ton of confusing spew.

include/wil/com.h
```diff
@@ -19,10 +19,10 @@
 #include "win32_helpers.h"
 #include "resource.h" // last to ensure _COMBASEAPI_H_ protected definitions are available
 
-#if __has_include(<tuple>)
+#if WI_HAS_INCLUDE(<tuple>, 1) // Tuple is C++11... assume available
```
dmachaj, line 22: Maybe its just me, but the fallback being `true` (if that is possible in the preprocessor) would be more understandable.  I thought `1` meant a version number until I came across the definition of the macro.

include/wil/win32_helpers.h
```diff
@@ -20,12 +20,12 @@
 #include <winreg.h>
 #include <objbase.h>
 
+#include "common.h"
+
 // detect std::bit_cast
-#ifdef __has_include
-#if (__cplusplus >= 202002L || _MSVC_LANG >= 202002L) && __has_include(<bit>)
+#if (__WI_LIBCPP_STD_VER >= 20) && WI_HAS_INCLUDE(<bit>, 1) // Assume present if C++20
```
dmachaj, line 26: Could this be rephrased as `#if WI_HAS_INCLUDE(<bit>, __WI_LIBCPP_STD_VER >= 20)` with equal correctness?

include/wil/windowing.h
dmachaj: I'm surprised that wasn't a build break.  Does this OSS project build WIL with the compiler warnings cranked up to the maximum and /analyze on the command line?

include/wil/winrt.h
```diff
@@ -30,14 +30,11 @@
 
 /// @cond
 #if defined(WIL_ENABLE_EXCEPTIONS) && !defined(__WI_HAS_STD_LESS)
-#ifdef __has_include
-#if __has_include(<functional>)
 #define __WI_HAS_STD_LESS 1
+#if WI_HAS_INCLUDE(<functional>, 1) // Functional header available since C++11... fall back to assume present
```
dmachaj, line 34: Is there any point using include guards for something dating to cpp11?  That is well below the support floor so having guards at all seems unnecessary.

include/wil/windowing.h
dunhor: It would just `ifdef` to false. I'm assuming that the tests include a different WIL header before this one, so it wasn't caught.

include/wil/common.h
dunhor: I removed this instance to reduce the number of macros/different checks we have for C++ standard version to just use the `__WI_LIBCPP_STD_VER` macro.

Or are you suggesting that we introduce an error when not building as C++17 (i.e. enforce a hard minimum). Maybe... but I'm not sure how much that might break. For the longest time C++14 was the minimum. It wasn't until somewhat recently that we bumped the minimum to C++17 for the tests (I.e. we're no longer ensuring C++14 builds, effectively bumping up the "soft" minimum), however I'm not quite sure that we can do that without breaking some dependent projects. Perhaps we should, and just tell them to update their version if they wish to update their version of WIL, however we might want to take a slightly more careful approach before we do that.

include/wil/com.h
dunhor: It seems that the C++ preprocessor _does_ understand `true` correctly. _However_ I'm a little cautious to use that here as C does not natively understand `true`. So if `true` is not defined to anything (e.g. if `stdbool.h` is not included), then `true` will get evaluated as the value `0`. In this case, that's probably a moot point - this header can't be included by C code - so it's probably fine to use `true` here. I'm just wary by default of using `true`/`false` when interfacing with the preprocessor.

include/wil/win32_helpers.h
dunhor: The PR description covers why the C++ standard version check is necessary in situations like this. The compiler doesn't use different include paths for different C++ versions, so if your version of the STL supports C++20, but you're compiling as C++17, the `__has_include` check will evaluate to `true` because the header file exists, however if you include it, you will likely get a warning/error saying that the header is unsupported for your target C++ standard version.

include/wil/winrt.h
dunhor: We still need to account for OS code that may not even _have_ access to the STL, in which case a check like this one will fail.

include/wil/common.h
dmachaj: Ah, so cpp17 is the minimum _for the tests_ but not necessarily for all consumers of WIL.

If cpp14 is the hard floor then maybe detect and error in that case?  Or is the hard floor even lower than that?

include/wil/win32_helpers.h
dmachaj: Ahh.  I misread that part of the PR description.  I did see that part but I didn't understand it well enough.

dmachaj (approved):

PR #485, opened 2024-11-14
Bump pipeline timeout to 90 min

dunhor:
Lately, it seems as though MSVC has been timing out with the 60 min timeout. Bumping up to 90 min.

Not sure why we're getting so many timeouts now; Clang takes ~20 min and I'm pretty sure MSVC used to as well... going to see if this helps at least.


dunhor:
Side note: timeout is probably due to low memory

> ##[warning]Free memory is lower than 5%; Currently used: 95.07%

Builds still seem to be progressing, so the longer timeout is hopefully enough

PR #490, opened 2024-12-18
Deduce STL availability

dunhor:
There's a pattern that's existed in WIL for a while that's only gotten worse over time. A lot of the code is sprinkled with things like:
```c++
#ifdef SOME_HEADER_INCLUDED
// Some functionality
#endif
```
Some headers do this in a "safe" way such that they can be re-included to pick up new definitions (e.g. `resource.h`), however many do not. In the best case scenario, this creates header include order dependencies that can cause builds to sporadically break when include orders change (such as a dependency including new headers). In the worst case scenario, this introduces the possibility for ODR violations when one translation unit defines something different than another that can result in security or reliability issues.

The change here tries to determine if it's "safe" to use the STL so that code doesn't need to check to see if a dependent STL header has been included and can instead include that header itself. This heuristic can be seen in `common.h` and can be overridden by defining `WIL_USE_STL`, ideally on the command line to ensure all translation units get the same definition. This isn't perfect and won't avoid _all_ ODR violations - in particular linking to other libraries is one place where it can be easy to accidentally introduce violations - however it should improve things a fair bit. In theory we could detect those violations with a `detect_mismatch` pragma, however that's likely to flag too many places where no realistic issue exists.

### DIFF
--- a/include/wil/common.h
+++ b/include/wil/common.h
@@ -305,6 +305,29 @@ static_assert(WIL_EXCEPTION_MODE <= 2, "Invalid exception mode");
 #error Must enable exceptions when WIL_EXCEPTION_MODE == 1
 #endif
 
+#ifdef WIL_DOXYGEN
+/** This define is used to control whether or not WIL assumes safe access to the STL.
+This define can be set manually (1 to enable, 0 to disable), otherwise heuristics will be applied in an attempt to
+deduce whether or not the STL is available and can be safely used.
+ */
+#define WIL_USE_STL 1
+#elif !defined(WIL_USE_STL)
+#if !defined(WIL_ENABLE_EXCEPTIONS) || !defined(__has_include)
+// Assume it's not safe to use the STL when:
+//      * Exceptions are not enabled, OR
+//      * We can't check for header presence
+#define WIL_USE_STL 0
+#else
+// Check for several STL headers that have been around since the dawn of time
+#if __has_include(<algorithm>) && __has_include(<exception>) && __has_include(<iterator>) && __has_include(<new>) && \
+    __has_include(<string>) && __has_include(<utility>) && __has_include(<vector>)
+#define WIL_USE_STL 1
+#else
+#define WIL_USE_STL 0
+#endif
+#endif
+#endif
+
 /// @cond
 #ifndef WIL_ITERATOR_DEBUG_LEVEL
 // NOTE: See the definition of 'RESULT_DEBUG' for commentary on the use of 'WIL_KERNEL_MODE' below
--- a/include/wil/registry.h
+++ b/include/wil/registry.h
@@ -223,7 +223,7 @@ namespace reg
     //
 #if defined(WIL_ENABLE_EXCEPTIONS)
 
-#if defined(_STRING_) || defined(WIL_DOXYGEN)
+#if WIL_USE_STL || defined(WIL_DOXYGEN)
     using key_iterator = ::wil::reg::iterator_t<::wil::reg::key_iterator_data<::std::wstring>>;
     using value_iterator = ::wil::reg::iterator_t<::wil::reg::value_iterator_data<::std::wstring>>;
 #endif
@@ -583,7 +583,7 @@ namespace reg
         ::wil::reg::set_value_expanded_string(key, nullptr, value_name, data);
     }
 
-#if (defined(_VECTOR_) && defined(_STRING_)) || defined(WIL_DOXYGEN)
+#if WIL_USE_STL || defined(WIL_DOXYGEN)
     /**
      * @brief The generic set_value template function to write a REG_MULTI_SZ value from a std::vector<std::wstring>
      * @param key An open or well-known registry key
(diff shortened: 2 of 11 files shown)


include/wil/common.h
```diff
@@ -305,6 +305,29 @@ static_assert(WIL_EXCEPTION_MODE <= 2, "Invalid exception mode");
 #error Must enable exceptions when WIL_EXCEPTION_MODE == 1
 #endif
 
+#ifdef WIL_DOXYGEN
```
dmachaj, line 308: Why does the DOXYGEN define have anything to do with STL usage?  On closer inspection this seems to be "if in doxygen mode use STL".  Maybe a comment would help?

include/wil/common.h
```diff
@@ -305,6 +305,29 @@ static_assert(WIL_EXCEPTION_MODE <= 2, "Invalid exception mode");
...
+//      * We can't check for header presence
+#define WIL_USE_STL 0
+#else
+// Check for several STL headers that have been around since the dawn of time
+#if __has_include(<algorithm>) && __has_include(<exception>) && __has_include(<iterator>) && __has_include(<new>) && \
+    __has_include(<string>) && __has_include(<utility>) && __has_include(<vector>)
+#define WIL_USE_STL 1
+#else
+#define WIL_USE_STL 0
+#endif
```
dmachaj, line 327: _Consider_ tagging these endif's with the thing they are ending.

EDIT: Looks like the style preference has changed and a bunch of such comments were removed.  So this seems intentional.

include/wil/registry.h
```diff
@@ -223,7 +223,7 @@ namespace reg
     //
 #if defined(WIL_ENABLE_EXCEPTIONS)
 
-#if defined(_STRING_) || defined(WIL_DOXYGEN)
+#if WIL_USE_STL || defined(WIL_DOXYGEN)
```
dmachaj, line 226: `WIL_DOXYGEN` defines `WIL_USE_STL` so the second condition will never be relevant.  If it is true then the `WIL_USE_STL` check will already be true, rendering it moot.

include/wil/winrt.h
```diff
@@ -29,25 +29,9 @@
 #endif
 
 /// @cond
-#if defined(WIL_ENABLE_EXCEPTIONS) && !defined(__WI_HAS_STD_LESS)
-#define __WI_HAS_STD_LESS 1
-#if WI_HAS_INCLUDE(<functional>, 1) // Functional header available since C++11... fall back to assume present
+#if WIL_USE_STL
 #include <functional>
-#else
-// Fall back to the old way of forward declaring std::less
```
dmachaj, line 37: So is this fallback so outdated that it simply isn't needed anymore?

include/wil/winrt.h
```diff
@@ -1075,7 +1059,7 @@ class iterable_range
 };
 #pragma endregion
 
-#if defined(__WI_HAS_STD_VECTOR) || defined(WIL_DOXYGEN)
+#if (WIL_USE_STL && defined(WIL_ENABLE_EXCEPTIONS)) || defined(WIL_DOXYGEN)
```
dmachaj, line 1062: This condition is repeated a few times.  Would it be useful to define `WIL_USE_STL_WITH_EXCEPTIONS` or something like that?

include/wil/common.h
dunhor: The Doxygen define just means "Doxygen is running" - it's more of a short-circuit so that Doxygen provides the targeted code in the generated documentation. In this case, the desire is for the following comment to be present in the documentation. The point here of providing a separate ifdef for Doxygen is so that we don't have to worry about _which_ "path" Doxygen evaluates & put the comment there

include/wil/common.h
dunhor: Your editor is much better at cascading/aligning than a possibly out-of-date comment

include/wil/registry.h
dunhor: See the other comment. The purpose of `WIL_DOXYGEN` is to short-circuit the logic so that Doxygen includes the code in its documentation pass. The purpose is to not think too hard about whether or not the surrounding logic will evaluate to true

include/wil/winrt.h
dunhor: Pretty much. If we can determine that it's safe to include `<functional>` - and include it - then there's no need to forward declare it (which is a violation of the standard anyway).

include/wil/common.h
dmachaj: Much of the time I'm not reading WIL headers inside an editor.  Such as when I'm code reviewing :p

dmachaj (approved):

ChrisGuzak (approved):

dunhor:
Leaving this open for a little while longer. @oldnewthing voiced some concern offline about kernel/driver code. In the worst-case scenario, this change shouldn't really impact existing code; anything not using the STL before shouldn't magically start using the STL after. Additionally, my naive thought is that kernel code should disable exceptions, in which case we should default deduce `WIL_USE_STL` to false. That said, we may still want to consider incorporating such scenarios into the heuristic logic.

PR #493, opened 2024-12-23
Improve VS debugging, skipping trivial wil functions by adding wil.natstepfilter

ChrisGuzak:
When debugging code that uses wil RAII types, "step in" ends up going into the operators and get functions. 
This change resolves that by
1) adding `wil.natstepfitler`
2) adding that to projects that reference the wil nuget via the targets file `<Natvis>` element.

Learn more [here](https://learn.microsoft.com/en-us/visualstudio/debugger/just-my-code?view=vs-2019#additional-information-on-natstepfilter-and-natjmc-files) and [here](https://learn.microsoft.com/en-us/shows/pure-virtual-cpp-2023/improved-step-filtering-in-the-visual-studio-debugger).

it is annoying that adding wil.natvis/wil.natstepfiler to the project clutters the project with those files. Is there a way to add then by hide them too?


dmachaj (approved):
Neat, I didn't know such a file type existed.

dunhor (approved):
Very cool!

PR #496, opened 2025-01-08
Avoid using bit_cast when !WIL_USE_STL.

0xbadfca11:
Make available for WIL_SUPPRESS_EXCEPTIONS.

### DIFF
--- a/include/wil/win32_helpers.h
+++ b/include/wil/win32_helpers.h
@@ -37,7 +37,7 @@
 #endif
 
 /// @cond
-#if __cpp_lib_bit_cast >= 201806L
+#if WIL_USE_STL && (__cpp_lib_bit_cast >= 201806L)
 #define __WI_CONSTEXPR_BIT_CAST constexpr
 #else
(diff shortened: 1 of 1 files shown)


dunhor:
The primary motivation behind `WIL_USE_STL` was to reduce the potential for ODR violations, which can be caused when two different translation units include the same WIL header file, however the definitions introduced by that file differ because of differing state between those two translation units (e.g. one includes some STL header file while the other doesn't). In that vein, this change is consistent with that goal and I'd call it righteous.

That said, `WIL_USE_STL` has little to do with the presence or absence of exceptions. While it's true that `WIL_ENABLE_EXCEPTIONS` is used by default when deducing whether or not it's safe to set `WIL_USE_STL` to true, the two need not be related to one another and the project can manually define `WIL_USE_STL` however it wishes regardless of whether or not exceptions are enabled. So, the PR description here does not make sense to me.

0xbadfca11:
I am withdrawing as I feel there is a disconnect between my goals and my means.

PR #497, opened 2025-01-09
Adding networking.h to provide RAII and iterator support for networking calls

keith-horton:
Additions:

1.	I #include’d network headers in the “special order” that gives access to all functions and structures
a.	This is because most networking headers follow the horrible practice of tight inter-dependencies
i.	I.e., networking headers will have #ifdef’s hiding functions & structures requiring you previously included some other header.
ii.	Rinse & repeat this process across a few headers, and you see a common grievance: “what’s the magical header includes to use to get this dang thing to compile!”
b.	Currently I just #include’d them all by default.
i.	I was tempted to only #include them all if the caller defined a special #define before including my header.
c.	I’m very much interesting in thoughts (especially by those who have had to deal with header inclusion issues with other wil headers)
i.	Additionally some #if guards to help developers avoid the confusing mess of Winsock 1.1 and Winsock 2.2. headers (they cannot both be included). [and by default Windows.h includes the old 1.1 headers …. Something that breaks almost every Windows network developer at some point]

2.	2 structures to load function pointers that are somewhat unpleasant to deal with.
a.	Winsock has many extensibility models – almost all being a huge PITA and failed miserably.
b.	One of the extensibility models is to expose “extension functions” (functions added after the Winsock APIs were agreed upon) by calling a Winsock IOCTL function (WSAIoctl) with special GUID values to receive function pointers for newer functions.
c.	I wrote 2 fairly straight forward structures that will load all of this stuff and just expose a struct with those function pointers.
d.	E.g. (from a test case)
        ::wil::networking::winsock_extension_function_table test_table = ::wil::networking::winsock_extension_function_table::load();
.
.
.
        auto acceptex_return = test_table.f.AcceptEx(
            listeningSocket.get(),
            acceptSocket.get(),
            acceptex_output_buffer,
            0,
            singleAddressOutputBufferSize,
            singleAddressOutputBufferSize,
            &acceptex_bytes_received,
            &acceptex_overlapped);

e.	Notice that I just make the structure within the winsock_extension_function_table visible – with the simple name ‘f’.
f.	This looks somewhat kludgy – so I’m open to thoughts on this. I followed the other extension API set (RIO – ‘registered I/O’) that returns a well-defined function table: RIO_EXTENSION_FUNCTION_TABLE.
i.	vs. creating methods off of these wil extension classes – mirroring the function declarations in these function tables. Which seemed low value.

3.	A class to encapsulate the most painful structure (imho) in Networking – the sockaddr for IPv4 and IPv6 addresses.
a.	A ‘sockaddr’ is a BSD structure – a 1970’s version of “inheritance” implemented in C. It’s required that all “socket address structures” have the exact same fields as the root ‘sockaddr’ in the exact same layout – then just add on new properties as needed.
i.	For example, for IPv4 addresses there’s a ‘sockaddr_in’; for IPv6 addresses there’s a sockaddr_in6.
ii.	E.g. [SOCKADDR_IN (ws2def.h) - Win32 apps | Microsoft Learn](https://learn.microsoft.com/en-us/windows/win32/api/ws2def/ns-ws2def-sockaddr_in), [sockaddr(3type) - Linux manual page](https://www.man7.org/linux/man-pages/man3/sockaddr.3type.html), [Internet Address Formats (The GNU C Library)](https://www.gnu.org/software/libc/manual/html_node/Internet-Address-Formats.html)
b.	If you’ve read much networking code, you might find the strangest/hackiest/full-of-random-C-casts code all dealing with these unpleasant data structures.
i.	This comment from my header might help explain some of the goals:
    //
    // encapsulates working with the sockaddr datatype
    //
    // sockaddr is a generic type - similar to a base class, but designed for C with BSD sockets (1983-ish)
    // 'derived' structures are cast back to sockaddr* (so the initial struct members must be aligned)
    //
    // this data type was built to be 'extensible' so new network types could create their own address structures
    // - appending fields to the initial fields of the sockaddr
    //
    // note that the address and port fields of TCPIP sockaddr* types were designed to be encoded in network-byte order
    // - hence the common use of "host-to-network" and "network-to-host" APIs, e.g. htons(), htonl(), ntohs(), ntohl()
    //
    // Socket APIs that accept a socket address will accept 2 fields:
    // - the sockaddr* (the address of the derived sockaddr type, cast down to a sockaddr*)
    // - the length of the 'derived' socket address structure referenced by the sockaddr*
    //
    // Commonly used sockaddr* types that are using with TCPIP networking:
    //
    // sockaddr_storage / SOCKADDR_STORAGE
    //   - a sockaddr* derived type that is guaranteed to be large enough to hold any possible socket address
    // sockaddr_in / SOCKADDR_IN
    //   - a sockaddr* derived type designed to contain an IPv4 address and port number
    // sockaddr_in6 / SOCKADDR_IN6
    //   - a sockaddr* derived type designed to contain an IPv6 address, port, scope id, and flow info
    // SOCKADDR_INET
    //   - a union of sockaddr_in and sockaddr_in6 -- i.e., large enough to contain any TCPIP address
    // in_addr / IN_ADDR
    //   - the raw address portion of a sockaddr_in
    // in6_addr / IN6_ADDR
    //   - the raw address portion of a sockaddr_in6
    //
    // SOCKET_ADDRESS
    //   - not a derived sockaddr* type
    //   - a structure containing both a sockaddr* and its length fields, returned from some networking functions
    //
c.	The class is hopefully straight forward – it handles many of the unpleasantries moving between types, dealing with embedded types, proving type-safe accessors, and in handling the cases when one must deal with the unpleasant task of knowing when & how to convert between host-byte-order and network-byte-order and vice versa.
i.	Many ways to construct/set_* the IP address in the class – including from a string expression of the address.
ii.	Setting other properties of an address (port, scope id, flow info).
iii.	Writing out an address to either a WCHAR[] or CHAR[] guaranteed to be large enough to hold the string (so no dynamic memory allocation – just a small variable pushed on the stack); as well as a std::wstring if that’s what’s requested.
iv.	Providing many accessors to retrieve various pointers into the underlying sockaddr.

4.	A small class that provides RAII + iterator semantics for walking the returned list of IP addresses after resolving a name:

    //! class addr_info encapsulates the ADDRINFO structure
    //! this structure contains a linked list of addresses returned from resolving a name via GetAddrInfo
    //! iterator semantics are supported to safely access these addresses
    class addr_info

a.	And free standing functions to call the underlying API to resolve a name – returning the addr_info containing the response.
b.	E.g. a few lines of code makes a tool to dump out local addresses, and prints out addresses when resolving a name:

#include <string>
#include "c:/Users/khorton/source/repos/wil-keith-horton/include/wil/networking.h"

int wmain(int argc, wchar_t** argv)
try
{
   if (argc != 2)
   {
       return 0;
   }

   const auto wsa_cleanup = wil::networking::WSAStartup();

   wprintf(L"Local addresses:\n");
   for (const auto& address : wil::networking::resolve_local_addresses())
   {
       wprintf(L" - %ws\n", address.write_address().c_str());
   }

   wprintf(L"\nResolving %ws:\n", argv[1]);
   for (const auto& address : wil::networking::resolve_name(argv[1]))
   {
       wprintf(L" - %ws\n", address.write_address().c_str());
   }
}
catch (const std::exception& e)
{
    wprintf(L"\nFailed  - %hs\n", e.what());
}

C:\Users\khorton\source\repos\test\x64\Debug>test.exe microsoft.com
Local addresses:
- fe80::fa5c:5327:18fe:5e37
- fe80::7218:2b8a:195b:450c
- fe80::2089:24a5:515b:c06
- 192.168.1.27
- 172.17.112.1

Resolving microsoft.com:
- 20.70.246.20
- 20.76.201.171
- 20.112.250.133
- 20.231.239.246
- 20.236.44.162

### DIFF
--- a/include/wil/network.h
+++ b/include/wil/network.h
@@ -20,8 +20,8 @@
 
 // define WIN32_LEAN_AND_MEAN at the project level to avoid windows.h including older winsock 1.1 headers from winsock.h
 // as including both the Winsock 1.1 header 'winsock.h' and the Winsock 2 header 'winsock2.h' will create compiler errors
-// alternatively, including winsock2.h before windows.h to prevent inclusion of the Winsock 1.1 winsock.h header from within windows.h
-// note: winsock2.h will include windows.h if not already included
+// alternatively, including winsock2.h before windows.h to prevent inclusion of the Winsock 1.1 winsock.h header from within
+// windows.h note: winsock2.h will include windows.h if not already included
 //
 // define _SECURE_SOCKET_TYPES_DEFINED_ at the project level to have access to SocketSecurity* functions in ws2tcpip.h
 //
@@ -407,8 +407,7 @@ namespace network
 
     // begin() and end() support - enabling range-based for loop
     template <typename T>
-    constexpr
-    addr_info_iterator_t<T> end(addr_info_iterator_t<T>) WI_NOEXCEPT
+    constexpr addr_info_iterator_t<T> end(addr_info_iterator_t<T>) WI_NOEXCEPT
     {
         return {};
     }
@@ -487,8 +486,8 @@ namespace network
         //   (this should never happen the caller has a reference, but we failed to get a WSA reference)
         // THEN
         //   this object cannot carry forward any function pointers - it must show successfully loaded == false
-        socket_extension_function_table_t(const socket_extension_function_table_t& rhs) WI_NOEXCEPT :
-            wsa_reference_count{WSAStartup_nothrow()}
+        socket_extension_function_table_t(const socket_extension_function_table_t& rhs) WI_NOEXCEPT
+            : wsa_reference_count{WSAStartup_nothrow()}
         {
             if (!wsa_reference_count || !rhs.wsa_reference_count)
             {
@@ -521,8 +520,7 @@ namespace network
 
     private:
         // constructed via load()
-        socket_extension_function_table_t() WI_NOEXCEPT :
-            wsa_reference_count{WSAStartup_nothrow()}
+        socket_extension_function_table_t() WI_NOEXCEPT : wsa_reference_count{WSAStartup_nothrow()}
         {
         }
 
@@ -546,8 +544,7 @@ namespace network
     }
 
     template <>
-    inline winsock_extension_function_table socket_extension_function_table_t<
-        WINSOCK_EXTENSION_FUNCTION_TABLE>::load() WI_NOEXCEPT
+    inline winsock_extension_function_table socket_extension_function_table_t<WINSOCK_EXTENSION_FUNCTION_TABLE>::load() WI_NOEXCEPT
     {
         winsock_extension_function_table table;
         // if WSAStartup failed, immediately exit
@@ -568,15 +565,7 @@ namespace network
             constexpr DWORD bytes{sizeof(functionPtr)};
             DWORD unused_bytes{};
             const auto error{::WSAIoctl(
-                lambdaSocket,
-                controlCode,
-                &extensionGuid,
-                DWORD{sizeof(extensionGuid)},
-                functionPtr,
-                bytes,
-                &unused_bytes,
-                nullptr,
-                nullptr)};
+                lambdaSocket, controlCode, &extensionGuid, DWORD{sizeof(extensionGuid)}, functionPtr, bytes, &unused_bytes, nullptr, nullptr)};
             return error == 0 ? S_OK : HRESULT_FROM_WIN32(::WSAGetLastError());
         };
 
@@ -622,16 +611,7 @@ namespace network
         table.f.cbSize = bytes;
 
         DWORD unused_bytes{};
-        if (::WSAIoctl(
-                localSocket.get(),
-                controlCode,
-                &rioGuid,
-                DWORD{sizeof(rioGuid)},
-                &table.f,
-                bytes,
-                &unused_bytes,
-                nullptr,
-                nullptr) != 0)
+        if (::WSAIoctl(localSocket.get(), controlCode, &rioGuid, DWORD{sizeof(rioGuid)}, &table.f, bytes, &unused_bytes, nullptr, nullptr) != 0)
         {
             LOG_IF_WIN32_ERROR(::WSAGetLastError());
 
@@ -1120,9 +1100,8 @@ namespace network
         }
 
         const void* const pAddr{
-            family() == AF_INET
-                ? static_cast<const void*>(&m_sockaddr.Ipv4.sin_addr)
-                : static_cast<const void*>(&m_sockaddr.Ipv6.sin6_addr)};
+            family() == AF_INET ? static_cast<const void*>(&m_sockaddr.Ipv4.sin_addr)
+                                : static_cast<const void*>(&m_sockaddr.Ipv6.sin6_addr)};
 
         // the last param to InetNtopW is # of characters, not bytes
         const auto* error_value{::InetNtopW(family(), pAddr, address, INET6_ADDRSTRLEN)};
@@ -1145,9 +1124,8 @@ namespace network
         }
 
         const void* const pAddr{
-            family() == AF_INET
-                ? static_cast<const void*>(&m_sockaddr.Ipv4.sin_addr)
-                : static_cast<const void*>(&m_sockaddr.Ipv6.sin6_addr)};
+            family() == AF_INET ? static_cast<const void*>(&m_sockaddr.Ipv4.sin_addr)
+                                : static_cast<const void*>(&m_sockaddr.Ipv6.sin6_addr)};
 
         // the last param to InetNtopA is # of characters, not bytes
         const auto* error_value{::InetNtopA(family(), pAddr, address, INET6_ADDRSTRLEN)};
--- a/include/wil/network.h
+++ b/include/wil/network.h
@@ -302,11 +302,6 @@ namespace network
     class addr_info_iterator_t
     {
     public:
-        static constexpr addr_info_iterator_t end() WI_NOEXCEPT
-        {
-            return addr_info_iterator_t<T>{};
-        }
-
         // defining iterator_traits allows STL <algorithm> functions to be used with this iterator class.
         // Notice this is a forward_iterator
         // - does not support random-access (e.g. vector::iterator)
@@ -409,6 +404,20 @@ namespace network
         T* m_addrinfo_ptr{nullptr};
         ::wil::network::socket_address m_socket_address{};
     }; // class addr_info_iterator_t
+
+    // begin() and end() support - enabling range-based for loop
+    template <typename T>
+    constexpr
+    addr_info_iterator_t<T> end(addr_info_iterator_t<T>) WI_NOEXCEPT
+    {
+        return {};
+    }
+
+    template <typename T>
+    addr_info_iterator_t<T> begin(addr_info_iterator_t<T> addrinfo) WI_NOEXCEPT
+    {
+        return addrinfo;
+    }
 
     using addr_info_ansi_iterator = addr_info_iterator_t<ADDRINFOA>;
     using addr_info_iterator = addr_info_iterator_t<ADDRINFOW>;
--- a/include/wil/network.h
+++ b/include/wil/network.h
@@ -97,195 +97,6 @@ namespace network
     }
 #endif
 
-    struct WINSOCK_EXTENSION_FUNCTION_TABLE
-    {
-        LPFN_ACCEPTEX AcceptEx{nullptr};
-        LPFN_CONNECTEX ConnectEx{nullptr};
-        LPFN_DISCONNECTEX DisconnectEx{nullptr};
-        LPFN_GETACCEPTEXSOCKADDRS GetAcceptExSockaddrs{nullptr};
-        LPFN_TRANSMITFILE TransmitFile{nullptr};
-        LPFN_TRANSMITPACKETS TransmitPackets{nullptr};
-        LPFN_WSARECVMSG WSARecvMsg{nullptr};
-        LPFN_WSASENDMSG WSASendMsg{nullptr};
-    };
-
-    template <typename F>
-    struct socket_extension_function_table_t;
-    using winsock_extension_function_table = socket_extension_function_table_t<WINSOCK_EXTENSION_FUNCTION_TABLE>;
-    using rio_extension_function_table = socket_extension_function_table_t<RIO_EXTENSION_FUNCTION_TABLE>;
-
-    template <typename F>
-    struct socket_extension_function_table_t
-    {
-        static socket_extension_function_table_t load() WI_NOEXCEPT;
-
-        ~socket_extension_function_table_t() WI_NOEXCEPT = default;
-
-        // can copy, but the new object needs its own WSA reference count
-        // (getting a WSA reference count should be no-fail once the first reference it taken)
-        //
-        // IF the target could not get a WSA reference count
-        //   OR
-        // IF we couldn't get our own WSA reference count
-        //   (this should never happen the caller has a reference, but we failed to get a WSA reference)
-        // THEN
-        //   this object cannot carry forward any function pointers - it must show successfully loaded == false
-        socket_extension_function_table_t(const socket_extension_function_table_t& rhs) WI_NOEXCEPT :
-            wsa_reference_count{WSAStartup_nothrow()}
-        {
-            if (!wsa_reference_count || !rhs.wsa_reference_count)
-            {
-                ::memset(&f, 0, sizeof(f));
-            }
-            else
-            {
-                ::memcpy_s(&f, sizeof(f), &rhs.f, sizeof(rhs.f));
-            }
-        }
-
-        socket_extension_function_table_t& operator=(const socket_extension_function_table_t& rhs) WI_NOEXCEPT
-        {
-            if (!wsa_reference_count || !rhs.wsa_reference_count)
-            {
-                ::memset(&f, 0, sizeof(f));
-            }
-            else
-            {
-                ::memcpy_s(&f, sizeof(f), &rhs.f, sizeof(rhs.f));
-            }
-
-            return *this;
-        }
-
-        //! Returns true if all functions were loaded, holding a WSAStartup reference
-        WI_NODISCARD explicit operator bool() const WI_NOEXCEPT;
-
-        F f{};
-
-    private:
-        // constructed via load()
-        socket_extension_function_table_t() WI_NOEXCEPT :
-            wsa_reference_count{WSAStartup_nothrow()}
-        {
-        }
-
-        // must guarantee Winsock does not unload while we have dynamically loaded function pointers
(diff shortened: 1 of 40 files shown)


include/wil/networking.h
```diff
@@ -0,0 +1,1487 @@
...
+//!
+//! define _SECURE_SOCKET_TYPES_DEFINED_ at the project level to have access to SocketSecurity* functions in ws2tcpip.h
+//!
+//! define INCL_WINSOCK_API_TYPEDEFS at the project level to make function typedefs available across various networking headers
+//! note, winsock2.h defaults is to not include function typedefs - but these can be necessary when supporting multiple OS versions
+//!
+//! Link libs for functions referenced in this file: ws2_32.lib, ntdll.lib, and Fwpuclnt.lib (for secure socket functions)
+
+#if !defined(_WINSOCK2API_) && defined(_WINSOCKAPI_)
+#error The Winsock 1.1 winsock.h header was included before the Winsock 2 winsock2.h header - this will cause compilation errors - define WIN32_LEAN_AND_MEAN to avoid winsock.h included by windows.h
```
DrusTheAxe, line 34: The comment above says the problem is avoided by `#define`'ing `WIN32_LEAN_AND_MEAN` -OR- `#include`'ing `winsock.h` before `Windows.h` but the error message only says the former. The error message should include both options

include/wil/networking.h
```diff
@@ -0,0 +1,1487 @@
...
+
+// the wil header for RAII types
+#include "resource.h"
+
+namespace wil
+{
+//! Functions and classes that support networking operations and structures
+namespace networking
+{
+    class socket_address;
```
DrusTheAxe, line 57: Remove this line (unnecessary)

Forward declaration here but I see no references to it before the class is defined (line 295)

include/wil/networking.h
```diff
@@ -0,0 +1,1487 @@
...
+    //! A type that calls WSACleanup on destruction (or reset()).
+    //! WSAStartup must be called for the lifetime of all Winsock APIs (synchronous and asynchronous)
```
DrusTheAxe, line 59: Does WIL use `//!` or `///` notation?

include/wil/networking.h
```diff
@@ -0,0 +1,1487 @@
...
+        LPFN_CONNECTEX ConnectEx{nullptr};
+        LPFN_DISCONNECTEX DisconnectEx{nullptr};
+        LPFN_GETACCEPTEXSOCKADDRS GetAcceptExSockaddrs{nullptr};
+        LPFN_TRANSMITFILE TransmitFile{nullptr};
+        LPFN_TRANSMITPACKETS TransmitPackets{nullptr};
+        LPFN_WSARECVMSG WSARecvMsg{nullptr};
+        LPFN_WSASENDMSG WSASendMsg{nullptr};
+    };
+
+    struct winsock_extension_function_table
```
DrusTheAxe, line 112: Why is this needed? Why not rely on the usual build time import linkage and IAT?

DrusTheAxe, line 112: No doc-comment

include/wil/networking.h
```diff
@@ -0,0 +1,1487 @@
...
+    //! Calls WSAStartup and throws on error; returns an RAII object that reverts
+    WI_NODISCARD inline unique_wsacleanup_call WSAStartup()
+    {
+        WSADATA unused_data{};
+        THROW_IF_WIN32_ERROR(::WSAStartup(WINSOCK_VERSION, &unused_data));
+        return {};
+    }
+#endif
+
+    struct WINSOCK_EXTENSION_FUNCTION_TABLE
```
DrusTheAxe, line 100: No doc-comment

include/wil/networking.h
```diff
@@ -0,0 +1,1487 @@
...
+    {
+        LPFN_ACCEPTEX AcceptEx{nullptr};
```
DrusTheAxe, line 102: Isn't this redundant?  e.g. should be `LPFN_ACCEPTEX AcceptEx{};`

same comment applies throughout

include/wil/networking.h
```diff
@@ -0,0 +1,1487 @@
...
+        // we need a temporary socket for the IOCTL to load the functions
+        const ::wil::unique_socket localSocket{::socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP)};
+        if (INVALID_SOCKET == localSocket.get())
+        {
+            return table;
+        }
+
+        constexpr DWORD controlCode{SIO_GET_MULTIPLE_EXTENSION_FUNCTION_POINTER};
+        constexpr DWORD bytes{sizeof(table.f)};
+        GUID rioGuid = WSAID_MULTIPLE_RIO;
```
DrusTheAxe, line 1459: NIT: `GUID rioGuid{ WSAID_MULTIPLE_RIO };`

[ES.23: Prefer the {}-initializer syntax](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#es23-prefer-the--initializer-syntax)

Same comment applies throughout. Usually done, this leaped at given the 2 immediately previous lines do it but this line didn't. Any other like cases should be equally updated

include/wil/networking.h
```diff
@@ -0,0 +1,1487 @@
...
+
+        ::memset(&table.f, 0, bytes);
+        table.f.cbSize = bytes;
+
+        DWORD unused_bytes;
```
DrusTheAxe, line 1464: NIT: `DWORD unused_bytes{};`

Always initialize variables

include/wil/networking.h
```diff
@@ -0,0 +1,1487 @@
...
+                controlCode,
+                &rioGuid,
+                DWORD{sizeof(rioGuid)},
+                &table.f,
+                bytes,
+                &unused_bytes,
+                nullptr,
+                nullptr) != 0)
+        {
+            const auto gle = ::WSAGetLastError();
```
DrusTheAxe, line 1476: `const auto gle{ ::WSAGetLastError() };`

see previous comment

include/wil/networking.h
```diff
@@ -0,0 +1,1487 @@
...
+                functionPtr = &table.f.WSASendMsg;
+                break;
+            }
+
+            default:
+                FAIL_FAST();
+            }
+
+            constexpr DWORD controlCode{SIO_GET_EXTENSION_FUNCTION_POINTER};
+            constexpr DWORD bytes{sizeof(void*)};
```
DrusTheAxe, line 1412: Better if `sizeof(functionPtr)`

include/wil/networking.h
```diff
@@ -0,0 +1,1487 @@
...
+            case 0:
+            {
+                constexpr GUID acceptex_guid = WSAID_ACCEPTEX;
+                extensionGuid = acceptex_guid;
+                functionPtr = &table.f.AcceptEx;
+                break;
+            }
+            case 1:
+            {
+                constexpr GUID connectex_guid = WSAID_CONNECTEX;
```
DrusTheAxe, line 1359: That's a rather novel use of loop+switch. Why not more straightforward linear code e.g.

```
inline HRESULT load_function(SOCKET localSocket, GUID extensionGuid, void* functionPtr)
{
    constexpr DWORD controlCode{SIO_GET_EXTENSION_FUNCTION_POINTER};
    constexpr DWORD bytes{sizeof(functionPtr)};
    DWORD unused_bytes{};
    RETURN_IF_WIN32_BOOL_FALSE(::WSAIoctl(localSocket, controlCode, &extensionGuid, DWORD{sizeof(extensionGuid)},
                                          functionPtr, bytes, &unused_bytes, nullptr, nullptr));
    return S_OK;
}

inline winsock_extension_function_table winsock_extension_function_table::load() WI_NOEXCEPT
{
    winsock_extension_function_table table;

    // if WSAStartup failed, immediately exit
    if (!table.wsa_reference_count)
    {
        return table;
    }

    // we need a temporary socket for the IOCTL to load the functions
    const ::wil::unique_socket localSocket{::socket(AF_INET, SOCK_DGRAM, IPPROTO_UDP)};
    if (INVALID_SOCKET == localSocket.get())
    {
        return S_OK;
    }

    // Load the functions into the table
    if (FAILED_LOG(load_function(localSocket.get(), WSAID_ACCEPTEX, &table.f.AcceptEx)) ||
        FAILED_LOG(load_function(localSocket.get(), WSAID_CONNECTEX, &table.f.ConnectEx)) ||
        FAILED_LOG(load_function(localSocket.get(), WSAID_DISCONNECTEX, &table.f.DisconnectEx)) ||
        FAILED_LOG(load_function(localSocket.get(), WSAID_GETACCEPTEXSOCKADDRS, &table.f.GetAcceptExSockaddrs)) ||
        FAILED_LOG(load_function(localSocket.get(), WSAID_TRANSMITFILE, &table.f.TransmitFile)) ||
        FAILED_LOG(load_function(localSocket.get(), WSAID_TRANSMITPACKETS, &table.f.TransmitPackages)) ||
        FAILED_LOG(load_function(localSocket.get(), WSAID_WSARECVMSG, &table.f.WSARecvMsg)) ||
        FAILED_LOG(load_function(localSocket.get(), WSAID_WSASENDMSG, &table.f.WSASendMsg)))
    {
        // if any failed to be found, something is very broken
        // all should load, or all should fail
        ::memset(&table.f, 0, sizeof(table.f));
    }

    return table;
}
```

DrusTheAxe: The PR description has code starting with

```
#include
#include "c:/Users/khorton/source/repos/wil-keith-horton/include/wil/networking.h"
...
```

Is that 1st `#include` a typo or missing something?

P.S. PR descriptions support markdown and would be easier to read e.g. the code marked as code

```
#include "c:/Users/khorton/source/repos/wil-keith-horton/include/wil/networking.h"

int wmain(int argc, wchar_t** argv)
try
{
if (argc != 2)
{
return 0;
}
...
```

include/wil/network.h
```diff
@@ -0,0 +1,1328 @@
...
+//! Helpers for using BSD sockets and Winsock functions and structures.
+//! Does not require the use of the STL or C++ exceptions (see _nothrow functions)
+#ifndef __WIL_NETWORK_INCLUDED
+#define __WIL_NETWORK_INCLUDED
+
+#ifdef _KERNEL_MODE
+#error This header is not supported in kernel-mode.
+#endif
+
+//! define WIN32_LEAN_AND_MEAN at the project level to avoid windows.h including older winsock 1.1 headers from winsock.h
```
dunhor, line 21: Not documenting anything so don't use doc comments (i.e. use `//` instead of `//!`). This comment applies throughout the PR

include/wil/network.h
```diff
@@ -0,0 +1,1328 @@
...
+
+        // write_address prints the IP address, not the scope id or port
+        HRESULT write_address_nothrow(socket_address_wstring& address) const WI_NOEXCEPT;
+        HRESULT write_address_nothrow(socket_address_string& address) const WI_NOEXCEPT;
+
+        // write_complete_address_nothrow() prints the IP address as well as the scope id and port values
+        HRESULT write_complete_address_nothrow(socket_address_wstring& address) const WI_NOEXCEPT;
+        HRESULT write_complete_address_nothrow(socket_address_string& address) const WI_NOEXCEPT;
+
+#if defined(WIL_ENABLE_EXCEPTIONS) && (defined(_STRING_) || defined(WIL_DOXYGEN))
```
dunhor, line 233: Don't do this (checking for the definition of `_STRING_` etc.). See #490 for examples on how to use the STL in WIL moving forward

include/wil/network.h
```diff
@@ -0,0 +1,1328 @@
...
+
+    //! class addr_info encapsulates the ADDRINFO-related structures returned from the socket functions
+    //! getaddrinfo, GetAddrInfoW, GetAddrInfoWEx
+    //! iterator semantics are supported to safely access these addresses
+    // ! template T supports pointers to the 3 address structures: ADDRINFOA*, ADDRINFOW*, ADDRINFOEXW*
+    template <typename T>
+    class addr_info_iterator_t
+    {
+    public:
+        static constexpr addr_info_iterator_t end() WI_NOEXCEPT
```
dunhor, line 305: Why have `end` with no `begin`? If you're going for a `std::filesystem::directory_iterator`-like design, you'll need both. E.g. for
```c++
for (auto xxx : add_info_iterator_t(...)) { ... }
```

include/wil/network.h
```diff
@@ -0,0 +1,1328 @@
...
+
+    using addr_info_ansi_iterator = addr_info_iterator_t<ADDRINFOA>;
+    using addr_info_iterator = addr_info_iterator_t<ADDRINFOW>;
+    // not defining a type for ADDRINFOEXA as that type is formally __declspec(deprecated)
+    using addr_infoex_iterator = addr_info_iterator_t<ADDRINFOEXW>;
+
+#if defined(WIL_ENABLE_EXCEPTIONS)
+    //! wil function to capture resolving IP addresses assigned to the local machine, throwing on error
+    //! returning an RAII object containing the results
+    inline ::wil::unique_addrinfo resolve_local_addresses() WI_NOEXCEPT
```
dunhor, line 421: `noexcept` on a throwing function?

include/wil/network.h
```diff
@@ -0,0 +1,1328 @@
...
+        default:
+            // if not AF_INET or AF_INET6, and families don't match
+            // then just raw memcmp the largest field of the union (v6)
+            return ::memcmp(&lhs.m_sockaddr.Ipv6, &rhs.m_sockaddr.Ipv6, sizeof(SOCKADDR_IN6)) < 0;
+        }
+    }
+
+    inline bool socket_address::operator>(const ::wil::network::socket_address& rhs) const WI_NOEXCEPT
+    {
+        return !(*this < rhs);
```
dunhor, line 794: This would be a valid definition of `operator>=`, not `operator>`

keith-horton:
@dunhor , do you know who owns the ComTests? This is failing on the build machines in the "RPC timeout test" -- I don't know why this COM mechanism is not working on the build machines - as it's running correctly on my system when I run the test locally. (all tests passed on all build flavors).

The only change was adding some whitespace because I ran clang-format  --> and the script ended up formatting those other files.

dmachaj:
> @dunhor , do you know who owns the ComTests? This is failing on the build machines in the "RPC timeout test" -- I don't know why this COM mechanism is not working on the build machines - as it's running correctly on my system when I run the test locally. (all tests passed on all build flavors).
> 
> The only change was adding some whitespace because I ran clang-format --> and the script ended up formatting those other files.
> 
> edit: I added a comment where the test failed after I looked up what that failure was.

I don't own the ComTests as a whole but the timeout test is one that I added (see #459).  It was then tweaked significantly by someone else in #477.  This concept has proven challenging to test reliably :(.  I'm not sure if the timeouts simply need to be extended by a bit or if there is a deeper race somewhere.

keith-horton:
@dunhor , I've run scripts\run-clang-format.cmd and checked that in -- but the build still fails saying I need to run clang-format :(

I'm running scripts\run-clang-format.cmd -- but no files are edited.
Should I be running something else?

(also I think I fixed the COM tests ... I left printf calls there for future folks that need to debug)

PR #498, opened 2025-01-15
`cppwinrt.h`: Improve conversion of error information

### LINKED ISSUE #495: `cppwinrt.h`: Improve conversion of error information
When using WIL with C++/WinRT with the `cppwinrt.h` header, the `winrt_to_hresult_handler` global handler is hooked from WIL directly into the C++/WinRT codebase.

The handler allows WIL exceptions to be understood by C++/WinRT and their HRESULTs extracted, but messages are lost in the way when going over the API boundaries because only the HRESULT is converted in [`winrt::to_hresult()`](https://github.com/microsoft/cppwinrt/blob/a77a2ef2e99059f5d8411bbb08a8299c348f3e73/strings/base_error.h#L429).

C++/WinRT has a magic way to recover the messages when re-creating the exceptions on the caller side. It does that by storing them with `WINRT_IMPL_SetErrorInfo()` and later retrieving them with `WINRT_IMPL_GetErrorInfo()` when re-creating the `winrt::hresult_error`. The set is done inside `winrt::hresult_error::to_abi()` in `winrt::to_hresult()` on the server side, but only when `winrt_to_hresult_handler` is not hooked.

If WIL were to call the `to_abi()` method in the WIL hook for `winrt_to_hresult_handler`, it would properly trigger the storage of the message and later on it could be retrieved when C++/WinRT creates a new `winrt::hresult_error` on the caller side.

arthuraraujo-msft:
Closes #495

When using WIL with C++/WinRT with the `cppwinrt.h` header, the `winrt_to_hresult_handler` global handler is hooked from WIL directly into the C++/WinRT codebase.

The handler allows WIL exceptions to be understood by C++/WinRT and their HRESULTs extracted, but messages are lost in the way when going over the API boundaries because only the HRESULT is converted in [`winrt::to_hresult()`](https://github.com/microsoft/cppwinrt/blob/a77a2ef2e99059f5d8411bbb08a8299c348f3e73/strings/base_error.h#L429).

C++/WinRT makes use of Get and [SetErrorInfo()](https://learn.microsoft.com/en-us/windows/win32/api/oleauto/nf-oleauto-seterrorinfo) to recover the messages when re-creating the exceptions on the caller side. The set is done inside `winrt::hresult_error::to_abi()` in `winrt::to_hresult()` on the server side, but only when `winrt_to_hresult_handler` is not hooked. The retrieval happens when re-creating the `winrt::hresult_error` on the caller.

If WIL were to call the `to_abi()` method in the WIL hook for `winrt_to_hresult_handler`, it would properly trigger the storage of the message and later on it could be retrieved when C++/WinRT creates a new `winrt::hresult_error` on the caller side.

That's what this PR is doing.
I also added a test to cover the message being passed.

### DIFF
--- a/include/wil/cppwinrt.h
+++ b/include/wil/cppwinrt.h
@@ -121,6 +121,20 @@ inline void MaybeGetExceptionString(
     }
 }
 
+inline HRESULT CppWinRtHResultFromCode(HRESULT hr, PCWSTR debugString)
+{
+    if (FAILED(hr))
+    {
+        // Use winrt::hresult_error::to_abi() to save the debugString so it can later be retrieved by C++/WinRT
+        if (debugString)
+        {
+            return winrt::hresult_error(hr, debugString).to_abi();
+        }
+        return winrt::hresult_error(hr).to_abi();
+    }
+    return hr;
+}
+
 inline HRESULT __stdcall ResultFromCaughtException_CppWinRt(
     _Inout_updates_opt_(debugStringChars) PWSTR debugString,
     _When_(debugString != nullptr, _Pre_satisfies_(debugStringChars > 0)) size_t debugStringChars,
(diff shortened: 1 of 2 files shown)


dunhor:
Correct me if I'm wrong, but you seem to effectively be using C++/WinRT as a way to call `RoOriginateError`? There's already a header for that: [result_originate.h](https://github.com/microsoft/wil/blob/master/include/wil/result_originate.h).

The new test failures seem very suspicious, though I can't quite comment on their cause without digging into it.

arthuraraujo-msft:
@dunhor I had no idea that existed! I just tested it out and it does lead to the exceptions having the proper messages.
I don't see this in the docs, is there a place where this file is recommended?

dunhor:
> I don't see this in the docs, is there a place where this file is recommended?

I don't see anything mentioning it in the error handling docs, which is where I'd expect to find it if something existed. I can add a pointer somewhere there later if nobody beats me to it.

Now, there's a question that I think this PR raises as to whether or not cppwinrt.h should include result_originate.h by default as I think it's safe to say that if you're using C++/WinRT, you're using WinRT and probably want error origination. Idk if that might have any adverse side effects, though. E.g. idk if that would lead to the same sorts of issues as you're seeing with the test failures with this PR. @dmachaj might have some insight here as I believe he's more familiar with the error origination logic than I am

arthuraraujo-msft:
@dunhor Thanks! That's a good point. I agree it could be interesting to include it by default.

dmachaj:
If it isn't mentioned in the docs then that is definitely a gap that would benefit from being filled.  I think the result_originate functionality is super useful and that any binaries using COM extensively should include it.

"Now, there's a question that I think this PR raises as to whether or not cppwinrt.h should include result_originate.h by default as I think it's safe to say that if you're using C++/WinRT, you're using WinRT and probably want error origination."

I don't think cppwinrt headers have any dependency on WIL so that would be the explanation for why this doesn't already happen and would be tricky to add.  The cppwinrt error origination already calls `RoOriginateError` (such as in the `winrt::hresult_error` constructor).  The case it doesn't cover is when WIL is also used for logging and control flow.  It is valid to use cppwinrt without using WIL so its not really cppwinrt's job to ensure that you are including all of the best WIL headers.

I think that if we want to improve the default situation then we should consider including <wil/result_originate.h> from some of the other WIL headers such as <wil/cppwinrt.h> or <wil/com.h>.

arthuraraujo-msft:
Thanks folks. I'll close this PR since there is an existing flow to produce the same results!

PR #499, opened 2025-01-22
Add ArgvToCommandLine, the reverse of CommandLineToArgvW

dunhor:
A separate project I was working on required this functionality; thought I'd copy it here. The general gist behind its purpose is that `CreateProcess*` requires a string for the command line as opposed to an `argv` array; when you're in a scenario where you have an `argv` array and need to call `CreateProcess*`, this function is useful. The documentation for the function mentions two specific scenarios in which this is useful, but I'm copying it here for quick reference:

1. When implementing a "driver" application. That is, an application that consumes some command line arguments, translates others into new arguments, and preserves the rest, "forwarding" the resulting command line to a separate application.
2. When reading command line arguments from some data storage, e.g. from a JSON array, which then need to get compiled into a command line string that's used for creating a new process.

Note that the [documentation for `CommandLineToArgvW`](https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-commandlinetoargvw) describes how arguments are parsed; this just does the reverse. The goal of this function is to ensure that the target application's `argv` array is identical to the one passed into this function. This is potentially problematic if the target application does not use its `argv` array (e.g. if it gets the command line string directly, e.g. with `GetCommandLineW`) and instead attempts to perform its own command line processing, which may not handle quotes, etc. identically to `CommandLineToArgvW`. One notable example is `cmd.exe` which does not handle some quoted arguments as one would logically expect. For example, `cmd.exe "/c" "echo foo"` does not behave as you would expect it to:
```cmd
>cmd "/c" "echo foo"
'"echo foo' is not recognized as an internal or external command,
operable program or batch file.
```
This is the motivating factor behind `ArgvToCommandLineFlags::ForceQuotes`. Ideally, we would always surround all arguments with quotes as that's the more efficient thing to do, however as you can see from the above, this is not always safe to do.

### DIFF
--- a/include/wil/win32_helpers.h
+++ b/include/wil/win32_helpers.h
@@ -40,6 +40,10 @@
 #include "wistd_functional.h"
 #include "wistd_type_traits.h"
 
+#if defined(WIL_ENABLE_EXCEPTIONS) && (__cpp_lib_string_view >= 201606L)
+#include <string>
+#endif
+
 /// @cond
 #if _HAS_CXX20 && defined(_STRING_VIEW_) && defined(_COMPARE_)
 // If we're using c++20, then <compare> must be included to use the string ordinal functions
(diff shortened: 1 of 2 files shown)


include/wil/win32_helpers.h
```diff
@@ -49,6 +50,10 @@
 #include "wistd_functional.h"
 #include "wistd_type_traits.h"
 
+#if defined(WIL_ENABLE_EXCEPTIONS) && (__cpp_lib_string_view >= 201606L)
+#include <string>
+#endif
+
```
dunhor, line 56:
```suggestion
```
From a previous version; removing

include/wil/win32_helpers.h
```diff
@@ -888,6 +889,237 @@ bool init_once(_Inout_ INIT_ONCE& initOnce, T func)
...
+        while (pos < str.size())
+        {
+            auto nextPos = str.find_first_of(searchString, pos);
+            if ((nextPos != str.npos) && ((str[nextPos] == ' ') || (str[nextPos] == '\t')))
+            {
+                // Insert the quote now since we'll need to otherwise stomp over data we're about to write
+                // NOTE: By updating the search string here, we don't need to worry about manually inserting the
+                // character later since we'll just include it in our next iteration
+                WI_ASSERT(!forceQuotes);               // Otherwise, shouldn't be part of our search string
+                searchString = search_string_no_space; // We're already adding a space; don't do it again
```
dunhor, line 1015:
```suggestion
                searchString = search_string_no_space; // We're already adding a quote; don't do it again
```

jonwis (approved):

ChrisGuzak (approved):
love this. forming the CreateProcess args is complicated so glad to see this.

PR #506, opened 2025-02-25
Adding a new network header with associated unit tests

keith-horton:
Adding a new network header with associated unit tests.


keith-horton:
@dunhor FYI: I started over with a new topic branch --> new PR. And now I'm no longer getting clang-format problems, and the winrt COM tests are passing :)

dunhor:
> @dunhor FYI: I started over with a new topic branch --> new PR. And now I'm no longer getting clang-format problems, and the winrt COM tests are passing :)

Sorry for the delayed response; been busy with a bunch of other things at the moment.

Regarding clang-format, I took a brief look at your other PR and it looks like you ran clang-format on all files & committed the result. This can have some unintended side effects because different versions of clang-format will format code differently. It's a little unfortunate, but it's not something we have much control over. To help alleviate pains here, we do two things. First, we prefer to use the version of clang-format that ships with Visual Studio for consistency across developers. The second is that the CI machines leverage `git clang-format` to check to see if the change is formatted correctly. This is all described in the [readme](https://github.com/microsoft/wil?tab=readme-ov-file#formatting). The simplest thing to do is to run `git clang-format` yourself for formatting changes. If that's not feasible (e.g. you don't have LLVM installed), you'll have to do a bit more manual labor to ensure you only commit changes you made. Even with all this, it may still not be perfect if the version of Visual Studio on the CI machine isn't quite up to date. This is fine - we can ignore the formatting failure so long as the code is reasonably formatted.

Regarding the COM tests, it's been failing for a while, but typically passes after one (or two) re-runs. It's on my TODO list for when I get the time to dig into it. Hopefully this can be fixed with something as simple as a detour to make it more deterministic.

keith-horton:
> > @dunhor FYI: I started over with a new topic branch --> new PR. And now I'm no longer getting clang-format problems, and the winrt COM tests are passing :)
> 
> Sorry for the delayed response; been busy with a bunch of other things at the moment.
> 
> Regarding clang-format, I took a brief look at your other PR and it looks like you ran clang-format on all files & committed the result. This can have some unintended side effects because different versions of clang-format will format code differently. It's a little unfortunate, but it's not something we have much control over. To help alleviate pains here, we do two things. First, we prefer to use the version of clang-format that ships with Visual Studio for consistency across developers. The second is that the CI machines leverage `git clang-format` to check to see if the change is formatted correctly. This is all described in the [readme](https://github.com/microsoft/wil?tab=readme-ov-file#formatting). The simplest thing to do is to run `git clang-format` yourself for formatting changes. If that's not feasible (e.g. you don't have LLVM installed), you'll have to do a bit more manual labor to ensure you only commit changes you made. Even with all this, it may still not be perfect if the version of Visual Studio on the CI machine isn't quite up to date. This is fine - we can ignore the formatting failure so long as the code is reasonably formatted.
> 
> Regarding the COM tests, it's been failing for a while, but typically passes after one (or two) re-runs. It's on my TODO list for when I get the time to dig into it. Hopefully this can be fixed with something as simple as a detour to make it more deterministic.

Thanks for the explanation!

Yes, this time I just did a git clang-format for edited files.
And the pipeline is happy :)

keith-horton:
@dunhor : well, those COM tests decided to start failing after I clang-formatted network.h. :(

Any thoughts on how to move this PR forward? can I just force that test to try again?

dmachaj:
> @dunhor : well, those COM tests decided to start failing after I clang-formatted network.h. :(
> 
> Any thoughts on how to move this PR forward? can I just force that test to try again?

I think you should disable them to unblock yourself.  Those tests have been far more trouble than they are worth.  The original com_timeout PR had me spend about 5-10x as long writing the tests as I did the non-test code.  And despite that they still aren't reliable enough.  The code they are testing is relatively small and has already been validated so I think that would be ok.

keith-horton:
> > @dunhor : well, those COM tests decided to start failing after I clang-formatted network.h. :(
> > Any thoughts on how to move this PR forward? can I just force that test to try again?
> 
> I think you should disable them to unblock yourself. Those tests have been far more trouble than they are worth. The original com_timeout PR had me spend about 5-10x as long writing the tests as I did the non-test code. And despite that they still aren't reliable enough. The code they are testing is relatively small and has already been validated so I think that would be ok.
> 
> Also, I'm unsure if it is interesting that the two failures are x64+Release. x86 isn't failing; Debug is not failing.

Thanks!
I'll put ifdefs around them so folks can still run the tests locally if they want to verify these objects.

keith-horton:
Looks like clang no longer wants to compile this:

FAILED: tests/app/CMakeFiles/witest.app.dir/__/FileSystemTests.cpp.obj 
C:\PROGRA~1\LLVM\bin\clang-cl.exe  /nologo -TP -DCATCH_CONFIG_FAST_COMPILE -DNOMINMAX -DWIL_FAST_BUILD -DWINAPI_FAMILY=WINAPI_FAMILY_PC_APP -ID:\a\1\s\include -imsvcD:\a\1\s\build\clang32debug\include -imsvcD:\a\1\s\tests\.\workarounds\wrl -imsvcD:\a\1\s\build\clang32debug\vcpkg_installed\x86-windows\include -m32  /DWIN32 /D_WINDOWS /GR- /EHsc /Zi /Ob0 /Od /RTC1 -std:c++17 -MDd /W4 /WX -Wno-missing-field-initializers -Wno-language-extension-token -Wno-c++17-attribute-extensions -Wno-gnu-zero-variadic-macro-arguments -Wno-extra-semi -Wno-self-assign-overloaded -Wno-self-move -mcx16 -fno-delayed-template-parsing /YuD:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/cmake_pch.hxx /FpD:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/./cmake_pch.cxx.pch /FID:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/cmake_pch.hxx /showIncludes /Fotests\app\CMakeFiles\witest.app.dir\__\FileSystemTests.cpp.obj /Fdtests\app\CMakeFiles\witest.app.dir\ -c -- D:\a\1\s\tests\FileSystemTests.cpp
In file included from D:\a\1\s\tests\FileSystemTests.cpp:25:
D:\a\1\s\include\wil/stl.h(216,39): error: identifier '_zv' preceded by whitespace in a literal operator declaration is deprecated [-Werror,-Wdeprecated-literal-operator]
  216 |     constexpr zstring_view operator"" _zv(const char* str, std::size_t len) noexcept
      |                            ~~~~~~~~~~~^~~
      |                            operator""_zv
D:\a\1\s\include\wil/stl.h(221,40): error: identifier '_zv' preceded by whitespace in a literal operator declaration is deprecated [-Werror,-Wdeprecated-literal-operator]
  221 |     constexpr zwstring_view operator"" _zv(const wchar_t* str, std::size_t len) noexcept
      |                             ~~~~~~~~~~~^~~
      |                             operator""_zv
2 errors generated.

keith-horton:
> Looks like clang no longer wants to compile this:
> 
> FAILED: tests/app/CMakeFiles/witest.app.dir/__/FileSystemTests.cpp.obj C:\PROGRA~1\LLVM\bin\clang-cl.exe /nologo -TP -DCATCH_CONFIG_FAST_COMPILE -DNOMINMAX -DWIL_FAST_BUILD -DWINAPI_FAMILY=WINAPI_FAMILY_PC_APP -ID:\a\1\s\include -imsvcD:\a\1\s\build\clang32debug\include -imsvcD:\a\1\s\tests.\workarounds\wrl -imsvcD:\a\1\s\build\clang32debug\vcpkg_installed\x86-windows\include -m32 /DWIN32 /D_WINDOWS /GR- /EHsc /Zi /Ob0 /Od /RTC1 -std:c++17 -MDd /W4 /WX -Wno-missing-field-initializers -Wno-language-extension-token -Wno-c++17-attribute-extensions -Wno-gnu-zero-variadic-macro-arguments -Wno-extra-semi -Wno-self-assign-overloaded -Wno-self-move -mcx16 -fno-delayed-template-parsing /YuD:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/cmake_pch.hxx /FpD:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/./cmake_pch.cxx.pch /FID:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/cmake_pch.hxx /showIncludes /Fotests\app\CMakeFiles\witest.app.dir__\FileSystemTests.cpp.obj /Fdtests\app\CMakeFiles\witest.app.dir\ -c -- D:\a\1\s\tests\FileSystemTests.cpp In file included from D:\a\1\s\tests\FileSystemTests.cpp:25: D:\a\1\s\include\wil/stl.h(216,39): error: identifier '_zv' preceded by whitespace in a literal operator declaration is deprecated [-Werror,-Wdeprecated-literal-operator] 216 | constexpr zstring_view operator"" _zv(const char* str, std::size_t len) noexcept | ~~~~~~~~~~~^~~ | operator""_zv D:\a\1\s\include\wil/stl.h(221,40): error: identifier '_zv' preceded by whitespace in a literal operator declaration is deprecated [-Werror,-Wdeprecated-literal-operator] 221 | constexpr zwstring_view operator"" _zv(const wchar_t* str, std::size_t len) noexcept | ~~~~~~~~~~~^~~ | operator""_zv 2 errors generated.

I think it's related to this:
https://releases.llvm.org/20.1.0/tools/clang/docs/ReleaseNotes.html

The warning -Wdeprecated-literal-operator is now on by default, as this is something that WG21 has shown interest in removing from the language. The result is that anyone who is compiling with -Werror should see this diagnostic. To fix this diagnostic, simply removing the space character from between the operator"" and the user defined literal name will make the source no longer deprecated. This is consistent with CWG2521 <https://cplusplus.github.io/CWG/issues/2521.html>_.

keith-horton:
> > Looks like clang no longer wants to compile this:
> > FAILED: tests/app/CMakeFiles/witest.app.dir/**/FileSystemTests.cpp.obj C:\PROGRA~1\LLVM\bin\clang-cl.exe /nologo -TP -DCATCH_CONFIG_FAST_COMPILE -DNOMINMAX -DWIL_FAST_BUILD -DWINAPI_FAMILY=WINAPI_FAMILY_PC_APP -ID:\a\1\s\include -imsvcD:\a\1\s\build\clang32debug\include -imsvcD:\a\1\s\tests.\workarounds\wrl -imsvcD:\a\1\s\build\clang32debug\vcpkg_installed\x86-windows\include -m32 /DWIN32 /D_WINDOWS /GR- /EHsc /Zi /Ob0 /Od /RTC1 -std:c++17 -MDd /W4 /WX -Wno-missing-field-initializers -Wno-language-extension-token -Wno-c++17-attribute-extensions -Wno-gnu-zero-variadic-macro-arguments -Wno-extra-semi -Wno-self-assign-overloaded -Wno-self-move -mcx16 -fno-delayed-template-parsing /YuD:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/cmake_pch.hxx /FpD:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/./cmake_pch.cxx.pch /FID:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/cmake_pch.hxx /showIncludes /Fotests\app\CMakeFiles\witest.app.dir**\FileSystemTests.cpp.obj /Fdtests\app\CMakeFiles\witest.app.dir\ -c -- D:\a\1\s\tests\FileSystemTests.cpp In file included from D:\a\1\s\tests\FileSystemTests.cpp:25: D:\a\1\s\include\wil/stl.h(216,39): error: identifier '_zv' preceded by whitespace in a literal operator declaration is deprecated [-Werror,-Wdeprecated-literal-operator] 216 | constexpr zstring_view operator"" _zv(const char* str, std::size_t len) noexcept | ~~~~~~~~~~~^~~ | operator""_zv D:\a\1\s\include\wil/stl.h(221,40): error: identifier '_zv' preceded by whitespace in a literal operator declaration is deprecated [-Werror,-Wdeprecated-literal-operator] 221 | constexpr zwstring_view operator"" _zv(const wchar_t* str, std::size_t len) noexcept | ~~~~~~~~~~~^~~ | operator""_zv 2 errors generated.
> 
> I think it's related to this: https://releases.llvm.org/20.1.0/tools/clang/docs/ReleaseNotes.html
> 
> The warning -Wdeprecated-literal-operator is now on by default, as this is something that WG21 has shown interest in removing from the language. The result is that anyone who is compiling with -Werror should see this diagnostic. To fix this diagnostic, simply removing the space character from between the operator"" and the user defined literal name will make the source no longer deprecated. This is consistent with CWG2521 https://cplusplus.github.io/CWG/issues/2521.html_.

Yep - that should be it.
I use ReSharper - and it flagged that exactly. (with the required fix :))

keith-horton:
And we get another clang compiler issue:

FAILED: tests/app/CMakeFiles/witest.app.dir/__/ResourceTests.cpp.obj 
C:\PROGRA~1\LLVM\bin\clang-cl.exe  /nologo -TP -DCATCH_CONFIG_FAST_COMPILE -DNOMINMAX -DWIL_FAST_BUILD -DWINAPI_FAMILY=WINAPI_FAMILY_PC_APP -ID:\a\1\s\include -imsvcD:\a\1\s\build\clang32debug\include -imsvcD:\a\1\s\tests\.\workarounds\wrl -imsvcD:\a\1\s\build\clang32debug\vcpkg_installed\x86-windows\include -m32  /DWIN32 /D_WINDOWS /GR- /EHsc /Zi /Ob0 /Od /RTC1 -std:c++17 -MDd /W4 /WX -Wno-missing-field-initializers -Wno-language-extension-token -Wno-c++17-attribute-extensions -Wno-gnu-zero-variadic-macro-arguments -Wno-extra-semi -Wno-self-assign-overloaded -Wno-self-move -mcx16 -fno-delayed-template-parsing /YuD:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/cmake_pch.hxx /FpD:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/./cmake_pch.cxx.pch /FID:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/cmake_pch.hxx /showIncludes /Fotests\app\CMakeFiles\witest.app.dir\__\ResourceTests.cpp.obj /Fdtests\app\CMakeFiles\witest.app.dir\ -c -- D:\a\1\s\tests\ResourceTests.cpp
In file included from D:\a\1\s\tests\ResourceTests.cpp:4:
D:\a\1\s\include\wil/resource.h(1011,23): error: first argument in call to 'memset' is a pointer to non-trivially copyable type 'wil::unique_struct<ThingToDestroy2, void (ThingToDestroy2::*)() __attribute__((thiscall)), &CATCH2_INTERNAL_TEST_28()::ThingToDestroy2::destroy>' [-Werror,-Wnontrivial-memcall]
 1011 |         RtlZeroMemory(this, sizeof(*this));
      |                       ^
D:\a\1\s\include\wil/resource.h(917,9): note: in instantiation of member function 'wil::unique_struct<ThingToDestroy2, void (ThingToDestroy2::*)() __attribute__((thiscall)), &CATCH2_INTERNAL_TEST_28()::ThingToDestroy2::destroy>::call_init' requested here
  917 |         call_init(use_default_init_fn());
      |         ^
D:\a\1\s\tests\ResourceTests.cpp(847,109): note: in instantiation of member function 'wil::unique_struct<ThingToDestroy2, void (ThingToDestroy2::*)() __attribute__((thiscall)), &CATCH2_INTERNAL_TEST_28()::ThingToDestroy2::destroy>::unique_struct' requested here
  847 |         wil::unique_struct<ThingToDestroy2, decltype(&ThingToDestroy2::destroy), &ThingToDestroy2::destroy> other;
      |                                                                                                             ^
D:\a\1\s\include\wil/resource.h(1011,23): note: explicitly cast the pointer to silence this warning
 1011 |         RtlZeroMemory(this, sizeof(*this));
      |                       ^
1 error generated.

keith-horton:
> And we get another clang compiler issue:
> 
> FAILED: tests/app/CMakeFiles/witest.app.dir/__/ResourceTests.cpp.obj C:\PROGRA~1\LLVM\bin\clang-cl.exe /nologo -TP -DCATCH_CONFIG_FAST_COMPILE -DNOMINMAX -DWIL_FAST_BUILD -DWINAPI_FAMILY=WINAPI_FAMILY_PC_APP -ID:\a\1\s\include -imsvcD:\a\1\s\build\clang32debug\include -imsvcD:\a\1\s\tests.\workarounds\wrl -imsvcD:\a\1\s\build\clang32debug\vcpkg_installed\x86-windows\include -m32 /DWIN32 /D_WINDOWS /GR- /EHsc /Zi /Ob0 /Od /RTC1 -std:c++17 -MDd /W4 /WX -Wno-missing-field-initializers -Wno-language-extension-token -Wno-c++17-attribute-extensions -Wno-gnu-zero-variadic-macro-arguments -Wno-extra-semi -Wno-self-assign-overloaded -Wno-self-move -mcx16 -fno-delayed-template-parsing /YuD:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/cmake_pch.hxx /FpD:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/./cmake_pch.cxx.pch /FID:/a/1/s/build/clang32debug/tests/app/CMakeFiles/witest.app.dir/cmake_pch.hxx /showIncludes /Fotests\app\CMakeFiles\witest.app.dir__\ResourceTests.cpp.obj /Fdtests\app\CMakeFiles\witest.app.dir\ -c -- D:\a\1\s\tests\ResourceTests.cpp In file included from D:\a\1\s\tests\ResourceTests.cpp:4: D:\a\1\s\include\wil/resource.h(1011,23): error: first argument in call to 'memset' is a pointer to non-trivially copyable type 'wil::unique_struct<ThingToDestroy2, void (ThingToDestroy2::_)() **attribute**((thiscall)), &CATCH2_INTERNAL_TEST_28()::ThingToDestroy2::destroy>' [-Werror,-Wnontrivial-memcall] 1011 | RtlZeroMemory(this, sizeof(_this)); | ^ D:\a\1\s\include\wil/resource.h(917,9): note: in instantiation of member function 'wil::unique_struct<ThingToDestroy2, void (ThingToDestroy2::_)() **attribute**((thiscall)), &CATCH2_INTERNAL_TEST_28()::ThingToDestroy2::destroy>::call_init' requested here 917 | call_init(use_default_init_fn()); | ^ D:\a\1\s\tests\ResourceTests.cpp(847,109): note: in instantiation of member function 'wil::unique_struct<ThingToDestroy2, void (ThingToDestroy2::_)() **attribute**((thiscall)), &CATCH2_INTERNAL_TEST_28()::ThingToDestroy2::destroy>::unique_struct' requested here 847 | wil::unique_struct<ThingToDestroy2, decltype(&ThingToDestroy2::destroy), &ThingToDestroy2::destroy> other; | ^ D:\a\1\s\include\wil/resource.h(1011,23): note: explicitly cast the pointer to silence this warning 1011 | RtlZeroMemory(this, sizeof(*this)); | ^ 1 error generated.


This apparently is also in the clang release notes:


[Modified Compiler Flags](https://releases.llvm.org/20.1.0/tools/clang/docs/ReleaseNotes.html#id239)
The -ffp-model option has been updated to enable a more limited set of optimizations when the fast argument is used and to accept a new argument, aggressive. The behavior of -ffp-model=aggressive is equivalent to the previous behavior of -ffp-model=fast. The updated -ffp-model=fast behavior no longer assumes finite math only and uses the promoted algorithm for complex division when possible rather than the less basic (limited range) algorithm.

The -fveclib option has been updated to enable -fno-math-errno for -fveclib=ArmPL and -fveclib=SLEEF. This gives Clang more opportunities to utilize these vector libraries. The behavior for all other vector function libraries remains unchanged.

The -Wnontrivial-memcall warning has been added to warn about passing non-trivially-copyable destination parameter to memcpy, memset and similar functions for which it is a documented undefined behavior. It is implied by -Wnontrivial-memaccess

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+
+    [[nodiscard]] inline bool equals(const ::in6_addr& lhs, const ::in6_addr& rhs) WI_NOEXCEPT
+    {
+        return 0 == ::memcmp(&lhs, &rhs, sizeof(::in6_addr));
+    }
+
+    [[nodiscard]] inline bool not_equals(const ::in6_addr& lhs, const ::in6_addr& rhs) WI_NOEXCEPT
+    {
+        return !::wil::network::equals(lhs, rhs);
+    }
```
jonwis, line 128: Should there also be `operator==`/`operator!=` free functions? Why `equals` and `not_equals`?

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+        bool operator<(const socket_address&) const WI_NOEXCEPT;
+        bool operator>(const socket_address&) const WI_NOEXCEPT;
+
+        void swap(socket_address&) WI_NOEXCEPT;
+        void reset(ADDRESS_FAMILY family = AF_UNSPEC) WI_NOEXCEPT;
+
+        // set_sockaddr overwrites the entire sockaddr in the object (including address family)
+        template <typename T>
+        void set_sockaddr(_In_reads_bytes_(addr_size) const SOCKADDR* addr, T addr_size) WI_NOEXCEPT;
+        void set_sockaddr(const SOCKADDR_IN*) WI_NOEXCEPT;
```
jonwis, line 205: `const SOCKADDR_IN&` - when possible pass by reference. Lets you avoid null checks (or access violations) later. Unless your callers mostly only have pointers...

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+        {
+            m_sockaddr.si_family = AF_INET6;
+            return S_OK;
+        }
+
+        reset(AF_UNSPEC);
+        return E_INVALIDARG;
+    }
+
+    inline HRESULT socket_address::write_address_nothrow(socket_address_wstring& address) const WI_NOEXCEPT
```
jonwis, line 1098: `write_*` suggests writing to a file or output stream. Should this be like "format"? Or "get"?

include/wil/network.h
keith-horton: yeah. callers are almost always going to have pointers to their sockaddr's ... that's how all our socket APIs work :(

include/wil/network.h
keith-horton: ah... 'format'. that semantic works much better. thanks! I'll change this to format_address_*

include/wil/network.h
keith-horton: I didn't want to overload that operator on existing structures (e.g. surprising existing code that might be using == and != with the structures today, but I just changed their behavior if/when they adding a using namespace in their cpp). so I was following something like what winrt.h does.

(I'm trying to remember the other reason I did this)

does this seem completely wrong to use these verbs as opposed to overloading those operators?

keith-horton: hmm. I just played with this. another reason is that in order for operator== to be chosen, someone would needed to have declared using wil::network. which I don't want to require or even recommend. (and seems inconsistent with the other wil headers).

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+    using unique_wsacleanup_call = ::wil::unique_call<decltype(&::WSACleanup), ::WSACleanup>;
+
+    // Calls WSAStartup; returns an RAII object that reverts, the RAII object will resolve to bool 'false' if failed
+    WI_NODISCARD inline ::wil::network::unique_wsacleanup_call WSAStartup_nothrow() WI_NOEXCEPT
+    {
+        WSADATA unused_data{};
+        const auto error{::WSAStartup(WINSOCK_VERSION, &unused_data)};
+        LOG_IF_WIN32_ERROR(error);
+
+        ::wil::network::unique_wsacleanup_call return_cleanup{};
```
jonwis, line 78: You can also say `return unique_wsacleanup_call{ error != 0 };` here - unique_call(bool), when passed "true" arms the callback. Passing "false" returns an unarmed callback.

jonwis (approved):

ChrisGuzak (dismissed):

dunhor:
> Azure Pipelines successfully started running 1 pipeline(s).

Yay, it worked

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+        template <typename T>
+        void set_sockaddr(_In_reads_bytes_(addr_size) const SOCKADDR* addr, T addr_size) WI_NOEXCEPT;
+        void set_sockaddr(const SOCKADDR_IN*) WI_NOEXCEPT;
+        void set_sockaddr(const SOCKADDR_IN6*) WI_NOEXCEPT;
+        void set_sockaddr(const SOCKADDR_INET*) WI_NOEXCEPT;
+        void set_sockaddr(const SOCKET_ADDRESS*) WI_NOEXCEPT;
+#if defined(WIL_ENABLE_EXCEPTIONS)
+        void set_sockaddr(SOCKET);
+        void set_sockaddr(PCWSTR);
+        void set_sockaddr(PCSTR);
```
dunhor, line 212: Why not also call these `reset` (to match the constructors)? AFAICT they all clear out `m_sockaddr`.

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+    typedef WCHAR socket_address_wstring[INET6_ADDRSTRLEN];
+    typedef CHAR socket_address_string[INET6_ADDRSTRLEN];
+
+    class socket_address final
+    {
+    public:
+        constexpr socket_address() WI_NOEXCEPT = default;
+        explicit socket_address(ADDRESS_FAMILY family) WI_NOEXCEPT;
+        template <typename T>
+        explicit socket_address(_In_reads_bytes_(addr_size) const SOCKADDR* addr, T addr_size) WI_NOEXCEPT;
```
dunhor, line 177: Does this need to be a template? Just pass as `size_t`

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+        [[nodiscard]] const SOCKADDR* sockaddr() const WI_NOEXCEPT;
+        [[nodiscard]] const SOCKADDR_IN* sockaddr_in() const WI_NOEXCEPT;
+        [[nodiscard]] const SOCKADDR_IN6* sockaddr_in6() const WI_NOEXCEPT;
+        [[nodiscard]] const SOCKADDR_INET* sockaddr_inet() const WI_NOEXCEPT;
+        [[nodiscard]] const IN_ADDR* in_addr() const WI_NOEXCEPT;
+        [[nodiscard]] const IN6_ADDR* in6_addr() const WI_NOEXCEPT;
+
+        [[nodiscard]] SOCKADDR_STORAGE sockaddr_storage() const WI_NOEXCEPT;
+
+        [[nodiscard]] constexpr int length() const
```
dunhor, line 271:
```suggestion
        [[nodiscard]] constexpr int size() const
```
Length implies enumerability (e.g. an array) whereas this is communicating the size of a structure

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+        // defining iterator_traits allows STL <algorithm> functions to be used with this iterator class.
+        // Notice this is a forward_iterator
+        // - does not support random-access (e.g. vector::iterator)
+        // - does not support bidirectional access (e.g. list::iterator)
+#if WIL_USE_STL || defined(WIL_DOXYGEN)
+        using iterator_category = ::std::forward_iterator_tag;
+#endif
+        using value_type = ::wil::network::socket_address;
+        using difference_type = size_t;
+        using distance_type = size_t;
```
dunhor, line 321: `distance_type` should be signed (e.g. `ptrdiff_t`)

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+        explicit socket_address(PCWSTR, unsigned short port = 0);
+#endif
+
+        ~socket_address() = default;
+        socket_address(const socket_address&) WI_NOEXCEPT = default;
+        socket_address& operator=(const socket_address&) WI_NOEXCEPT = default;
+        socket_address(socket_address&&) WI_NOEXCEPT = default;
+        socket_address& operator=(socket_address&&) WI_NOEXCEPT = default;
+
+        bool operator==(const socket_address&) const WI_NOEXCEPT;
```
dunhor, line 194:
```suggestion
#endif

        bool operator==(const socket_address&) const WI_NOEXCEPT;
```
That's already the default

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+            }
+        }
+
+        ~addr_info_iterator_t() WI_NOEXCEPT = default;
+        addr_info_iterator_t(const addr_info_iterator_t&) WI_NOEXCEPT = default;
+        addr_info_iterator_t& operator=(const addr_info_iterator_t&) WI_NOEXCEPT = default;
+        addr_info_iterator_t(addr_info_iterator_t&&) WI_NOEXCEPT = default;
+        addr_info_iterator_t& operator=(addr_info_iterator_t&&) WI_NOEXCEPT = default;
+
+        const ::wil::network::socket_address& operator*() const WI_NOEXCEPT
```
dunhor, line 343:
```suggestion
        }

        const ::wil::network::socket_address& operator*() const WI_NOEXCEPT
```
Already the default

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+
+        const ::wil::network::socket_address* operator->() const WI_NOEXCEPT
+        {
+            return &m_socket_address;
+        }
+
+        const ::wil::network::socket_address* operator->() WI_NOEXCEPT
+        {
+            return &m_socket_address;
+        }
```
dunhor, line 361: Having both these functions defined is unnecessary given 
```suggestion
        const ::wil::network::socket_address* operator->() const WI_NOEXCEPT
        {
            return &m_socket_address;
        }
```
Non-const function is unnecessary given that the return type of both is identical

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+        // - does not support random-access (e.g. vector::iterator)
+        // - does not support bidirectional access (e.g. list::iterator)
+#if WIL_USE_STL || defined(WIL_DOXYGEN)
+        using iterator_category = ::std::forward_iterator_tag;
+#endif
+        using value_type = ::wil::network::socket_address;
+        using difference_type = size_t;
+        using distance_type = size_t;
+        using pointer = ::wil::network::socket_address*;
+        using reference = ::wil::network::socket_address&;
```
dunhor, line 323:
```suggestion
        using pointer = const ::wil::network::socket_address*;
        using reference = const ::wil::network::socket_address&;
```
Your operators return `const` pointers/references. These types should reflect that

include/wil/network.h
dunhor: While you're at it, update the return types of this and `operator*` to return `pointer` and `reference` respectively

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+                    else
+                    {
+                        m_socket_address.reset();
+                    }
+                }
+            }
+
+            return *this;
+        }
+
```
dunhor, line 408:
```suggestion

```
This is a forward iterator. It should not have `operator+=`.

The fact that this is a `for` loop - and therefore not O(1) - can be surprising and unexpected.

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+    addr_info_iterator_t<T> begin(addr_info_iterator_t<T> addrinfo) WI_NOEXCEPT
+    {
+        return addrinfo;
+    }
+
+    template <typename T>
+    constexpr addr_info_iterator_t<T> end(addr_info_iterator_t<T>) WI_NOEXCEPT
+    {
+        return {};
+    }
```
dunhor, line 437: Move these to be member functions of the iterator type

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+    inline ::wil::unique_addrinfo resolve_local_addresses()
+    {
+        constexpr auto* local_address_name_string = L"";
+        ADDRINFOW* addrResult{};
+        if (::GetAddrInfoW(local_address_name_string, nullptr, nullptr, &addrResult) != 0)
+        {
+            THROW_WIN32(::WSAGetLastError());
+        }
+
+        return ::wil::unique_addrinfo{addrResult};
```
dunhor, line 456: Given the identical code below, is there any benefit to providing an exception-throwing wrapper of `GetAddrInfoW` that these two functions call? The benefit being both code reuse here and function availability to consumers

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+    using winsock_extension_function_table = socket_extension_function_table_t<WINSOCK_EXTENSION_FUNCTION_TABLE>;
+    using rio_extension_function_table = socket_extension_function_table_t<RIO_EXTENSION_FUNCTION_TABLE>;
+
+    template <typename F>
+    struct socket_extension_function_table_t
+    {
+        static socket_extension_function_table_t load() WI_NOEXCEPT;
+
+        ~socket_extension_function_table_t() WI_NOEXCEPT = default;
+
```
dunhor, line 497:
```suggestion

```

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+                ::memcpy_s(&f, sizeof(f), &rhs.f, sizeof(rhs.f));
+            }
+
+            return *this;
+        }
+
+        // Returns true if all functions were loaded, holding a WSAStartup reference
+        WI_NODISCARD explicit operator bool() const WI_NOEXCEPT;
+
+        F f{};
```
dunhor, line 537: Should probably give this a more descriptive name if it is to be public

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+            FAILED_LOG(load_function_pointer(localSocket.get(), WSAID_GETACCEPTEXSOCKADDRS, &table.f.GetAcceptExSockaddrs)) ||
+            FAILED_LOG(load_function_pointer(localSocket.get(), WSAID_TRANSMITFILE, &table.f.TransmitFile)) ||
+            FAILED_LOG(load_function_pointer(localSocket.get(), WSAID_TRANSMITPACKETS, &table.f.TransmitPackets)) ||
+            FAILED_LOG(load_function_pointer(localSocket.get(), WSAID_WSARECVMSG, &table.f.WSARecvMsg)) ||
+            FAILED_LOG(load_function_pointer(localSocket.get(), WSAID_WSASENDMSG, &table.f.WSASendMsg)))
+        {
+            // if any failed to be found, something is very broken
+            // all should load, or all should fail
+            ::memset(&table.f, 0, sizeof(table.f));
+        }
```
dunhor, line 602:
```suggestion
        const auto load_function_pointer = [&](GUID extensionGuid, void* functionPtr) WI_NOEXCEPT {
            constexpr DWORD controlCode{SIO_GET_EXTENSION_FUNCTION_POINTER};
            constexpr DWORD bytes{sizeof(functionPtr)};
            DWORD unused_bytes{};
            const auto error{::WSAIoctl(
                localSocket.get(), controlCode, &extensionGuid, DWORD{sizeof(extensionGuid)}, functionPtr, bytes, &unused_bytes, nullptr, nullptr)};
            return error == 0 ? S_OK : HRESULT_FROM_WIN32(::WSAGetLastError());
        };

        if (FAILED_LOG(load_function_pointer(WSAID_ACCEPTEX, &table.f.AcceptEx)) ||
            FAILED_LOG(load_function_pointer(WSAID_CONNECTEX, &table.f.ConnectEx)) ||
            FAILED_LOG(load_function_pointer(WSAID_DISCONNECTEX, &table.f.DisconnectEx)) ||
            FAILED_LOG(load_function_pointer(WSAID_GETACCEPTEXSOCKADDRS, &table.f.GetAcceptExSockaddrs)) ||
            FAILED_LOG(load_function_pointer(WSAID_TRANSMITFILE, &table.f.TransmitFile)) ||
            FAILED_LOG(load_function_pointer(WSAID_TRANSMITPACKETS, &table.f.TransmitPackets)) ||
            FAILED_LOG(load_function_pointer(WSAID_WSARECVMSG, &table.f.WSARecvMsg)) ||
            FAILED_LOG(load_function_pointer(WSAID_WSASENDMSG, &table.f.WSASendMsg)))
        {
            // if any failed to be found, something is very broken
            // all should load, or all should fail
            ::memset(&table.f, 0, sizeof(table.f));
        }
```
Simplify by capturing the socket in the lambda object

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+        if (INVALID_SOCKET == localSocket.get())
+        {
+            return table;
+        }
+
+        constexpr DWORD controlCode{SIO_GET_MULTIPLE_EXTENSION_FUNCTION_POINTER};
+        constexpr DWORD bytes{sizeof(table.f)};
+        GUID rioGuid = WSAID_MULTIPLE_RIO;
+
+        ::memset(&table.f, 0, bytes);
```
dunhor, line 628:
```suggestion
```
Unnecessary since it'll get zero-initialized on construction

include/wil/network.h
```diff
@@ -0,0 +1,1338 @@
...
+        }
+        return !(*this < rhs);
+    }
+
+    inline void socket_address::swap(socket_address& addr) WI_NOEXCEPT
+    {
+        SOCKADDR_INET tempAddr{};
+        ::memcpy_s(&tempAddr, sizeof(tempAddr), &addr.m_sockaddr, sizeof(addr.m_sockaddr));
+        ::memcpy_s(&addr.m_sockaddr, sizeof(addr.m_sockaddr), &m_sockaddr, sizeof(m_sockaddr));
+        ::memcpy_s(&m_sockaddr, sizeof(m_sockaddr), &tempAddr, sizeof(tempAddr));
```
dunhor, line 814:
```suggestion
        wistd::swap_wil(*this, addr);
```
Much simpler and clearer

include/wil/network.h
keith-horton: Except the socket_address that the caller is calling operator-> might not be const. So I needed a non-const method so it can be called on both.

dunhor: I assume you mean `addr_info_iterator_t`, not `socket_address`? If that's the case, that's not how member functions work. You can call a `const` function on a non-`const` object (or pointer/reference to a maybe `const` object).

keith-horton: ugh. yeah. sorry, I was thinking backwards.

include/wil/network.h
keith-horton: I made it a template since I wanted to avoid unnecessary casting that was likely going to be happening all over.

include/wil/network.h
keith-horton: move operators will automatically be created too?

include/wil/network.h
keith-horton: Great idea.

keith-horton: This helped fix related names too.

include/wil/network.h
keith-horton: yeah. I'm trying to keep this name extremely short, since callers will invoke it like test_table.ft.AcceptEx(...).

And I didn't want to replicate every function in the function table to a method on that object. So they have to just reference that table when they want to call a function.

include/wil/network.h
keith-horton: oh, it really doesn't matter for this trivial class.

include/wil/network.h
keith-horton: that becomes very odd usage. e.g.

test_iterator.begin(test_iterator)

vs.
begin(test_iterator)

include/wil/network.h
dunhor: Something to possibly also consider doing to reduce the typing overhead (and therefore reducing the burden of the name here) is providing an `operator->` that returns a pointer to the table. E.g. your snippet becomes `test_table->AcceptEx(...)`. That's starting to dip your toes into "feature abuse" territory, but IMO it's not _that bad_

Another thing to consider is to derive this type from the table type, so all data members are accessible directly off the object. E.g. your snippet now becomes `test_table.AcceptEx(...)`. Your `memset` and `memcpy` calls become a _little_ bit more complex, but `static_cast` is your friend.

include/wil/network.h
dunhor: Direct usage would never call `test_iterator.begin()` - you'd just use it as an iterator (and you'd default construct to get the end iterator). E.g. see [`std::filesystem::directory_iterator` in the STL](https://en.cppreference.com/w/cpp/filesystem/directory_iterator) which follows this pattern. If you wanted to call something in the `<algorithm>` header, as an example, it would look like:
```c++
auto itr = std::find_if(wil::network::addr_info_iterator(addr), wil::network::addr_info_iterator(), ...);
```
And your for-each code would look the same
```c++
for (auto&& val : wil::network::addr_info_iterator(addr)) { ... }
```

The `begin`/`end` functions are really only for "range"-like algorithms that don't mention the type of the object explicitly.

include/wil/network.h
dunhor: I'll link some resources later once I find it again. The TL;DR is that the specification for which constructors/assignment operators are defined for you when is a _little_ complex, however if you follow the very general "rule of five" (where you either specify all or none of the copy/move constructors, copy/move assignment operators, and the destructor), you'll get "the right thing" 99.9% of the time.

The downside to "well I'll just always specify them just to be safe" is that `= default` is not the same as implicit and this can affect things like whether or not the type is trivial, which can have performance implications.

include/wil/network.h
dunhor: The downside to this approach is that this code is now responsible for proper casting. What if someone calls with a negative signed integer? Or a value larger than a `size_t` can hold? Also, I'm hoping callers are using `sizeof` when calling, which already has type `size_t`

include/wil/network.h
```diff
@@ -0,0 +1,1337 @@
...
+        WSADATA unused_data{};
+        THROW_IF_WIN32_ERROR(::WSAStartup(WINSOCK_VERSION, &unused_data));
+        return {};
+    }
+#endif
+
+    //
+    // utility functions to compare inaddr types
+    //
+    [[nodiscard]] inline bool equals(const ::in_addr& lhs, const ::in_addr& rhs) WI_NOEXCEPT
```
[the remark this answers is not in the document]
keith-horton: The reason is "I forgot WIL had a macro for that" :)

include/wil/network.h
keith-horton: Great ideas. Thanks!

include/wil/network.h
dunhor: As promised...

Here's an overview of "the rule of five" (and "the rule of zero"): https://www.youtube.com/watch?v=juAZDfsaMvY

Here's the chart from that talk about which functions are defined and when, which is a good resource (but is not necessary if you follow the rule of five/zero):
image
And the "blog post" that this image comes from: https://howardhinnant.github.io/classdecl.html

From the Core Guidelines:
[C.20 The rule of zero](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-zero)
[C.21 The rule of five](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-five)

include/wil/network.h
keith-horton: Hmm. something odd is happening then.

const wil::unique_addrinfo test_addr = wil::network::resolve_local_addresses();

// This works with **non-member** begin() and end()
for (const auto& address : addr_info_iterator(test_addr.get()))

// must use **make_range** with member begin() and end()
for (const auto& address : wil::make_range(addr_info_iterator(test_addr.get()), addr_info_iterator{}))

dunhor: Would you mind sharing the definitions of begin/end & the compilation errors you get?

keith-horton: Sure.

    template <typename T>
    class addr_info_iterator_t
    {
...
        addr_info_iterator_t begin(addr_info_iterator_t addrinfo) const WI_NOEXCEPT
        {
            return addrinfo;
        }

        constexpr addr_info_iterator_t end(addr_info_iterator_t) const WI_NOEXCEPT
        {
            return {};
        }
}


C:\Users\khorton\source\repos\wil-network\tests\NetworkTests.cpp(2147,38): error: too few arguments to function call, single argument 'addrinfo' was not specified
 2147 |             for (const auto& address : addr_info_iterator(test_addr.get()))
      |                                      ^
C:\Users\khorton\source\repos\wil-network\include\wil/network.h(370,30): note: 'begin' declared here
  370 |         addr_info_iterator_t begin(addr_info_iterator_t addrinfo) const WI_NOEXCEPT
      |                              ^     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
C:\Users\khorton\source\repos\wil-network\tests\NetworkTests.cpp(2147,38): note: when looking up 'begin' function for range expression of type 'addr_info_iterator' (aka 'addr_info_iterator_t<addrinfoW>')
 2147 |             for (const auto& address : addr_info_iterator(test_addr.get()))
      |                                      ^

            uint32_t count = 0;
            for (const auto& address : addr_info_iterator(test_addr.get())) <<<<<< line 2147
            {
                const auto family = address.family();
                REQUIRE((family == AF_INET || family == AF_INET6));
            }
            for (const auto& address : wil::make_range(addr_info_iterator(test_addr.get()), addr_info_iterator{}))
            {
                const auto family = address.family();
                REQUIRE((family == AF_INET || family == AF_INET6));



these were the non-member functions:

    /*
    template <typename T>
    addr_info_iterator_t<T> begin(addr_info_iterator_t<T> addrinfo) WI_NOEXCEPT
    {
        return addrinfo;
    }

    template <typename T>
    constexpr addr_info_iterator_t<T> end(addr_info_iterator_t<T>) WI_NOEXCEPT
    {
        return {};
    }
    */

duncanhorn: (Ignore the different account...) your member begin/end functions should not accept any arguments. I.e.`begin()` and `end()`

keith-horton: OK - I thought I might need to 'reset the pointer in the iterator to the original head pointer' -- but I did even more testing - reusing iterator objects. And this seems to behave correctly.

Please let me know if I'm missing something.

include/wil/network.h
```diff
@@ -0,0 +1,1337 @@
...
+        ::wil::network::socket_address m_socket_address{};
+    }; // class addr_info_iterator_t
+
+    // begin() and end() enable range-based for loops after resolving a name
+    //
+    // For example:
+    //     wil::unique_addrinfo resolved_addr;
+    //     if (0 == GetAddrInfoW(L"name_to_resolve.xyz", nullptr, nullptr, resolved_addr.addressof()))
+    //     {
+    //         for (const auto& address : wil::network::addr_info_iterator{resolved_addr.get()})
```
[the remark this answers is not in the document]
keith-horton: @dunhor , thoughts?

This is a good point. Maybe I need to create an adapter for the unique_any<> for the wil::unique_addrinfo RAII object (not just managing the raw pointer) - and put begin() and end() on that adapter --> so the range-for loop is given the wil::unique_addrinfo RAII object?

include/wil/network.h
keith-horton: it's handled in the eventual function + they'll get a WI_ASSERT.

    template <typename T>
    void socket_address::reset(_In_reads_bytes_(addr_size) const SOCKADDR* addr, T addr_size) WI_NOEXCEPT
    {
        WI_ASSERT(static_cast<size_t>(addr_size) <= static_cast<size_t>(size()));

        ::memset(&m_sockaddr, 0, size());
        if (addr)
        {
            ::memcpy_s(&m_sockaddr, sizeof(m_sockaddr), addr, addr_size);
        }
    }

include/wil/network.h
```diff
@@ -0,0 +1,1337 @@
...
+
+        void set_address_any() WI_NOEXCEPT;
+        void set_address_any(ADDRESS_FAMILY family) WI_NOEXCEPT;
+        void set_address_loopback() WI_NOEXCEPT;
+        void set_address_loopback(ADDRESS_FAMILY family) WI_NOEXCEPT;
+        void set_address(const IN_ADDR*) WI_NOEXCEPT;
+        void set_address(const IN6_ADDR*) WI_NOEXCEPT;
+
+        #if defined(WIL_ENABLE_EXCEPTIONS)
+        void reset_address(SOCKET);
```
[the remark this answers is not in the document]
keith-horton: Thanks. I think I fixed all of these.

include/wil/network.h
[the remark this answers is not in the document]
keith-horton: Or maybe a better/easier option: implement non-member begin and end functions which take a unique_addrinfo as input and return a addr_info_iterator object. (?)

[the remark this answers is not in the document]
guhetier: Hum, by the time I reached that point of the PR, I forgot this was a simply typedef of a `wil::unique_any` (but is a conceptually a list).
Having a call to make it an iterator makes sense in that case.

I would likely try to go for a range adapter that takes the unique pointer and implement the begin / end functions instead of having them directly implemented on the iterator type (in the spirit of `wil::make_range`).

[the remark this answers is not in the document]
keith-horton: OK - so is your suggestion would be to have:
for (const auto& address : wil::network::make_range(resolved_addr))
or
for (const auto& address : wil::network::make_range(resolved_addr.get()))


instead of:
for (const auto& address : wil::network::addr_info_iterator{resolved_addr.get()})

?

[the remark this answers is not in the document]
dunhor: There's multiple options you can do here. While you _can_ add `begin`/`end` functions for `unique_addrinfo`, as pointed out, that's just a typedef for `unique_any`. That means you'd be forcing consumers to either have or construct an `unique_addrinfo`, which might be inconvenient for them (e.g. if they have a non-owning pointer).

Another option is effectively what you have here: an iterator type that also acts like a range. I call this the `std::filesystem::directory_iterator` pattern as that type does the same thing. This is a nice pattern that works pretty well when you have a resource or "handle" to a resource (OS handle, filesystem path, registry path, etc.) that, under the right conditions, is enumerable. This pattern fits pretty well here IMO.

Another option is to introduce a "range" type (e.g. `addr_info_range`) that has `begin`/`end` functions. While this is an _okay_ pattern, it can make things like calling STL `<algorithm>` functions a bit more tedious. E.g. you need a separate line for the range variable in the below:
```c++
wil::network::addr_info_range range(addr);
auto itr = std::find_if(range.begin(), range.end(), ...);
```
Of course, if the iterator type is also public/user-facing, consumers could still directly construct the iterators, although that's a less common pattern.

include/wil/network.h
```diff
@@ -0,0 +1,1328 @@
...
+            {
+                ::memset(&function_table, 0, sizeof(function_table));
+            }
+            else
+            {
+                ::memcpy_s(&function_table, sizeof(function_table), &rhs.function_table, sizeof(rhs.function_table));
+            }
+
+            return *this;
+        }
```
dunhor, line 499: Consider:

Add move constructor/assignment operator that also takes ownership of the `unique_wsacleanup_call`. This would mostly just buy you one less `WSAStartup` call (since the function pointers are copied), so not essential, but nice to have.

NOTE: You can _technically_ define these as `= default`, and they will _technically_ do the correct thing, however the moved-from type will (potentially) still hold non-null pointers in its table. Nobody should be using a moved-from type, so it _shouldn't_ matter, but if they do, might result in tricky to debug issues.

include/wil/network.h
```diff
@@ -0,0 +1,1328 @@
...
+
+    private:
+        // function table
+        F function_table{};
+
+        // constructed via load()
+        socket_extension_function_table_t() WI_NOEXCEPT :
+            wsa_reference_count{WSAStartup_nothrow()}
+        {
+        }
```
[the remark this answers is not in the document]
dunhor: You'd also probably want to change `load` such that it "deactivates" `wsa_reference_count` on failure. If you do that, you can collapse the `operator bool` into a single implementation that just checks `wsa_reference_count`.

include/wil/network.h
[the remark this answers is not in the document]
guhetier: Thanks for the details, @dunhor.
I wasn't familiar with `std::filesystem::directory_iterator`, if the pattern is in use in the STL, I don't have any complaints about using it here (and the compatibility with `std::ranges` is nice).

include/wil/network.h
```diff
@@ -0,0 +1,1328 @@
...
+    }
+
+    inline bool socket_address::operator>(const ::wil::network::socket_address& rhs) const WI_NOEXCEPT
+    {
+        if (*this == rhs)
+        {
+            return false;
+        }
+        return !(*this < rhs);
```
dunhor, line 792: Is it possible to write something like a `compare` function (that returns negative for less, zero for equal, and positive for greater)? As-is, you're doing one comparison for the price of two here

include/wil/network.h
dunhor: Do you have any example code that would use this constructor/`reset` function, but not use something like `sizeof`? In the tests, I only see two calls to the constructor, both of which use `sizeof` (though both also use either `SOCKADDR_IN` or `SOCKADDR_IN6`, each with their own constructor).

There's also a call to `reset` in the iterator type, however `ai_addrlen` is defined as a `size_t` in all of `ADDRINFOA`, `ADDRINFOW`, and `ADDRINFOEXW`.

include/wil/network.h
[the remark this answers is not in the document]
keith-horton: Thanks for the discussion. @guhetier , I'm going to resolve this. Please ping me if you think there are some better options.

include/wil/network.h
keith-horton: Yep. You're right. This can and should be size_t.
Fixing in the next update.

include/wil/network.h
keith-horton: This is a great idea, thanks! Updated.

include/wil/network.h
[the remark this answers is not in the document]
keith-horton: Will look at this feedback next!
(when I next make some evening time)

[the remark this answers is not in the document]
keith-horton: Yep. I reworked this.

include/wil/network.h
keith-horton: wil::unique_call really does not want its move c'tor called :(

so far I cannot get this to compile using the move c'tors for wil::unique_call. e.g., even this fails to compile

        socket_extension_function_table_t() WI_NOEXCEPT :
            wsa_reference_count{WSAStartup_nothrow()}
        {
            load();
            const auto test = ::wistd::move(wsa_reference_count); // I added this to just test to see if I could force the operator=(unique_call&&) --> but even this fails to compile ... I don't know what I'm missing :(
        }

keith-horton:
> /AzurePipelines run

@dunhor : can you run this command to kick off the CI pipeline? I tried and it said, "Commenter does not have sufficient privileges for PR 506 in repo microsoft/wil"

dunhor:
Thanks Jon.

FYI - I'm seriously contemplating moving to GitHub actions so we don't have this commenting requirement (for however long that lasts)

keith-horton:
> Thanks Jon.
> 
> FYI - I'm seriously contemplating moving to GitHub actions so we don't have this commenting requirement (for however long that lasts)

Thanks guys.

Turns out "git add *" + git clang-format --style file" didn't format these files. "scripts\format-changes.cmd upstream/master" didn't format these files.

But scripts\run-clang-format.cmd seems to have formatted them.

Would you mind restarting the CI pipeline? All other stages looked to have succeeded.

dunhor:
> Turns out "git add *" + git clang-format --style file" didn't format these files. "scripts\format-changes.cmd upstream/master" didn't format these files.

There should be a comment in the main readme regarding this. The TL;DR is that different versions of `clang-format` will format the same code differently. This is why we suggest using `git clang-format` instead (and use it for CI validation) since running `clang-format` on the entire codebase may touch files/lines unrelated to your change which can be a pain to review and merge. Not to mention the fact that it's very specific to the version of `clang-format` running on your machine, which may differ from the version running on someone else's, including the CI machine, which is probably the source of the behavior you see (there's also a version of `clang-format` that ships with Visual Studio, which we prefer to use for consistency, and _may_ play a role here).

That said, as long as you've formatted the code, that's all we really care about. If the CI fails because of inconsistent versions of `clang-format`, you can leave a comment and we can override that check because we know that it can be annoying and inconvenient to try and match things up exactly.

As an aside, one thing I've been wanting to do is throw something like a `git diff` into the pipeline so that the formatting differences are clear to anyone who looks, making an override decision easier _or_ allowing the PR author to jut make the change themselves.

As a super aside, it's been on my wishlist for a while to add a command we can run that will format the code and commit the result. I just haven't had the time to look into it yet.

include/wil/network.h
```diff
@@ -0,0 +1,1372 @@
...
+                }
+                else
+                {
+                    m_socket_address.reset();
+                }
+            }
+            else
+            {
+                m_socket_address.reset();
+            }
```
dunhor, line 363:
```suggestion
            if (m_current_ptr)
            {
                m_current_ptr = m_current_ptr->ai_next;
            }

            if (m_current_ptr)
            {
                m_socket_address.reset(m_current_ptr->ai_addr, m_current_ptr->ai_addrlen);
            }
            else
            {
                m_socket_address.reset();
            }
```

A little simpler to read/follow and trivial for the optimizer to handle.

Alternatively you could probably drop the last `else` since `m_socket_address` should be in an empty state if `m_current_ptr` is null.

include/wil/network.h
```diff
@@ -0,0 +1,1372 @@
...
+
+#if defined(WIL_ENABLE_EXCEPTIONS)
+    // function to capture resolving IP addresses assigned to the local machine, throwing on error
+    // returning an RAII object containing the results
+    inline ::wil::unique_addrinfo resolve_name(PCWSTR name)
+    {
+        ADDRINFOW* addrResult{};
+        if (::GetAddrInfoW(name, nullptr, nullptr, &addrResult) != 0)
+        {
+            THROW_WIN32(::WSAGetLastError());
```
dunhor, line 422: One thing that might be worth considering is adding something like a `THROW_WSA_LAST_ERROR()` macro (and its friends). That would be useful to consumers outside of just this file.

include/wil/network.h
```diff
@@ -0,0 +1,1372 @@
...
+            {
+                ::memset(&function_table, 0, sizeof(function_table));
+            }
+            else
+            {
+                ::memcpy_s(&function_table, sizeof(function_table), &rhs.function_table, sizeof(rhs.function_table));
+            }
+        }
+
+        socket_extension_function_table_t& operator=(const socket_extension_function_table_t& rhs) WI_NOEXCEPT
```
dunhor, line 507: What is the purpose to having assignment operators on this type? You declare the reference count as const, neutering your ability to take ownership, meaning assignment will only ever work if _both_ are properly initialized, but at that point AFAICT you're just overwriting the function table with the same values.

include/wil/network.h
```diff
@@ -0,0 +1,1372 @@
...
+            FAILED_LOG(load_function_pointer(WSAID_CONNECTEX, &function_table.ConnectEx)) ||
+            FAILED_LOG(load_function_pointer(WSAID_DISCONNECTEX, &function_table.DisconnectEx)) ||
+            FAILED_LOG(load_function_pointer(WSAID_GETACCEPTEXSOCKADDRS, &function_table.GetAcceptExSockaddrs)) ||
+            FAILED_LOG(load_function_pointer(WSAID_TRANSMITFILE, &function_table.TransmitFile)) ||
+            FAILED_LOG(load_function_pointer(WSAID_TRANSMITPACKETS, &function_table.TransmitPackets)) ||
+            FAILED_LOG(load_function_pointer(WSAID_WSARECVMSG, &function_table.WSARecvMsg)) ||
+            FAILED_LOG(load_function_pointer(WSAID_WSASENDMSG, &function_table.WSASendMsg)))
+        {
+            // if any failed to be found, something is very broken
+            // all should load, or all should fail
```
dunhor, line 634: This comment is a little confusing for two reasons. The first is that "if any failed to be found, something is very broken" conflicts with "or all should fail" - if the second comment is true, the first comment should be more like "if any failed to be found _after the first function_". The second thing is that if what I said previously is what you intended to say, we probably don't want to `FAILED_LOG` if the first function pointer fails to load if that's an expected outcome.

Alternatively, if your first comment is the correct one, the second comment should probably just be deleted (and the `FAILED_LOG` is okay)

include/wil/network.h
keith-horton: Not a bad idea. I looked at result_macros.h and how GetLastError is supported.
That ended up being a non-trivial # of edits to support that vast array of usage patterns.

So I think I'll defer this (good) idea to a later PR :)

include/wil/network.h
keith-horton: Good point. I've updated the comment to be more precise. These really should never fail at all - but because Winsock has some ancient 3rd party providers that could be installed, they might end up failing to find these extension functions if they have some super-old 3rd party software that doesn't support this.

I think "FAILED_LOG" is correct for this usage.

Happy to hear your thoughts.

include/wil/network.h
keith-horton: Great points. I don't think assignment operators are actually needed.
Will look more into this.

include/wil/network.h
dunhor: Based off the updated comment, I'd agree - the log is fine then

keith-horton:
@dunhor , @jonwis : do we need to kick this through the CI pipeline one more time?

(sorry, I merged with the master branch)

dunhor:
> @dunhor , @jonwis : do we need to kick this through the CI pipeline one more time?
> 
> (sorry, I merged with the master branch)

Kicked it off. If there's no other feedback by the end of the day I'll go ahead and squash the PR

Issue #363, opened 2023-09-20
wil::zwstring_view should have a constructor that takes a winrt::hstring


dmachaj:
When cppwinrt string types are being used `wil::zwstring_view` is unfortunately not a drop-in replacement for `std::wstring_view`.  There is no conversion operator from `winrt::hstring` to `wil::zwstring_view` so compilation fails.

This can be manually worked around by in-place constructing a `wil::zwstring_view` with the `c_str()` and `.size()` values from `winrt::hstring`.  That is ugly, though.  It would be better if it was automatic.

PR #520, opened 2025-04-21
Add noexcept and conversions to zwstring_view

### LINKED ISSUE #446: try_copy_to should be noexcept?
The `wil::com_ptr_t` method `try_copy_to` is supposed to be a non-exceptional way to copy the held COM interface pointer into a raw output pointer (potentially of a different interface type). However, it's not marked `noexcept` which can cause problems when using high warning levels on static analyzers (e.g. see [C26447](https://learn.microsoft.com/en-us/cpp/code-quality/c26447?view=msvc-170) from the C++ Core Guidelines checker). It seems like `try_copy_to` and the transitively called functions in `query_policy` should be marked `noexcept`, as they all return error codes and only interact with other non-exceptional APIs (e.g. `QueryInterface`).

jonwis:
This fixes #446 and #363.

Resolves warnings about missing noexcept for methods that only operate in HRESULTs.

Adds conversion from `winrt::hstring` and `std::filesystem::path` to `wil::zwstring_view`.  Detects the presence of a `.c_str()` and `.size()` member on the incoming type, using them if present. C++20 Concepts would make this much nicer looking.

### DIFF
--- a/include/wil/com.h
+++ b/include/wil/com.h
@@ -161,7 +161,7 @@ namespace details
     class agile_query_policy
     {
     public:
-        inline static HRESULT query(_In_ IAgileReference* ptr, REFIID riid, _COM_Outptr_ void** result)
+        inline static HRESULT query(_In_ IAgileReference* ptr, REFIID riid, _COM_Outptr_ void** result) WI_NOEXCEPT
         {
             WI_ASSERT_MSG(riid != __uuidof(IAgileReference), "Cannot resolve a agile reference to IAgileReference");
             auto hr = ptr->Resolve(riid, result);
@@ -170,7 +170,7 @@ namespace details
         }
 
         template <typename TResult>
-        static HRESULT query(_In_ IAgileReference* ptr, _COM_Outptr_ TResult** result)
+        static HRESULT query(_In_ IAgileReference* ptr, _COM_Outptr_ TResult** result) WI_NOEXCEPT
         {
             static_assert(!wistd::is_same<IAgileReference, TResult>::value, "Cannot resolve a agile reference to IAgileReference");
             return query(ptr, __uuidof(TResult), reinterpret_cast<void**>(result));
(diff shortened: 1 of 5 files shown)


dunhor:
Interesting test failure to keep an eye on that's not related to this change
```
-------------------------------------------------------------------------------
ComApartmentVariable::MultipleApartments
-------------------------------------------------------------------------------
D:\a\1\s\tests\ComApartmentVariableTests.cpp(391)
...............................................................................

D:\a\1\s\tests\ComApartmentVariableTests.cpp(391): FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGSEGV - Segmentation violation signal

```

dunhor:
> and `std::filesystem::path` to `wil::zwstring_view`

Idk how desirable this is. The "stock" `basic_string_view` doesn't try and support this, and if anyone wanted to construct something like it, they could just do `zstring_view(path.native())` etc. It's not immediately clear that it would be harmful, however so I'm not going to think too much about it.

dunhor (approved):

jonwis:
> > and `std::filesystem::path` to `wil::zwstring_view`
> 
> Idk how desirable this is. The "stock" `basic_string_view` doesn't try and support this, and if anyone wanted to construct something like it, they could just do `zstring_view(path.native())` etc. It's not immediately clear that it would be harmful, however so I'm not going to think too much about it.

`path` is just one example of types that have a `.c_str()` but not a `.size()` ... seemed reasonable to support both. I can fix the sample if you think it's an attractive-distractor. :)

dunhor:
> `path` is just one example of types that have a `.c_str()` but not a `.size()`

Well, one "problem" (not actually a problem, just a sub-optimality) is that something like `path` _does_ know its size. I.e. constructing a `zstring_view` with a `path` does an unnecessary `(str|wcs)len`, as opposed to actually calling `path.native()`, which gives back a `std::(w)string` , which does know its size.

That said, if there are other types you are thinking about that have `c_str` but no `size`, it's not that big of a deal

PR #522, opened 2025-05-09
wil::WaitForDebuggerPresent only supports exceptional callers

dmachaj:
I want to use WaitForDebuggerPresent from a component that compiles with exceptions disabled.  I was (unpleasantly) surprised that it is totally unusable without exceptions enabled.

`WaitForDebuggerPresent` and its dependency `GetCurrentProcessExecutionOption` now have throwing+failfasting+returning variants.  They are present outside the exception ifdef in this file.  The logic has been tweaked slightly so that the base methods are nothrow (GetCurrentProcessExecutionOption) or use a templated error policy (WaitForDebuggerPresent).  

There are now some unit test cases.  Nobody sets the regkey for the test binary so it will always return immediately.  This provides coverage that all three variants compile without any build breaks.

I also used the debugger to insert fake failures and confirmed how all three variants behave.  The nothrow uses the default value (do not wait).  The failfast and throw variants do what they should too.

Ran init_all, build_all, and runtests.  Everything is green.  Clang-format was also run.

### DIFF
--- a/include/wil/win32_helpers.h
+++ b/include/wil/win32_helpers.h
@@ -668,28 +668,28 @@ namespace details
         {
             const auto fileName = lastSlash + 1;
             wil::unique_cotaskmem_string keyPath;
-            RETURN_IF_FAILED(wil::str_concat_nothrow<wil::unique_cotaskmem_string>(keyPath,
-                LR"(SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\)", fileName));
+            RETURN_IF_FAILED(wil::str_concat_nothrow<wil::unique_cotaskmem_string>(
+                keyPath, LR"(SOFTWARE\Microsoft\Windows NT\CurrentVersion\Image File Execution Options\)", fileName));
             DWORD value{}, sizeofValue = sizeof(value);
             if (::RegGetValueW(
-                HKEY_LOCAL_MACHINE,
-                keyPath.get(),
-                valueName,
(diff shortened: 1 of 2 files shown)


dmachaj:
Finally :)

jonwis (approved):

include/wil/win32_helpers.h
```diff
@@ -694,31 +729,60 @@ inline DWORD GetCurrentProcessExecutionOption(PCWSTR valueName, DWORD defaultVal
...
-            return; // not configured, don't wait
-        }
+            DWORD configValue{1};
+            if (checkRegistryConfig)
+            {
+                DWORD registryConfigValue{};
+                // err_returncode_policy will continue running after this line on failure.  The default value of zero
+                // will apply in that case so we will still behave reasonably.
+                error_policy::HResult(details::GetCurrentProcessExecutionOptionNoThrow(L"WaitForDebuggerPresent", 0, &registryConfigValue));
+                configValue = registryConfigValue;
```
dunhor, line 746:
```suggestion
                // err_returncode_policy will continue running after this line on failure.  The default value of zero
                // will apply in that case so we will still behave reasonably.
                error_policy::HResult(details::GetCurrentProcessExecutionOptionNoThrow(L"WaitForDebuggerPresent", 0, &configValue));
```
Equivalent since the first line of `details::GetCurrentProcessExecutionOptionNoThrow` assigns the default value, but with less code

dunhor (approved):

PR #533, opened 2025-06-16
register_com_server: Fix incorrect CoRegisterClassObject rclsid argument

forderud:
The implementation is currently calling [`CoRegisterClassObject`](https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-coregisterclassobject) with the default interface ID GUID as `rclsid` argument. It should instead be passing the COM class ID GUID (ClassID) as rclsid argument.

Also, update affected test code to stay in sync.


forderud:
> Neat! Is there some improvement to `winrt::guid_of<>` to make it work with these types? https://github.com/microsoft/cppwinrt/blob/4c9e03fe1e3d7cbfbf742ec237e9af5b30add68a/strings/base_meta.h#L135-L140 is where it's implemented.

`winrt::guid_of<>` is unfortunately poorly documented, but it seems to return the IID of the first interface passed as template argument to the `winrt::implements<..>` base-class. This is different from `__uuidof(T)` which will give you the UUID of the class itself. My impression is therefore that `winrt::guid_of<>` is correct, but just not the right function to call if one wants the Class ID.

sylveon:
[The docs](https://learn.microsoft.com/en-us/uwp/cpp-ref-for-winrt/guid-of) say:
> A helper function template that retrieves the GUID of a runtime class, coclass, or interface.

So it does appear that the C++/WinRT function is broken! It should return the class ID in this case. I think what trips it up is that C++/WinRT doesn't normally expect you to implement COM classes with `winrt::implements`, so it assumes that the presence of `winrt::implements` means you're implementing a WinRT runtime object, in which case it returns the default interface for that object.

I think a fix could be to instead try `__uuidof` first (via SFINAE?) and then fallback to the default interface if that fails. But unfortunately any fix this might lead to subtle runtime changes (as it changes the return value of guid_of without compilation break), so it's a bit risky.

roxk:
> Either way, we can definitely merge this fix in IMO, because this is for _COM_ servers, not WinRT servers (as shown by it not calling `RoRegisterActivationFactories`), so one should expect COM classes to have `__declspec(uuid)` and straight `__uuidof` should always be correct.

While these helpers are indeed for COM server, the classes provided by a COM server doesn't necessarily have to be COM classes. Implementing a COM server with WinRT types is valid and should be supported, so I'd object to merging this PR.

As a person who pushed for using [WinRT for everything in windows](https://github.com/roxk/packaged-oop-winrt-server-app-extension/issues/2#issuecomment-1622693585), only to get disappointed by the fact that widget API needs to support windows 10 so it cannot use WinRT server due to MSIX limitation...These helpers represent my last ditch effort to reduce friction of implementing everything in WinRT. Merging this PR defeats the purpose of these helpers. I want developers using wil use more WinRT, not more COM.

COM support is welcome, but WinRT support is a must. It's the baseline of my contribution.

TLDR: IMO the correct fix is to use `__uuidof` for COM class, and then `winrt::guid_of` for the rest. We can do this in wil instead of cppwinrt to avoid affecting everyone (only developers using `register_com_servers` are affected).

include/wil/cppwinrt_register_com_server.h
```diff
@@ -48,7 +48,7 @@ void register_com_server(std::vector<wil::unique_com_class_object_cookie>& regis
 {
     DWORD registration{};
     winrt::check_hresult(CoRegisterClassObject(
-        winrt::guid_of<T>(), winrt::make<CppWinRTClassFactory<T>>().get(), CLSCTX_LOCAL_SERVER, REGCLS_MULTIPLEUSE, &registration));
+        __uuidof(T), winrt::make<CppWinRTClassFactory<T>>().get(), CLSCTX_LOCAL_SERVER, REGCLS_MULTIPLEUSE, &registration));
```
roxk, line 51: Please do not remove support for `winrt::guid_of`, thanks!

tests/CppWinRTComServerTests.cpp
```diff
@@ -20,15 +20,17 @@ void notifier()
     _comExit.SetEvent();
 }
 
-struct MyServer : winrt::implements<MyServer, winrt::Windows::Foundation::IStringable>
+struct __declspec(uuid("8D0E7037-9855-4659-B48D-F5DBD50F7836")) MyServer
```
roxk, line 23: These are meant to be winrt type types so they wouldn't explicitly define UUID. Please revert this and add `MyComClass` for testing support for COM classes instead, thanks!

roxk (changes requested):

tests/CppWinRTComServerTests.cpp
sylveon: If this is removed, `winrt::guid_of` returns the UUID of `IStringable`, which is definitely not correct behavior for this purpose. `CoCreateClass(__uuidof(ISomeInterface))` should not be valid.

sylveon:
> Implementing a COM server with WinRT types is valid and should be supported, so I'd object to merging this PR.

However, you have to abide by COM rules and have a CLSID. The current implementation does not do that.

tests/CppWinRTComServerTests.cpp
sylveon: > These are meant to be winrt type types so they wouldn't explicitly define UUID.

However, they are being registered to a *COM* server so they need a CLSID. The fact these classes implement WinRT interfaces does not detract from this.

include/wil/cppwinrt_register_com_server.h
forderud: > Please do not remove support for `winrt::guid_of`, thanks!

The WIL `register_com_server` function in question registers COM servers based on CLSID. I do then believe that `__uuidof(T)` would be more appropriate than `winrt::guid_of<T>()` for getting the CLSID for the COM class.

Do you have a concrete example of a COM class where it would be more appropriate to instead use `winrt::guid_of<T>()` for retrieving the CLSID, so that I can better understand your comment?

dunhor:
Going through this I can see valid arguments for both sides. First, some comments...

> C++/WinRT doesn't normally expect you to implement COM classes with winrt::implements, so it assumes that the presence of winrt::implements means you're implementing a WinRT runtime object

Using C++/WinRT for non-WinRT types is quite common, particularly within the Windows codebase itself. @ChrisGuzak has been a major proponent of this usage. That said, it's not incredibly common to `declspec(uuid())` on a class implementation, so the C++/WinRT behavior here seems somewhat reasonable, though IMO this behavior should be limited to just class implementations with a default `exclusiveto` interface. Regardless, the ship has sailed here and I will assume the design to be immutable.

> Implementing a COM server with WinRT types is valid and should be supported

This change doesn't affect that. Really all it's doing is making it more explicit about _which_ GUID to use for the CLSID. You can `__declspec(uuid())` your class definitions and it will still work. This is arguably better for the reasons @sylveon already pointed out in the review since CLSIDs and UUIDs are two distinct things. One could argue that the UUID for an `exclusiveto` interface _might_ be fine to use as a CLSID since it should be unique to the implementation, at least in the same ways CLSIDs are, however AFAICT there's no way to enforce this, e.g. with some `static_assert(winrt::is_exclusiveto<winrt::default_interface<T>>)`.

> this is for COM servers, not WinRT servers

> winrt::guid_of is only valid here when you have a full class defined in metadata with an exclusiveto default_interface. And if you have that, why the heck are you not implementing a WinRT server.

There are valid reasons you would want to register a WinRT class as a COM server since many features of Windows use COM types as currencies. Additionally, I think you mean "useful" not "valid" since the (original) tests pretty clearly demonstrate that it compiles with `IStringable` as the first interface (regardless of how bad of an idea that might be).

> I think a fix could be to instead try __uuidof first (via SFINAE?)

This is my suggestion. I wrote the following code some years ago; perhaps it's time for a revival:

```c++
template <typename T>
struct has_iid
{
    template <typename U = T, std::enable_if_t<std::is_same_v<GUID, std::decay_t<decltype(__uuidof(U))>>, int> = 0>
    static std::true_type invoke(int);

    template <typename U = T>
    static std::false_type invoke(float);

    static constexpr bool value = decltype(invoke(0))::value;
};

template <typename T>
constexpr bool has_iid_v = has_iid<T>::value;
```

You could then `if constexpr` using this type trait to decide whether to use `__uuidof()` or `winrt::guid_of<>`. This is still not the greatest since it's still possible to register your type with a non-`exclusiveto` interface as the CLSID, but at that point you are willingly shooting yourself in the foot.

dunhor:
> It is in the working draft now! It got merged last month. Static reflection is actually coming :)

Oh, interesting. Attribute reflection really sounded like a future thing when first introduced; glad to see it expedited. Still, a template metaprogramming solution is probably more preferable in the near term should someone wish to pursue that route.

> I think the best solution to keep everyone happy would be defer the best way to pick the CLSID to the user by simply switching to taking it as a parameter (would have to remove the fold expression and do registration one-by-one)

That sounds like a reasonable alternative to me. Just make sure to use something like `std::array` so that we can `static_assert` that `sizeof...` the template arguments and the size of the array match. @roxk can still get what he wants with a simple wrapper, something like:
```c++
template <typename... Types>
constexpr auto make_clsid_array()
{
    return std::array<GUID, sizeof...(Types)>{ winrt::guid_of<Types>()... };
}
```

PR #537, opened 2025-07-24
COM server refactor and improvements

sylveon:
I originally only wanted to do the CLSID parameter change, but ended up spotting a lot of potential improvements:
- Moved the class factory somewhere public because it can be useful for more things than just the COM server. Also made it so that the class factory only uses `winrt::no_module_lock` for usage in OOP servers like `register_com_server` because an in-proc class factory does need to increment the module lock.
- Allow callers to alter the extra parameters of `CoRegisterClassObject`
- Added reasons to the `nodiscard`s
- Made use of `REGCLS_SUSPENDED` in the overload that registers multiple instances
- Use `CoAddRefServerProcess` and `CoReleaseServerProcess` in the lock we use - and also turn it purpose specific. This now means that as soon as the ref count hits 0, further activations are suspended to avoid potentially racy conditions
- Now the COM server functionality lives in `cppwinrt_authoring.h`
- Disabled weak references to the class factory, to mirror WRL.

I wanted to look into adding aggregation support to the factory, but backed off for now because cppwinrt's composition support machinery is almost entirely within the `winrt::impl` namespace and assumes `IInspectable`.

The end consumer API is almost identical. Most tests where modified very lightly if at all.

Supplants #533


include/wil/cppwinrt_authoring.h
```diff
@@ -111,6 +111,84 @@ struct single_threaded_rw_property : single_threaded_property<T>
...
+std::vector<unique_com_class_object_cookie> register_com_server(
+    std::array<GUID, sizeof...(Ts)> const& guids, DWORD context = CLSCTX_LOCAL_SERVER, DWORD flags = REGCLS_MULTIPLEUSE)
+{
+    std::vector<wil::unique_com_class_object_cookie> registrations;
+    registrations.reserve(sizeof...(Ts));
+
+    std::size_t i = 0;
+    (registrations.push_back(wil::register_com_server<Ts>(guids[i++], context, flags | REGCLS_SUSPENDED)), ...);
+
+    winrt::check_hresult(CoResumeClassObjects());
```
dunhor, line 186: Should this be conditional on `flags` not containing `REGCLS_SUSPENDED`?

sylveon: Ah I see. Yeah that's a good idea.

dunhor:
@roxk as FYI for the changes in file names/types

dunhor:
> I wonder if it hit an `int 3` or something like that

If it is indeed an issue with coroutine support in Clang, it could also be a bad jump pointer, etc.

sylveon:
It's somehow hitting a compiler-generated `ud2`. When trying to call the AsyncActionCompletedHandler, i.e. after the coroutine is done running.

include/wil/common.h
```diff
@@ -398,6 +398,12 @@ check fails as opposed to the invalid parameter handler that the STL invokes. Th
 // Until we'll have C++17 enabled in our code base, we're falling back to SAL
 #define WI_NODISCARD __WI_LIBCPP_NODISCARD_ATTRIBUTE
 
+#if defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard) >= 201907L
```
dunhor, line 401: This needs to be split out - you can't use an "if defined" on a function-like macro & use it in the same pre-processor directive

dunhor: E.g. something like:
```c++
#if defined(__has_cpp_attribute)
#if __has_cpp_attribute(nodiscard) >= 201907L
#define WI_NODISCARD_REASON(x) [[nodiscard(x)]]
#endif
#endif
#ifndef WI_NODISCARD_REASON
#define WI_NODISCARD_REASON(x) WI_NODISCARD
#endif
```

dunhor (approved):

PR #539, opened 2025-08-02
feat: add make_array_of_(uuid|guid)

roxk:
This is a follow up PR of #537. Requires #537 before merging.

I'm mostly fine with sylveon's improvement in #537 given the current unfortunate situation, but every time I see that we have to [write class name twice during registration](https://github.com/microsoft/wil/pull/537/files#diff-d177dd9786cc83d0a5dd7747436303a146378db2f0b5549ebdd0aace212e0297R403) a little bit of me die inside:
```
auto revoker = wil::register_com_server<MyServer, BuggyServer>({{__uuidof(MyServer), __uuidof(BuggyServer)}});
                                             ^ 1st                            ^ 2nd
```
If you register N class you have to spell out the type 2N times. Not good. I consider [please repeat yourself twice](https://devblogs.microsoft.com/oldnewthing/20220408-00/?p=106438) a sign of bad API design.

So this PR adds `make_array_of_uuid` and `make_array_of_guid` to mitigate the problem:
```
auto revoker = wil::register_com_server(wil::make_array_of_uuid<MyServer, BuggyServer>());
```
It works by wrapping the array in a template and let template parameter deduction works out the rest.

Also disallow using `make_array_of_guid` if the type is a COM class, i.e. `__uuof(T)` is defined. Shamelessly borrowed from @dunhor's `has_iid_v` in previous PR.

This PR does not attempt to address the issue that the implementation type might not have a default interface that is exclusive to it.


include/wil/cppwinrt_authoring.h
```diff
@@ -111,6 +111,126 @@ struct single_threaded_rw_property : single_threaded_property<T>
...
+template <typename... Types>
+struct clsid_array
+{
+    std::array<GUID, sizeof...(Types)> value;
+};
+
+template <typename... Ts>
+WI_NODISCARD_REASON("The classes are unregistered when the returned value is destructed")
+std::vector<unique_com_class_object_cookie> register_com_server(
+    wil::clsid_array<Ts...> const& clsids, DWORD context = CLSCTX_LOCAL_SERVER, DWORD flags = REGCLS_MULTIPLEUSE)
```
sylveon, line 184: I would recommend keeping the std::array parameter and making clsid_array a new overload.

The overload can simply pass it on to the std::array version

roxk: I don't mind that, but let's see what @dunhor vote for.

include/wil/cppwinrt_authoring.h
```diff
@@ -192,6 +198,37 @@ std::vector<unique_com_class_object_cookie> register_com_server(
...
+
+        static constexpr bool value = decltype(invoke(0))::value;
+    };
+
+    template <typename T>
+    constexpr bool has_iid_v = has_iid<T>::value;
+}
+
+template <typename... Types>
+constexpr auto make_array_of_uuid()
```
dunhor, line 220: I think I like the names `make_clsid_array_uuid` and `make_clsid_array_guid` (or perhaps `make_clsid_array_uuidof` and `make_clsid_array_guid_of` to make it more explicit which operation is being used to form the CLSIDs). What do others think here?

include/wil/cppwinrt_authoring.h
dunhor: Gotcha, I see.

roxk: Note that keeping the original `std::array` signature produces `error C2668: 'wil::register_com_server': ambiguous call to overloaded function`. I'd solve it by making `clsid_array`'s ctor explicit. I pushed a commit implementing it.

Btw, removing the original array signature on the other hand, allow both initializer-list style aka `{ _uuidof() }` and `make_array_of_x` to work. So it's literally the best of both world with less code, no explicit ctor etc. I guess keeping the original array signature makes the code "easier" to read?

include/wil/cppwinrt_authoring.h
```diff
@@ -198,6 +192,23 @@ std::vector<unique_com_class_object_cookie> register_com_server(
     return registrations;
 }
 
+template <typename... Types>
+struct clsid_array
+{
+    std::array<GUID, sizeof...(Types)> value;
+    explicit clsid_array(std::array<GUID, sizeof...(Types)> value) : value(std::move(value))
```
roxk, line 199: Note that the pattern here is "copy then move". In practice this value is always move-constructed from temporary, or in more recent standard just materialized without any temporary.

include/wil/cppwinrt_authoring.h
sylveon: Hm, if the original style still works, then I retract my suggestion. Clever.

include/wil/cppwinrt_authoring.h
sylveon: What about `make_clsid_array` and `make_clsid_array_winrt`? I don't particularly like `uuid` vs `guid` because it is not clear what are the differences.

roxk: > What about make_clsid_array and make_clsid_array_winrt? I don't particularly like uuid vs guid because it is not clear what are the differences.

Seems dunhor's suggestion can answer the question? I'd expect user typing `make_clsid_array_` and then seeing `uuidof` vs `guid_of` would be able to tell which is which

include/wil/cppwinrt_authoring.h
dunhor: `std::move` is pointless here. It's a `std::array` and `GUID`s. Accept by `const` reference and just assign and make it `constexpr` while you are at it.

include/wil/cppwinrt_authoring.h
sylveon: The double braces are required for `std::array`. We would've needed `std::initializer_list` for single braces.

dunhor: `std::array` supports aggregate initialization, so single braces should be fine https://godbolt.org/z/o5cPPjYKx

include/wil/cppwinrt_authoring.h
sylveon: Suggestion
```cpp
wil::register_com_server(wil::winrt_clsid_array<Foo, Bar>()); // -> uses winrt::guid_of
wil::register_com_server(wil::com_clsid_array<Buz, Quux>()); // -> uses __uuidof
```

Bit shorter name, and hopefully clearer?

PR #557, opened 2025-11-02
Fixed comparison function for non-network address families

keith-horton:
Found through a separate tool when I manually copied this header to that project.


ChrisGuzak:
is it possible to create a test for this and similar cases?

PR #568, opened 2025-12-05
Fix defining __z breaking libcxx

justanotheranonymoususer:
https://github.com/llvm/llvm-project/blob/cc5b07c7618538b6065f6793524c10bdb34dd510/libcxx/include/__functional/hash.h#L237

### DIFF
--- a/include/wil/result_macros.h
+++ b/include/wil/result_macros.h
@@ -21,7 +21,6 @@
 
 #if !defined(__WIL_MIN_KERNEL) && !defined(WIL_KERNEL_MODE)
 #ifdef __MINGW32__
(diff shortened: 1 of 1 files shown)


justanotheranonymoususer:
I wonder if all these mingw-specific definitions should be moved to `include/wil/common.h` instead.

dunhor:
~~Looks like `result_macros.h` has the same definition. Is this not a problem there?~~

dunhor:
> Looks like `result_macros.h` has the same definition. Is this not a problem there?

Woops, I'm dumb. For some reason I thought this was touching a different file...

PR #575, opened 2025-12-05
Define __wchar_t if not available

justanotheranonymoususer:


include/wil/common.h
```diff
@@ -83,6 +83,12 @@
 #endif
 /// @endcond
 
+/// @cond
+#if defined(__MINGW32__) && !defined(_NATIVE_WCHAR_T_DEFINED)
+#define __wchar_t wchar_t
```
dunhor: Maybe better as a `typedef`?

dunhor: Or are there other places that define this macro that it may otherwise conflict with?

justanotheranonymoususer: Yes, that's better, done

include/wil/common.h
```diff
@@ -940,3 +946,4 @@ using hide_name = void(struct hidden_name);
 
 #endif // __cplusplus
 #endif // __WIL_COMMON_INCLUDED
+
```
dunhor: Looks like clang-format doesn't like this for some reason

include/wil/common.h
justanotheranonymoususer: Now that I think about it, maybe it should be:
```
#if !defined(_MSC_VER) && !defined(_NATIVE_WCHAR_T_DEFINED)
```
mingw is about include headers, and this is a compiler feature, right?

dunhor: According to [MDSN](https://learn.microsoft.com/en-us/cpp/build/reference/zc-wchar-t-wchar-t-is-native-type?view=msvc-170), `_NATIVE_WCHAR_T_DEFINED` is more about whether `wchar_t` maps to `unsigned char` or `__wchar_t`. So I actually don't think checking for it in this macro is correct. Idk how to correctly handle this with other compilers; it might be that `_MSC_VER` is the proper check, but idk since I've already been proven wrong once today on that

justanotheranonymoususer: OK, well, I know that both options work for clang which is what I need. I assume it'd be the same with gcc. So I don't mind having it merged in any form. That was just a pedantic thought. We can change it later if needed.

PR #576, opened 2025-12-05
Allow __declspec(noinline) RESULT_NORETURN for clang

justanotheranonymoususer:


include/wil/result_macros.h
```diff
@@ -395,7 +395,7 @@ WI_ODR_PRAGMA("WIL_FreeMemory", "0")
     template <unsigned int optimizerCounter> \
     inline __declspec(noinline) RESULT_NORETURN RetType MethodName
 #else
-#if defined(_MSC_VER)
+#if defined(_MSC_VER) || defined(__clang__)
```
dunhor: I thought that Clang defines `_MSC_VER` with `-fms-extensions` (which enables support `__declspec` among other things). Is this understanding wrong?

justanotheranonymoususer: It doesn't. I didn't find any define that can detect `-fms-extensions` in a reliable way.

And it works without `-fms-extensions` as well, clang seems to support it in both cases.

justanotheranonymoususer: Yes, maybe, it works on windows but not on linux
https://cpp.godbolt.org/z/8TcP4ojfE

PR #598, opened 2026-01-08
Yet another clang-format workflow fix

dunhor:
In this week's edition, the exit code of `git diff` is preserved and used as the exit code for the script, causing the step to fail


dunhor:
But the good news is, the commenting system is working... !run-clang-format

dunhor:
> > ❌ clang-format failed. Please check the [action logs](https://github.com/microsoft/wil/actions/runs/20808903215) for details.
> 
> Hmmm, you were supposed to succeed... This implies that there are local changes, which is unexpected

Oh my goodness. I'm calling the wrong script 🤦‍♂️

PR #600, opened 2026-01-08
Yet another run-clang-tidy workflow fix

dunhor:
In this episode, I forgot that `git clang-format` uses a non-zero exit code to communicate that it made changes....


dunhor:
> Argh... it's not commenting again

Hmm.... There's still a space at the end of `false `...

PR #609, opened 2026-01-17
Sync with OS

dunhor:
This has been long overdue

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -8321,6 +8321,11 @@ namespace details
     struct lock_proof_traits<std::lock_guard<TMutex>> : exclusive_lock_proof
     {
     };
+
+    template <typename TMutex>
+    struct lock_proof_traits<std::scoped_lock<TMutex>> : exclusive_lock_proof
+    {
+    };
 } // namespace details
 /// @endcond
 
--- a/tests/ResourceTests.cpp
+++ b/tests/ResourceTests.cpp
@@ -8,6 +8,7 @@
 
 // Headers to "light up" functionality in resource.h
 #include <memory>
+#include <shared_mutex>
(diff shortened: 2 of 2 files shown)


include/wil/result.h
```diff
@@ -32,9 +32,11 @@
 #error This header is not supported in kernel-mode.
 #endif
 
-// The updated behavior of running init-list ctors during placement new is proper & correct, disable the warning that requests developers verify they want it
 #pragma warning(push)
+// The updated behavior of running init-list ctors during placement new is proper & correct, disable the warning that requests developers verify they want it
 #pragma warning(disable : 4351)
+// remove this after fixing - volatile access of '<expression>' is subject to /volatile:<iso|ms> setting; consider using __iso_volatile_load/store intrinsic functions
```
dunhor: FYI @jonwis I noticed this when copying the changes

include/wil/stl.h
```diff
@@ -286,15 +286,64 @@ inline namespace literals
...
+template <typename... T>
+overloaded(T...) -> overloaded<T...>;
+
+#endif // __WI_LIBCPP_STD_VER >= 17
+
 } // namespace wil
 
+// This suppression is a temporary workaround to allow libraries built with C++20 to link into binaries built with
+// earlier standard versions such as C++17. This appears to be an issue even when this specialization goes unused
+#ifndef WIL_SUPPRESS_STD_FORMAT_USE
```
dunhor: Also FYI @jonwis - seems that a couple projects build libraries with C++20, but link in a binary built with a lower version. From at least one comment, this is due to dependencies (IIRC ATL was called out) blocking those binaries from using C++20. Hopefully this can be temporary

PR #615, opened 2026-02-06
Fix incompatibility with C++14.

teo-tsirpanis:
`std::scoped_lock` is available in C++17. Using it without a feature guard causes compile errors on earlier C++ versions.

One affected case is the conda-forge distribution of the Azure SDK for C++.

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -8322,7 +8322,7 @@ namespace details
     {
     };
 
-#ifdef __cpp_lib_scoped_lock
+#if __cpp_lib_scoped_lock >= 201703L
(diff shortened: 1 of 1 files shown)


include/wil/resource.h
```diff
@@ -8322,10 +8322,12 @@ namespace details
     {
     };
 
+#ifdef __cpp_lib_scoped_lock
```
dunhor:
```suggestion
#if __cpp_lib_scoped_lock >= 201703L
```
When using feature test macros, compare against an explicit version (in this case, when the type was introduced officially into the standard).

PR #555, opened 2025-10-29
feat: add support for 64-bit registry access in key_access enum

ESiebigteroth:
This PR extends the key_access enum in registry_helpers.h to support 64-bit registry access. The following new values have been added:

- read64: Opens a key for reading from the 64-bit registry.
- readwrite64: Opens a key for reading and writing from the 64-bit registry.

Additionally, the get_access_flags function has been updated to handle these new enum values appropriately.

PR #616, opened 2026-02-12
feat: add support for 32-bit registry access in key_access enum

### LINKED ISSUE #435: Accessing 32-bit registry view from 64-bit process
Some keys in the registry are redirected for 32-bit and 64-bit processes.
Is there a way to access the 32-bit view of the registry from a 64-bit process (and vice versa) using the wil helpers?

Usually you would specify `KEY_WOW64_32KEY` or `KEY_WOW64_64KEY` when opening the key:
https://learn.microsoft.com/en-us/windows/win32/winprog64/accessing-an-alternate-registry-view

Brainy0207:
This PR extends the key_access enum in registry_helpers.h to support 32-bit registry access. The following new values have been added:

- read32: Opens a key for reading from the 64-bit registry.
- readwrite32: Opens a key for reading and writing from the 64-bit registry.

Additionally, the get_access_flags function has been updated to handle these new enum values appropriately.

This is a follow up to #555 which added access to the 64-bit registry for 32-bit processes.
This adds access to the 32-bit registry for 64-bit processes.

Fixes #435

### DIFF
--- a/include/wil/registry_helpers.h
+++ b/include/wil/registry_helpers.h
@@ -65,7 +65,7 @@ namespace reg
     // Access rights for opening registry keys. See https://learn.microsoft.com/en-us/windows/win32/sysinfo/registry-key-security-and-access-rights.
     enum class key_access
     {
-        // Open key for reading.
+        // Open key for reading. Equivalent to KEY_READ.
         read,
 
         // Open key for writing. Equivalent to KEY_WRITE.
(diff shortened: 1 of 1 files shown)


dunhor (approved):

Brainy0207:
It seems #632 will add write and write64 keys to key_access, so we would end up without a write32 key, if both PRs are merged. Depending on which PR gets merged first, the other one should add the write32 key then.

Brainy0207:

Brainy0207:
I have added write32 to match #632

dunhor (approved):

PR #617, opened 2026-02-21
Add support of RemoveDirectoryOptions::RemoveReadOnlyDirectory which remove read-only directories

### LINKED ISSUE #565: RemoveDirectoryRecursive fails if directory has read-only attribute
1. Create some directory and set read-only attribute.
2. Create some sub-directories files in that directory.
3. Call RemoveDirectoryRecursive.
4. Observe ERROR_ACCESS_DENIED.

dimhotepus:
Closes https://github.com/microsoft/wil/issues/565

### DIFF
--- a/include/wil/filesystem.h
+++ b/include/wil/filesystem.h
@@ -200,6 +200,8 @@ enum class RemoveDirectoryOptions
     None = 0,
     KeepRootDirectory = 0x1,
     RemoveReadOnly = 0x2,
+    RemoveReadOnlyFile = RemoveReadOnly,
+    RemoveReadOnlyDirectory = 0x4,
 };
 DEFINE_ENUM_FLAG_OPERATORS(RemoveDirectoryOptions);
 
@@ -315,9 +317,9 @@ inline HRESULT RemoveDirectoryRecursiveNoThrow(
                 // Try a DeleteFile.  Some errors may be recoverable.
                 if (!::DeleteFileW(pathToDelete.get()))
                 {
-                    // Fail for anything other than ERROR_ACCESS_DENIED with option to RemoveReadOnly available
+                    // Fail for anything other than ERROR_ACCESS_DENIED with option to RemoveReadOnlyFile available
                     bool potentiallyFixableReadOnlyProblem =
-                        WI_IsFlagSet(options, RemoveDirectoryOptions::RemoveReadOnly) && ::GetLastError() == ERROR_ACCESS_DENIED;
+                        WI_IsFlagSet(options, RemoveDirectoryOptions::RemoveReadOnlyFile) && ::GetLastError() == ERROR_ACCESS_DENIED;
                     RETURN_LAST_ERROR_IF(!potentiallyFixableReadOnlyProblem);
 
                     // Fail if the file does not have read-only set, likely just an ACL problem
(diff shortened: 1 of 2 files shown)


include/wil/filesystem.h
```diff
@@ -378,7 +378,29 @@ inline HRESULT RemoveDirectoryRecursiveNoThrow(
...
+            if (!::RemoveDirectoryW(path.get()))
+            {
+                // Fail for anything other than ERROR_ACCESS_DENIED with option to RemoveReadOnly available
+                bool potentiallyFixableReadOnlyProblem =
+                    WI_IsFlagSet(options, RemoveDirectoryOptions::RemoveReadOnly) && ::GetLastError() == ERROR_ACCESS_DENIED;
+                RETURN_LAST_ERROR_IF(!potentiallyFixableReadOnlyProblem);
+
+                // Fail if the directory does not have read-only set, likely just an ACL problem
+                DWORD directoryAttr = ::GetFileAttributesW(path.get());
+                RETURN_LAST_ERROR_IF(!WI_IsFlagSet(directoryAttr, FILE_ATTRIBUTE_READONLY));
```
dunhor, line 393: Presumably `GetLastError` won't contain failure because you just called `GetFileAttributesW` right before this (which probably succeeded). And even if it did fail, you probably want the original error from the call to `RemoveDirectoryW`

include/wil/filesystem.h
```diff
@@ -378,7 +378,29 @@ inline HRESULT RemoveDirectoryRecursiveNoThrow(
...
+
+                // Remove read-only flag, setting to NORMAL if completely empty
+                WI_ClearFlag(directoryAttr, FILE_ATTRIBUTE_READONLY);
+                if (directoryAttr == 0)
+                {
+                    directoryAttr = FILE_ATTRIBUTE_NORMAL;
+                }
```
dunhor, line 400: Is this necessary? Presumably the attributes have `FILE_ATTRIBUTE_DIRECTORY` set at a minimum.

dunhor:
@ChrisGuzak and @jonwis for their thoughts.

This jumps out as potentially concerning because it's changing behavior of something that an existing application could theoretically be depending on. The safest route would be to introduce a new flag - `RemoveDirectoryOptions::RemoveReadOnlyDirectory` - and possibly add a new `RemoveDirectoryOptions::RemoveReadOnlyFile` that aliases `RemoveDirectoryOptions::RemoveReadOnly`.

That said, looking around at existing uses of `RemoveDirectoryOptions::RemoveReadOnly`, there doesn't appear to be that many instances outside of test code, so this might just be okay.

dunhor:
The impact is presumably limited to just the directories themselves, so all of the directory _contents_ would still get deleted, so the impact from, say, a security perspective, seems low.

jonwis:
Yeah, I like that additional flag for "... and your little readonly dirs, too!" mode.  Thanks for the change.

include/wil/filesystem.h
dimhotepus: Followed similar approach as when file is deleted: https://github.com/microsoft/wil/blob/2edb7901fc154f14028aef4d29bd8c2faea93785/include/wil/filesystem.h#L328

Set `FILE_ATTRIBUTE_DIRECTORY` instead, looks like `SetFileAttributesW` accepts it for directory.

include/wil/filesystem.h
dimhotepus: Followed the similar approach for file deletion: https://github.com/microsoft/wil/blob/2edb7901fc154f14028aef4d29bd8c2faea93785/include/wil/filesystem.h#L325

dunhor: IMO that seems like a bug for that code path as well. `potentiallyFixableReadOnlyProblem` should logically be something more like:
```c++
                auto err = ::GetLastError();
                bool potentiallyFixableReadOnlyProblem =
                    (WI_IsFlagSet(options, RemoveDirectoryOptions::RemoveReadOnlyDirectory)) &&
                    (err == ERROR_ACCESS_DENIED) &&
                    WI_IsFlagSet(::GetFileAttributesW(path.get()), FILE_ATTRIBUTE_READONLY);
                if (!potentiallyFixableReadOnlyProblem)
                {
                    RETURN_WIN32(err);
                }
```

include/wil/filesystem.h
dunhor: The difference here is that this is a directory - it _must_ have `FILE_ATTRIBUTE_DIRECTORY` set. In contrast, `FILE_ATTRIBUTE_NORMAL` is only valid when used alone. If a file has `FILE_ATTRIBUTE_READONLY` set, it's valid for that to be the _only_ flag set.

That said, this isn't harmful, just an unnecessary check.

PR #620, opened 2026-02-24
Add SAL annotations to safecast.h helper methods

dmachaj:
After adding some usage of these helpers in the OS we received a static analysis bug that was a false positive.  The underlying problem is that these methods lack annotations on the result so the static analysis tooling had to assume the worst.

Add SAL annotations so that static analysis understands that upon success the input and output values are unchanged.

As part of the OS.  This is a port to GitHub after the fact.

### DIFF
--- a/include/wil/safecast.h
+++ b/include/wil/safecast.h
@@ -265,6 +265,7 @@ namespace details
 
 // Unsafe conversion where failure results in fail fast.
 template <typename NewT, typename OldT, wistd::enable_if_t<details::is_supported_unsafe_cast_no_wchar_v<NewT, OldT>, int> = 0>
+_Out_range_(==, var)
 NewT safe_cast_failfast(const OldT var)
 {
     NewT newVar;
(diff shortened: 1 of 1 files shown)


include/wil/safecast.h
```diff
@@ -327,43 +327,43 @@ NewT safe_cast(const OldT var)
...
+_Out_range_(==, var) NewT safe_cast(const OldT var)
 {
     return static_cast<NewT>(var);
 }
 #endif
 
 // This conversion is unsafe, therefore the two parameter version of safe_cast_nothrow must be used
 template <typename NewT, typename OldT, wistd::enable_if_t<details::is_supported_unsafe_cast_v<NewT, OldT>, int> = 0>
-NewT safe_cast_nothrow(const OldT /*var*/)
+_Out_range_(==, _Param_(1)) NewT safe_cast_nothrow(const OldT /*var*/)
```
dunhor, line 338: Just out of curiosity, if this is absent, does the analyzer error still trigger? This is a static assert that should always fire (and if it doesn't, the function doesn't have a `return` statement and should warn). Just curious if the analyzer is smart enough to determine the correct overload.

dunhor (approved):

include/wil/safecast.h
dmachaj: This variant wasn't part of the false positive.  I included the annotation for completeness.  It _probably_ isn't necessary but it also doesn't hurt to include.

PR #627, opened 2026-03-24
Avoid Fail fast for registry key path not found during watcher re-arm

SrinivasNunna:
During a watcher re-arm attempt after key is deleted, applications using AppV layer, try to recreate the delted registry from cache and often fail with ERROR_FILE_NOT_FOUND, thus cauing wil callback to failfast and crash the user application

### DIFF
--- a/include/wil/registry.h
+++ b/include/wil/registry.h
@@ -3234,6 +3234,8 @@ private:
                 break;
 
             case ERROR_KEY_DELETED:
+            case ERROR_FILE_NOT_FOUND:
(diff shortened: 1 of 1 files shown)


ChrisGuzak:
I understand this error codes are emitted by AppV and not the default Win32 API behavior. I also understand changing AppV is hard or not possible. please confirm that in your PR description.

dunhor (approved):

alexlamtest (approved):

jonwis (approved):

ChrisGuzak (approved):

PR #629, opened 2026-04-02
Suppress PREfast false positive (C28167) for RAII lock wrappers

KintoCA:
PREfast warning C28167 ("The function changes the lock state and does not restore it") fires as a false positive on RAII-based locking patterns in resource.h. PREfast's interprocedural analysis cannot track that the lock is released inside a C++ destructor, causing it to incorrectly report that callers of WIL lock guards leave locks in an inconsistent state.

This change adds C28167 to the existing #pragma warning(disable ...) block that already suppresses the related C26135 and C26110 warnings for the same reason.

Testing:
Verified that the suppression eliminates the false-positive PREfast diagnostics without masking any real lock-state issues. No functional behavior changes.

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -22,6 +22,7 @@
 
 #pragma warning(push)
 #pragma warning(disable : 26135 26110) // Missing locking annotation, Caller failing to hold lock
+#pragma warning(disable : 28167) // Prevent false positive by PREfast checker due to inability to detect unlocking in C++ destructors
 #pragma warning(disable : 4714)        // __forceinline not honored
 #pragma warning(disable : 4820)        // padding added after data member
 


KintoCA:
CI is green and this PR is ready for review.

This change suppresses PREfast warning C28167 in resource.h, which appears to be a false positive for the RAII lock pattern because the analyzer does not correctly model lock release through C++ destructors. There is no intended functional or behavioral change.

KintoCA:
@dunhor would you mind taking a look when you have a chance?

This suppresses PREfast C28167 for RAII lock wrappers, alongside the existing
C26135 / C26110 suppressions in the same block. Verified it removes the false
positive without masking real lock-state issues.

Happy to adjust if you’d prefer a different scope or placement.

dunhor (approved):

KintoCA:
@dunhor, 
Thanks for the review!

Since I don’t have merge permissions on this repo, could someone please merge when convenient?

PR #631, opened 2026-04-20
Add set_value_multistring_nothrow for REG_MULTI_SZ writes

### LINKED ISSUE #479: Function missing wil::reg::set_value_multistring_nothrow
I am able to see and use the following functions without any issues:

```
wil::reg::set_value_string_nothrow
wil::reg::set_value_multistring
wil::reg::get_value_multistring_nothrow
```

However, I can only find the following function in the documentation and not in the codebase:
`wil::reg::set_value_multistring_nothrow`

Am I missing something or doing something wrong?

benhillis:
Adds `set_value_multistring_nothrow()` — the missing nothrow counterpart to the existing `set_value_multistring()`.

Two overloads, matching the existing pattern:
```cpp
HRESULT set_value_multistring_nothrow(HKEY key, PCWSTR subkey, PCWSTR value_name, const std::vector<std::wstring>& data);
HRESULT set_value_multistring_nothrow(HKEY key, PCWSTR value_name, const std::vector<std::wstring>& data);
```

Uses `try/CATCH_RETURN()` since `get_multistring_from_wstrings` allocates internally. Gated behind `WIL_USE_STL`.

Tests cover round-trips via both overloads, empty arrays, default value names, `E_ACCESSDENIED` on read-only keys, and cross-verification with the cotaskmem nothrow reader.

Fixes #479

### DIFF
--- a/include/wil/registry.h
+++ b/include/wil/registry.h
@@ -970,6 +970,42 @@ namespace reg
         return ::wil::reg::set_value_expanded_string_nothrow(key, nullptr, value_name, data);
     }
 
+    /**
+     * @brief Writes a REG_MULTI_SZ value from an array of null-terminated strings
+     * @param key An open or well-known registry key
+     * @param subkey The name of the subkey to append to `key`.
+     *        If `nullptr`, then `key` is used without modification.
+     * @param value_name The name of the registry value whose data is to be updated.
+     *        Can be nullptr to write to the unnamed default registry value.
+     * @param data An array of `count` null-terminated strings to write to the specified registry value.
+     *        Each string is marshaled into a contiguous null-terminator-delimited multi-sz string.
+     * @param count The number of strings in `data`
+     * @return HRESULT error code indicating success or failure (does not throw C++ exceptions)
+     */
+    inline HRESULT set_value_multistring_nothrow(
+        HKEY key, _In_opt_ PCWSTR subkey, _In_opt_ PCWSTR value_name, _In_reads_(count) const PCWSTR* data, size_t count) WI_NOEXCEPT
+    {
+        ::wil::unique_process_heap_ptr<wchar_t> buffer;
+        DWORD bufferSizeBytes = 0;
+        RETURN_IF_FAILED(reg_view_details::get_multistring_from_strings_nothrow(data, count, buffer, &bufferSizeBytes));
+        return HRESULT_FROM_WIN32(::RegSetKeyValueW(key, subkey, value_name, REG_MULTI_SZ, buffer.get(), bufferSizeBytes));
+    }
+
+    /**
+     * @brief Writes a REG_MULTI_SZ value from an array of null-terminated strings
+     * @param key An open or well-known registry key
+     * @param value_name The name of the registry value whose data is to be updated.
+     *        Can be nullptr to write to the unnamed default registry value.
+     * @param data An array of `count` null-terminated strings to write to the specified registry value.
+     *        Each string is marshaled into a contiguous null-terminator-delimited multi-sz string.
+     * @param count The number of strings in `data`
+     * @return HRESULT error code indicating success or failure (does not throw C++ exceptions)
+     */
+    inline HRESULT set_value_multistring_nothrow(HKEY key, _In_opt_ PCWSTR value_name, _In_reads_(count) const PCWSTR* data, size_t count) WI_NOEXCEPT
+    {
+        return ::wil::reg::set_value_multistring_nothrow(key, nullptr, value_name, data, count);
+    }
+
 #if defined(__WIL_OBJBASE_H_) || defined(WIL_DOXYGEN)
     /**
      * @brief Writes raw bytes into a registry value under a specified key of the specified type
--- a/include/wil/registry_helpers.h
+++ b/include/wil/registry_helpers.h
@@ -237,6 +237,70 @@ namespace reg
         }
 #endif
 
+        /**
+         * @brief A nothrow translation function marshaling an array of null-terminated strings into a contiguous
+         *        null-terminator-delimited REG_MULTI_SZ buffer allocated on the process heap.
+         * @param data An array of `count` null-terminated strings to marshal into a multi-sz buffer
+         * @param count The number of strings in `data`
+         * @param buffer Receives ownership of the process-heap-allocated wchar_t buffer on success
+         * @param bufferSizeBytes Receives the size, in bytes, of the marshaled buffer (including terminators)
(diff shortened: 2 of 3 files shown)


dunhor:
It feels very weird and unnatural to have a "nothrow" function use exceptions internally -- I'm not aware of any existing place where this is done either. Coupled with the fact that the argument is `std::vector<std::wstring>`, so the caller has to be at least _somewhat_ aware that exceptions exist.

I get _why_ you are doing it, though. OOM and system errors are two very different classes of errors, and it's nice to have control flow that deals with error codes as opposed to exceptions, so I don't necessarily object to the idea. Curious what other folks think.

benhillis:
> not aware of any existing place where this is done either. Coupled with the fact that the argument is `std::vector<std::wstring>`, so the caller has to be at least _somewhat_ aw

I agree, and that's not the way things are typically done in wil, let me think about this a bit more.

include/wil/registry.h
```diff
@@ -884,6 +884,44 @@ namespace reg
...
+     * @return HRESULT error code indicating success or failure (does not throw C++ exceptions)
+     */
+    inline HRESULT set_value_multistring_nothrow(
+        HKEY key, _In_opt_ PCWSTR subkey, _In_opt_ PCWSTR value_name, const ::std::vector<::std::wstring>& data) WI_NOEXCEPT
+    {
+        wchar_t* buffer = nullptr;
+        DWORD bufferSizeBytes = 0;
+        RETURN_IF_FAILED(reg_view_details::get_multistring_from_wstrings_nothrow(data, &buffer, &bufferSizeBytes));
+        const auto hr = HRESULT_FROM_WIN32(::RegSetKeyValueW(key, subkey, value_name, REG_MULTI_SZ, buffer, bufferSizeBytes));
+        ::HeapFree(::GetProcessHeap(), 0, buffer);
```
dunhor: Is there a blocker for using `unique_processheap_ptr` here?

include/wil/registry_helpers.h
```diff
@@ -227,6 +227,57 @@ namespace reg
...
+            {
+                result[offset++] = L'\0';
+            }
+            else
+            {
+                for (const auto& str : data)
+                {
+                    memcpy(result + offset, str.c_str(), str.size() * sizeof(wchar_t));
+                    offset += str.size();
+                    result[offset++] = L'\0';
```
dunhor: `c_str` is guaranteed to return a null terminated string and it's valid to read that null character, so you could simplify slightly by just using `size() + 1`

include/wil/registry_helpers.h
```diff
@@ -227,6 +227,57 @@ namespace reg
...
+                }
+            }
+            result[offset] = L'\0';
+
```
dunhor, line 291: Maybe also `WI_ASSERT(offset == (total_size_bytes / 2))`

dunhor: Mostly sanity check, though there is the possibility for TOCTOU if the caller is modifying the collection during execution since the collection is passed by reference.

include/wil/registry_helpers.h
```diff
@@ -237,6 +237,68 @@ namespace reg
...
+                    return E_INVALIDARG;
+                }
+                total_size_bytes += entry_size;
+            }
+
+            if (data.empty())
+            {
+                // An empty multi-string still requires a leading null plus the final terminator.
+                total_size_bytes = 2 * sizeof(wchar_t);
+            }
```
dunhor: While I've used multi-strings before, I haven't thought too much about this case until now... AFAICT the correct way to represent an empty list is just a single null character: https://devblogs.microsoft.com/oldnewthing/20091008-00/?p=16443

Makes sense as empty strings cannot be present in a multi-string in the general case.

benhillis:
Format fixed.

For the empty multistring: Windows automatically pads REG_MULTI_SZ to 4 bytes minimum. Writing 1 null for `[]` gets padded to 2 nulls, same as `[""]`, making them indistinguishable on read. Current behavior (2 nulls for empty) maintains round-trip compatibility.

include/wil/registry.h
```diff
@@ -970,6 +970,42 @@ namespace reg
...
+     *        If `nullptr`, then `key` is used without modification.
+     * @param value_name The name of the registry value whose data is to be updated.
+     *        Can be nullptr to write to the unnamed default registry value.
+     * @param data An array of `count` null-terminated strings to write to the specified registry value.
+     *        Each string is marshaled into a contiguous null-terminator-delimited multi-sz string.
+     * @param count The number of strings in `data`
+     * @return HRESULT error code indicating success or failure (does not throw C++ exceptions)
+     */
+    inline HRESULT set_value_multistring_nothrow(
+        HKEY key, _In_opt_ PCWSTR subkey, _In_opt_ PCWSTR value_name, _In_reads_(count) const PCWSTR* data, size_t count) WI_NOEXCEPT
```
ChrisGuzak, line 986: span<PCWSTR> could replace data and count. can that be used here?

include/wil/registry_helpers.h
```diff
@@ -233,6 +233,70 @@ namespace reg
...
+         * @return HRESULT error code indicating success or failure (does not throw C++ exceptions)
+         */
+        inline HRESULT get_multistring_from_strings_nothrow(
+            _In_reads_(count) const PCWSTR* data, size_t count, ::wil::unique_process_heap_ptr<wchar_t>& buffer, _Out_ DWORD* bufferSizeBytes) WI_NOEXCEPT
+        {
+            buffer.reset();
+            *bufferSizeBytes = 0;
+
+            size_t total_size_chars = 1; // final terminator
+            for (size_t i = 0; i < count; ++i)
```
ChrisGuzak, line 252: wil::make_range enables use of range based for here I think.

include/wil/registry_helpers.h
```diff
@@ -233,6 +233,70 @@ namespace reg
...
+            {
+                return E_INVALIDARG;
+            }
+            const size_t total_size_bytes = total_size_chars * sizeof(wchar_t);
+
+            ::wil::unique_process_heap_ptr<wchar_t> result{static_cast<wchar_t*>(::HeapAlloc(::GetProcessHeap(), 0, total_size_bytes))};
+            RETURN_IF_NULL_ALLOC(result.get());
+
+            size_t offset = 0;
+            for (size_t i = 0; i < count; ++i)
```
ChrisGuzak, line 278: wil::make_range enables use of range based for here. maybe.

ChrisGuzak (approved):
see my comments. but overall looks good.

PR #633, opened 2026-04-21
Fix set_value template deduction for smart string types

### LINKED ISSUE #624: wil::reg::set_value doesn't work with std::wstring values
When you try to call `wil::reg::set_value` with an `std::wstring` argument you get a compiler error.

e.g.:
```c++
std::wstring my_value = L"Hello";
wil::reg::set_value(HKEY_CURRENT_USER, L"Software\\MyApp", L"MyValue", my_value);
```
results in:
```
static_assert failed: 'Unsupported type for set_value_type'
```

The issue seems to be that `reg_view_t::set_value` deduces the argument type as `const R& value`, which will always be the non-const version:

https://github.com/microsoft/wil/blob/3ecba603d9789435b0661be7b7154d45c195ef98/include/wil/registry_helpers.h#L1182-L1187

It then passes `R` to `reg_value_type_info::set_value_type<R>`, but this expects the `const` version:

https://github.com/microsoft/wil/blob/3ecba603d9789435b0661be7b7154d45c195ef98/include/wil/registry_helpers.h#L1008-L1012

This probably means that all other smart string variants listed there are also affected.

I think the best approach to fix this would be to switch all the affected `set_value_type` specializations to non-const.

benhillis:
`set_value_type<>` specializations for `std::wstring`, `BSTR`, `unique_bstr`, `shared_bstr`, `unique_cotaskmem_string`, and `shared_cotaskmem_string` incorrectly used const-qualified template parameters (e.g. `set_value_type<const std::wstring>`).

Since `reg_view_t::set_value` deduces `R` from `const R& value`, the template parameter `R` is always the non-const type. This caused a `static_assert` failure (`Unsupported type for set_value_type`) when calling `set_value` with any of these types:

\\\cpp
std::wstring value{L"Hello"};
wil::reg::set_value(hkey, L"MyValue", value); // static_assert failure
\\\

The fix removes the `const` qualifier from all six affected specializations so the deduced type matches.

Tests added for `std::wstring` lvalues (non-const, const, and empty) with both opened-key and string-key overloads.

Fixes #624

### DIFF
--- a/include/wil/registry_helpers.h
+++ b/include/wil/registry_helpers.h
@@ -1016,7 +1016,7 @@ namespace reg
             }
 
             template <>
-            constexpr DWORD set_value_type<const ::std::wstring>() WI_NOEXCEPT
+            constexpr DWORD set_value_type<::std::wstring>() WI_NOEXCEPT
             {
                 return REG_SZ;
             }
@@ -1035,13 +1035,13 @@ namespace reg
             }
 
             template <>
-            constexpr DWORD set_value_type<const BSTR>() WI_NOEXCEPT
+            constexpr DWORD set_value_type<BSTR>() WI_NOEXCEPT
             {
                 return REG_SZ;
             }
 
             template <>
-            constexpr DWORD set_value_type<const ::wil::unique_bstr>() WI_NOEXCEPT
+            constexpr DWORD set_value_type<::wil::unique_bstr>() WI_NOEXCEPT
             {
                 return REG_SZ;
             }
@@ -1056,7 +1056,7 @@ namespace reg
             }
 
             template <>
-            constexpr DWORD set_value_type<const ::wil::shared_bstr>() WI_NOEXCEPT
+            constexpr DWORD set_value_type<::wil::shared_bstr>() WI_NOEXCEPT
             {
                 return REG_SZ;
             }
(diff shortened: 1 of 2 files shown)


include/wil/registry_helpers.h
dunhor, line 1184:
```suggestion
                _In_opt_ PCWSTR subkey, _In_opt_ PCWSTR value_name, const R& value, DWORD type = reg_value_type_info::set_value_type<wistd::remove_cv_t<R>>()) const
```
Eliminates the problem/question of it being cv-qualified entirely (although we know `R` can't be `const` in this case, but it might be `volatile`).

Or even better.
```c++
            template <typename T>
            struct set_value_type_t
            {
                static_assert(!wistd::is_same_v<T, T>, "Unsupported type for set_value_type");
            };
            // Specialize the others with 'constexpr DWORD value = ...'

            template <typename T = void>
            constexpr DWORD set_value_type() WI_NOEXCEPT
            {
                return set_value_type_t<wistd::remove_cv_t<T>>::value;
            }
```
And that eliminates the need for the caller to do the correct type shenanigans. The function can still be fully specialized if needed in the future (all existing uses should be able to use the type).

dunhor (approved):

dunhor (approved):

dunhor (approved):

dunhor:
Feel free to publish as "ready for review" if this change is ready

dunhor (approved):

ChrisGuzak (approved):

PR #634, opened 2026-05-09
Add make_bstr*(std::wstring_view)

DrusTheAxe:
WIL only provides only PCWSTR overloads:

* make_bstr_nothrow(PCWSTR source)
* make_bstr_failfast(PCWSTR source)
* inline wil::unique_bstr make_bstr(PCWSTR source);

This PR adds `wstring_view` overloads calling `SysAllocStringLen` with the view's data and size directly, correctly handling non-null-terminated string views

* make_bstr_nothrow(std::wstring_view source)
* make_bstr_failfast(std::wstring_view source)
* make_bstr(std::wstring_view source)

This works for non-null-terminated views (substrings, slices), avoids unnecessary and dangerous wcslen walk (we know the wstring_view's length), and is a strict superset of the existing API.

NOTE: A `PCWSTR` implicitly converts to `wstring_view`, so these new overloads could replace the existing ones once C++17 is the floor. WIL still supports older standards, so these additional overloads is safer.

### DIFF
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -23,8 +23,8 @@
 #pragma warning(push)
 #pragma warning(disable : 26135 26110) // Missing locking annotation, Caller failing to hold lock
 #pragma warning(disable : 28167) // Prevent false positive by PREfast checker due to inability to detect unlocking in C++ destructors
-#pragma warning(disable : 4714) // __forceinline not honored
-#pragma warning(disable : 4820) // padding added after data member
+#pragma warning(disable : 4714)  // __forceinline not honored
+#pragma warning(disable : 4820)  // padding added after data member
 
 #ifndef __WIL_RESOURCE
 #define __WIL_RESOURCE
--- a/include/wil/resource.h
+++ b/include/wil/resource.h
@@ -23,8 +23,8 @@
 #pragma warning(push)
 #pragma warning(disable : 26135 26110) // Missing locking annotation, Caller failing to hold lock
 #pragma warning(disable : 28167) // Prevent false positive by PREfast checker due to inability to detect unlocking in C++ destructors
-#pragma warning(disable : 4714)        // __forceinline not honored
-#pragma warning(disable : 4820)        // padding added after data member
+#pragma warning(disable : 4714) // __forceinline not honored
+#pragma warning(disable : 4820) // padding added after data member
 
 #ifndef __WIL_RESOURCE
 #define __WIL_RESOURCE
--- a/include/wil/stl.h
+++ b/include/wil/stl.h
@@ -132,6 +132,28 @@ inline PCWSTR str_raw_ptr(const std::wstring& str)
     return str.c_str();
 }
 
+#if defined(__WIL_OLEAUTO_H_)
+#if defined(_STRING_VIEW_) || defined(__cpp_lib_string_view)
+// Create wil::unique_bstr from std::wstring_view (regardless if not null terminated, or if it contains embedded nulls)
+inline wil::unique_bstr make_bstr_nothrow(std::wstring_view source) noexcept
+{
+    return wil::unique_bstr(::SysAllocStringLen(source.data(), static_cast<UINT>(source.size())));
+}
+inline wil::unique_bstr make_bstr_failfast(std::wstring_view source) noexcept
+{
+    return wil::unique_bstr(FAIL_FAST_IF_NULL_ALLOC(::SysAllocStringLen(source.data(), static_cast<UINT>(source.size()))));
+}
+#ifdef WIL_ENABLE_EXCEPTIONS
+inline wil::unique_bstr make_bstr(std::wstring_view source)
+{
+    wil::unique_bstr result(make_bstr_nothrow(source));
+    THROW_IF_NULL_ALLOC(result);
+    return result;
+}
+#endif // WIL_ENABLE_EXCEPTIONS
+#endif // defined(_STRING_VIEW_) || defined(__cpp_lib_string_view)
+#endif // defined(__WIL_OLEAUTO_H_)
+
 #if __cpp_lib_string_view >= 201606L
 
 /**
--- a/include/wil/stl.h
+++ b/include/wil/stl.h
@@ -132,16 +132,23 @@ inline PCWSTR str_raw_ptr(const std::wstring& str)
     return str.c_str();
 }
 
+#if __cpp_lib_string_view >= 201606L
+
 #if defined(__WIL_OLEAUTO_H_)
-#if defined(_STRING_VIEW_) || defined(__cpp_lib_string_view)
 // Create wil::unique_bstr from std::wstring_view (regardless if not null terminated, or if it contains embedded nulls)
 inline wil::unique_bstr make_bstr_nothrow(std::wstring_view source) noexcept
 {
+    if (source.size() > 0xffffffffUL /*UINT32_MAX*/)
+    {
+        return wil::unique_bstr{};
+    }
     return wil::unique_bstr(::SysAllocStringLen(source.data(), static_cast<UINT>(source.size())));
 }
 inline wil::unique_bstr make_bstr_failfast(std::wstring_view source) noexcept
 {
-    return wil::unique_bstr(FAIL_FAST_IF_NULL_ALLOC(::SysAllocStringLen(source.data(), static_cast<UINT>(source.size()))));
+    auto result(make_bstr_nothrow(source));
+    FAIL_FAST_IF_NULL_ALLOC(result);
+    return result;
 }
 #ifdef WIL_ENABLE_EXCEPTIONS
 inline wil::unique_bstr make_bstr(std::wstring_view source)
@@ -151,10 +158,7 @@ inline wil::unique_bstr make_bstr(std::wstring_view source)
     return result;
 }
 #endif // WIL_ENABLE_EXCEPTIONS
-#endif // defined(_STRING_VIEW_) || defined(__cpp_lib_string_view)
 #endif // defined(__WIL_OLEAUTO_H_)
-
-#if __cpp_lib_string_view >= 201606L
 
 /**
     zstring_view. A zstring_view is identical to a std::string_view except it is always nul-terminated (unless empty).
--- a/tests/StlTests.cpp
+++ b/tests/StlTests.cpp
@@ -68,7 +68,6 @@ TEST_CASE("StlTests::TestBstrAllocator", "[stl][bstr][string_view]")
 {
     std::string_view stlStringView_empty;
     const wil::unique_bstr bstrEmpty{wil::make_bstr_nothrow(stlStringView_empty)};
-    REQUIRE(bstrEmpty.data() == nullptr);
     REQUIRE(bstrEmpty.get() == nullptr);
 
     std::string_view stlStringView_fromLiteral{L"abc"};
--- a/tests/StlTests.cpp
+++ b/tests/StlTests.cpp
@@ -72,8 +72,6 @@ TEST_CASE("StlTests::TestBstrAllocator", "[stl][bstr][string_view]")
 
     std::string_view stlStringView_fromLiteral{L"abc"};
     const wil::unique_bstr bstrFromLiteral{wil::make_bstr_nothrow(stlStringView_fromLiteral)};
-    REQUIRE(!bstrFromLiteral.empty());
-    REQUIRE(bstrFromLiteral.data() != nullptr);
     REQUIRE(bstrFromLiteral.get() != nullptr);
     REQUIRE(wcslen(bstrFromLiteral.get()) == 3);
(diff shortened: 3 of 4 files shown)


include/wil/stl.h
```diff
@@ -132,6 +132,28 @@ inline PCWSTR str_raw_ptr(const std::wstring& str)
     return str.c_str();
 }
 
+#if defined(__WIL_OLEAUTO_H_)
+#if defined(_STRING_VIEW_) || defined(__cpp_lib_string_view)
+// Create wil::unique_bstr from std::wstring_view (regardless if not null terminated, or if it contains embedded nulls)
+inline wil::unique_bstr make_bstr_nothrow(std::wstring_view source) noexcept
```
dmachaj, line 139: For bonus points - there is also the `wil::zwstring_view` overloads that enforce null termination but are ortherwise just a string_view.

dmachaj (approved):

include/wil/stl.h
```diff
@@ -132,6 +132,28 @@ inline PCWSTR str_raw_ptr(const std::wstring& str)
...
+#if defined(__WIL_OLEAUTO_H_)
+#if defined(_STRING_VIEW_) || defined(__cpp_lib_string_view)
+// Create wil::unique_bstr from std::wstring_view (regardless if not null terminated, or if it contains embedded nulls)
+inline wil::unique_bstr make_bstr_nothrow(std::wstring_view source) noexcept
+{
+    return wil::unique_bstr(::SysAllocStringLen(source.data(), static_cast<UINT>(source.size())));
+}
+inline wil::unique_bstr make_bstr_failfast(std::wstring_view source) noexcept
+{
+    return wil::unique_bstr(FAIL_FAST_IF_NULL_ALLOC(::SysAllocStringLen(source.data(), static_cast<UINT>(source.size()))));
```
jonwis: Consider implementing this in terms of the nothrow variant, rather than repeating it here.

jonwis (approved):

include/wil/stl.h
dunhor: Can `basic_zstring_view` decay to `basic_string_view`? Or is there something special preventing object slicing?

include/wil/stl.h
```diff
@@ -132,6 +132,28 @@ inline PCWSTR str_raw_ptr(const std::wstring& str)
     return str.c_str();
 }
 
+#if defined(__WIL_OLEAUTO_H_)
+#if defined(_STRING_VIEW_) || defined(__cpp_lib_string_view)
```
dunhor: In the general case, this is a problematic check because `<version>` also defines `__cpp_lib_string_view` even though `<string_view>` may not have been included. That said, this is the `stl.h` header, so it already (1) assumes STL use and (2) already includes `<string_view>` with the correct guards, so this concern is moot, but it does make the `_STRING_VIEW_` check unnecessary. Better yet if you compare to a specific release. E.g. this could ideally become:
```suggestion
#if __cpp_lib_string_view >= 201606L
```
... which means you can just move this definition a few lines lower since this check is already present.

include/wil/stl.h
dmachaj: It might.  I didn't actually check and just _assumed_ a specialization would be needed.

include/wil/stl.h
```diff
@@ -132,6 +132,28 @@ inline PCWSTR str_raw_ptr(const std::wstring& str)
     return str.c_str();
 }
 
+#if defined(__WIL_OLEAUTO_H_)
```
dunhor, line 137: `stl.h` is not like `resource.h` and cannot be included multiple times to pick up new definitions, so this will be fragile and subject to include order dependencies. That said, trying to put this in `resource.h` may not be the best option either. In theory, there could be a `_STRING_VIEW_` check, however I'm not a fan of these since (1) this is an implementation detail and might change, (2) might break if someone uses something other than the MSVC STL, and (3) isn't compatible with modules. We could also add a conditional include such as:
```c++
#if WIL_USE_STL && (__WI_LIBCPP_STD_VER >= 17) && WI_HAS_INCLUDE(<string_view>, 1)
#include <string_view>
#endif
```
and then later
```c++
#if WIL_USE_STL && (__cpp_lib_string_view >= 201606L)
// definition goes here
#endif
```
which eliminates the "is `string_view` included already" check (the purpose of `WIL_USE_STL`), however given how widely used `resource.h` is used, the probability of unintentional side effects is high.

Another option is to add a function to `resource.h` that accepts a template argument that is constrained on [`is_string_view_like `](https://github.com/microsoft/wil/blob/37554797d5a1b958a3369836d539220fcb5a31ae/include/wil/common.h#L1086), so you don't have to bother with proper guards.

include/wil/stl.h
```diff
@@ -132,6 +132,28 @@ inline PCWSTR str_raw_ptr(const std::wstring& str)
     return str.c_str();
 }
 
+#if defined(__WIL_OLEAUTO_H_)
+#if defined(_STRING_VIEW_) || defined(__cpp_lib_string_view)
+// Create wil::unique_bstr from std::wstring_view (regardless if not null terminated, or if it contains embedded nulls)
+inline wil::unique_bstr make_bstr_nothrow(std::wstring_view source) noexcept
+{
+    return wil::unique_bstr(::SysAllocStringLen(source.data(), static_cast<UINT>(source.size())));
```
chwarr, line 145: Minor: this will silently truncate a string over 4 GiB. Should this instead fail to create the bstr?

include/wil/stl.h
jonwis: https://github.com/microsoft/wil/blob/4ea066a7bd8a3f41400e7c0bf63175316678bdd6/include/wil/stl.h#L145-L149

So yeah.  You could accept `std::basic_string_view<wchar_t> source` here and that's the only variant you need.

include/wil/stl.h
jonwis: Yes, probably.

include/wil/stl.h
DrusTheAxe: We've already got the check for __cpp_lib_string_view just below so I'll move make_bstr() below that #if rather than before

Ordering should be fine as general practice is to include most generic to most specific e.g. OS -> Language -> Libraries -> Local

```
#include <windows.h>
#include <oleauto.h>
#include <string>
#include <wil/stl.h>
#include "my_project_stuff.h"
```

include/wil/stl.h
DrusTheAxe: Searching for static_cast<> handling I find several such cases e.g. make_unique_string_nothrow<unique_hstring>if length parameter > UINT32_MAX we silently cast to UINT32. Likewise, string_maker::make(), wistd::enable_if_t<> and a few other places.

I'll trap it here in this new BSTR function and file a bug for the others

DrusTheAxe: FYI static_cast<UINT>() silently truncates >4GB to UINT32 _MAX [#651](https://github.com/microsoft/wil/issues/651)

include/wil/stl.h
DrusTheAxe: This would mean in

```
inline wil::unique_bstr make_bstr_nothrow(std::wstring_view source) noexcept
...
inline wil::unique_bstr make_bstr_failfast(std::wstring_view source) noexcept
...
inline wil::unique_bstr make_bstr(std::wstring_view source)
```

replace `std::wstring_view` with `std::basic_string_view<wchar_t>` ?

include/wil/stl.h
dunhor: The problem here is the check for `__WIL_OLEAUTO_H_` - this pattern is historically _really_ fragile and we/I have spent a decent amount of energy reducing these types of dependencies.

Side note: You typically want to include STL/language headers _before_ OS headers because the OS headers (unfortunately) use reserved names (e.g. leading underscores followed by capital letters). There's also macros like `min`/`max`, although the STL headers have been hardened pretty well against those two macros.

include/wil/stl.h
dunhor: `wstring_view` is just an alias for `basic_string_view<wchar_t>`, so the two would be identical.

Any benefit from using `basic_string_view` directly would be having the ability to allow different traits types, but that's probably not really worth it.

include/wil/stl.h
DrusTheAxe: The string_view conditionality was simple as there was already a test for it right after my new bstr functions, so I just moved them below.

What do you suggest we do for the OLEAUTO check? Some folks using STL won't be using COM and won't have OLE headers with BSTR type and functions defined. I'm not seeing a good answer, with the current as the least-bad. Got a better idea?

Side note: Good point, though same diff here -- OS+Language (regardless of order) are generally #include'd before other libraries (like WIL).

jonwis (approved):

include/wil/stl.h
DrusTheAxe: Agreed. Thanks

include/wil/stl.h
dunhor: The best suggestion I have was (kind of buried) in my original comment - move to `resource.h`, which was designed to be includable multiple times, and then accept `is_string_view_like` as the argument instead of explicitly mentioning `std::wstring_view` by name.

Otherwise, you could re-work this header so that it can be included multiple times, similar to `resource.h` and then move this definition to the bottom inside of its own guard, however that has other implications, particularly w.r.t compile times, though I'm not familiar with how much of a cost that might be; `stl.h` is relatively small and not super widely used, so it may be fine...

This is also something that can be changed in the future in a mostly non-breaking way if needed, so I'm not going to worry too much about it, and it looks like @jonwis has signed off, so unless you feel compelled to attempt the `resource.h` approach, I'd say it's good enough for now.

include/wil/resource.h
```diff
@@ -23,8 +23,8 @@
 #pragma warning(push)
 #pragma warning(disable : 26135 26110) // Missing locking annotation, Caller failing to hold lock
 #pragma warning(disable : 28167) // Prevent false positive by PREfast checker due to inability to detect unlocking in C++ destructors
-#pragma warning(disable : 4714)        // __forceinline not honored
-#pragma warning(disable : 4820)        // padding added after data member
+#pragma warning(disable : 4714) // __forceinline not honored
+#pragma warning(disable : 4820) // padding added after data member
```
dunhor:
```suggestion
#pragma warning(disable : 4714)  // __forceinline not honored
#pragma warning(disable : 4820)  // padding added after data member
```
To make clang-format happy

tests/StlTests.cpp
```diff
@@ -64,6 +64,23 @@ struct CustomNoncopyableString
     }
 };
 
+TEST_CASE("StlTests::TestBstrAllocator", "[stl][bstr][string_view]")
+{
+    std::string_view stlStringView_empty;
+    const wil::unique_bstr bstrEmpty{wil::make_bstr_nothrow(stlStringView_empty)};
+    REQUIRE(bstrEmpty.empty());
```
dunhor:
```suggestion
```
Not a function on the type.

tests/StlTests.cpp
```diff
@@ -64,6 +64,23 @@ struct CustomNoncopyableString
     }
 };
 
+TEST_CASE("StlTests::TestBstrAllocator", "[stl][bstr][string_view]")
+{
+    std::string_view stlStringView_empty;
+    const wil::unique_bstr bstrEmpty{wil::make_bstr_nothrow(stlStringView_empty)};
+    REQUIRE(bstrEmpty.empty());
+    REQUIRE(bstrEmpty.data() == nullptr);
```
dunhor:
```suggestion
```
Same

tests/StlTests.cpp
```diff
@@ -64,6 +64,23 @@ struct CustomNoncopyableString
...
 
+TEST_CASE("StlTests::TestBstrAllocator", "[stl][bstr][string_view]")
+{
+    std::string_view stlStringView_empty;
+    const wil::unique_bstr bstrEmpty{wil::make_bstr_nothrow(stlStringView_empty)};
+    REQUIRE(bstrEmpty.empty());
+    REQUIRE(bstrEmpty.data() == nullptr);
+    REQUIRE(bstrEmpty.get() == nullptr);
+
+    std::string_view stlStringView_fromLiteral{L"abc"};
```
dunhor:
```suggestion
    std::wstring_view stlStringView_fromLiteral{L"abc"};
```

tests/StlTests.cpp
```diff
@@ -64,6 +64,23 @@ struct CustomNoncopyableString
...
+    const wil::unique_bstr bstrFromLiteral{wil::make_bstr_nothrow(stlStringView_fromLiteral)};
+    REQUIRE(!bstrFromLiteral.empty());
+    REQUIRE(bstrFromLiteral.data() != nullptr);
```
dunhor:
```suggestion
```

DrusTheAxe:
Incorporated feedback and fixed tests. Need approval of latest commits

jonwis (approved):

dunhor (approved):

PR #641, opened 2026-06-10
Accept raw REGSAM access masks in open/create key helpers

benhillis:
Adds an `access_mask` type accepted by the open/create key helpers, implicitly constructible from `key_access` or a raw REGSAM, so callers can pass custom access rights. Includes Catch2 `[registry]` tests (and compile-time static_asserts) passing in normal & noexcept configs. Part of enabling WSL to drop its in-house registry helper in favor of wil::reg.

### DIFF
--- a/include/wil/registry_helpers.h
+++ b/include/wil/registry_helpers.h
@@ -63,26 +63,29 @@ namespace reg
     }
 
     // Access rights for opening registry keys. See https://learn.microsoft.com/en-us/windows/win32/sysinfo/registry-key-security-and-access-rights.
-    enum class key_access
+    // The named values map directly to their corresponding REGSAM access masks, so callers who need an access right not
+    // covered by a named value can supply a raw mask directly, for example key_access{KEY_QUERY_VALUE | KEY_SET_VALUE}.
+    enum class key_access : DWORD
     {
-        // Open key for reading.
-        read,
+        // Open key for reading. Equivalent to KEY_READ.
+        read = KEY_READ,
 
         // Open key for writing. Equivalent to KEY_WRITE.
-        write,
+        write = KEY_WRITE,
 
         // Open key for reading and writing. Equivalent to KEY_ALL_ACCESS.
-        readwrite,
+        readwrite = KEY_ALL_ACCESS,
 
-        // Open key for reading from 64-bit registry.
-        read64,
+        // Open key for reading from 64-bit registry. Equivalent to KEY_READ | KEY_WOW64_64KEY.
+        read64 = KEY_READ | KEY_WOW64_64KEY,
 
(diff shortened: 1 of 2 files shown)


include/wil/registry_helpers.h
```diff
@@ -84,6 +84,44 @@ namespace reg
...
+            case ::wil::reg::key_access::read64:
+                return KEY_READ | KEY_WOW64_64KEY;
+            case ::wil::reg::key_access::write64:
+                return KEY_WRITE | KEY_WOW64_64KEY;
+            case ::wil::reg::key_access::readwrite64:
+                return KEY_ALL_ACCESS | KEY_WOW64_64KEY;
+            }
+            FAIL_FAST();
+            RESULT_NORETURN_RESULT(0);
+        }
```
dunhor: Without thinking about it much... I wonder if it's time to just change `key_access` so its values just represent these values literally and then define enum flag operations for the type... That would allow us to keep using the `key_access` type and get rid of this new `access_mask` and just `static_cast` when we want the values.

dunhor: @keith-horton for his opinion since it seems @citelao is now at a different company

keith-horton: I'm personally good with Ben's change. we had long conversations about this years ago and I don't recall the reasoning specifically why we landed on this model. I'm personally good with the current model.

keith-horton (approved):
Thanks Ben!

include/wil/registry_helpers.h
dunhor: The main thing is that this is like the fourth or fifth PR to expand the set of allowed values (albeit this one does so without touching the enum, but the intent is the same). I think we have enough data to suggest that this is not the correct long-term design - it would be far more beneficial to just allow consumers to say `key_access{ FLAG1 | FLAG2... }` than have to issue a PR and wait for it to get merged/released.

keith-horton: ok. that suggestion makes sense to me. @benhillis, can you make the change that Duncan suggested?

citelao: Hi! Yes, the original intent of the enum was to give people a type-safe way of specifying these flags, since I like to add as much type-safety as possible, and I assumed a small list of modes (`read`, `readwrite`, etc) would be helpful. The fallback was (at one point, I think) to `static_cast` an int into a `::wil::reg::key_access`.

Since that abstraction is causing problems now, and people need the full flexibility, I have no issue with the proposed change from my end.

Hope that helps!

keith-horton: Hey @citelao! Looks like you found something new-and-improved after leaving Microsoft :)

This change looks good.

keith-horton (approved):
Thanks Ben!

dunhor (approved):

dunhor (approved):

PR #643, opened 2026-06-11
Add open_options support to registry open_*_key helpers

benhillis:
Adds open-with-options support to wil's registry `open_*_key` helpers.

## Motivation
Today `open_unique_key` / `open_shared_key` (and their `_nothrow` variants) hardcode `ulOptions` to `0` in their `RegOpenKeyExW` call. Callers that need a non-zero option — most notably `REG_OPTION_BACKUP_RESTORE` and `REG_OPTION_OPEN_LINK` — have no choice but to drop down to raw Win32. This is the last remaining gap preventing WSL from replacing its in-house registry helper with wil (WSL opens keys with `REG_OPTION_BACKUP_RESTORE`).

## Change
- New `wil::reg::open_options` flag enum: `none`, `open_link` (`REG_OPTION_OPEN_LINK`), `backup_restore` (`REG_OPTION_BACKUP_RESTORE`), `dont_virtualize` (`REG_OPTION_DONT_VIRTUALIZE`), with `DEFINE_ENUM_FLAG_OPERATORS`.
- Threaded as a trailing **defaulted** parameter (`= open_options::none`) through `open_unique_key`, `open_shared_key`, `open_unique_key_nothrow`, `open_shared_key_nothrow` and the internal `reg_view::open_key`. Source-compatible — existing callers are unaffected.
- Mirrors the established `key_options` precedent on the create side (kept separate because create/open option semantics differ).

## Tests
Added `BasicRegistryTests::open_options` (`[registry]`) covering:
- `static_assert` mapping of each enum value to its `REG_OPTION_*` constant and flag-OR composition.
- Default open parity (`none` behaves like a plain open).
- A volatile symbolic-link round-trip proving `open_link` returns the link key itself while a plain open follows through to the target.

Validated in both `normal` and `noexcept` configurations; full `[registry]` suite green.

Opening as **draft** for review.

### DIFF
--- a/tests/RegistryTests.cpp
+++ b/tests/RegistryTests.cpp
@@ -6539,3 +6539,173 @@ TEST_CASE("BasicRegistryTests::key_heap_string_nothrow_iterator", "[registry]")
         REQUIRE(test_iterator == test_end_iterator);
     }
 }
+
+// Creates a volatile registry symbolic link under HKEY_CURRENT_USER at |linkSubkey| that targets the (already existing)
+// key at HKEY_CURRENT_USER\|targetSubkey|. A registry symbolic link stores the absolute NT path of its target in a
+// REG_LINK value named "SymbolicLinkValue"; that path is resolved with NtQueryKey(KeyNameInformation), which is not part
+// of the public SDK and so is resolved dynamically from ntdll. Returns false if the link could not be created.
+inline bool CreateVolatileRegistrySymlink(PCWSTR targetSubkey, PCWSTR linkSubkey)
+{
+    wil::unique_hkey target;
+    if (FAILED(HRESULT_FROM_WIN32(
+            ::RegCreateKeyExW(HKEY_CURRENT_USER, targetSubkey, 0, nullptr, 0, KEY_READ | KEY_WRITE, nullptr, target.put(), nullptr))))
+    {
+        return false;
+    }
+
+    const auto ntdll = ::GetModuleHandleW(L"ntdll.dll");
+    if (!ntdll)
+    {
+        return false;
+    }
+
+    using NtQueryKey_t = LONG(__stdcall*)(HANDLE, int, PVOID, ULONG, PULONG);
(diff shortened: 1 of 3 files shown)


tests/RegistryTests.cpp
```diff
@@ -1326,8 +1326,174 @@ using ThrowingTypesToTest =
...
+    using NtQueryKey_t = LONG(__stdcall*)(HANDLE, int, PVOID, ULONG, PULONG);
+    const auto pfnNtQueryKey = reinterpret_cast<NtQueryKey_t>(::GetProcAddress(ntdll, "NtQueryKey"));
+    if (!pfnNtQueryKey)
+    {
+        return false;
+    }
+
+    // KEY_NAME_INFORMATION { ULONG NameLength; WCHAR Name[1]; }; KeyNameInformation == 3.
+    constexpr int c_KeyNameInformation = 3;
+    BYTE buffer[512]{};
```
dunhor:
```suggestion
    alignas(ULONG) BYTE buffer[512]{};
```
I believe this is technically guaranteed for all relevant ABIs for stack variables, but good hygiene

tests/RegistryTests.cpp
```diff
@@ -1326,8 +1326,174 @@ using ThrowingTypesToTest =
...
+    ULONG resultSize{};
+    if (pfnNtQueryKey(target.get(), c_KeyNameInformation, buffer, sizeof(buffer), &resultSize) < 0)
+    {
+        return false;
+    }
+
+    const auto nameLength = *reinterpret_cast<const ULONG*>(buffer);
+    const std::wstring targetNtPath(reinterpret_cast<const wchar_t*>(buffer + sizeof(ULONG)), nameLength / sizeof(wchar_t));
```
dunhor: I'm surprised this compiles outside of an `#ifdef WIL_ENABLE_EXCEPTIONS`. In my past experiences, Clang loves to complain about encountering `throw` statements when exceptions are disabled.

Issue #645, opened 2026-06-22
Query about no throw registry iterators


aursulis:
I have a question about `wil::reg::key_heap_string_nothrow_iterator` (and the `bstr` equivalent). Here's a snippet:
```c++
auto it = wil::reg::key_heap_string_nothrow_iterator((HKEY)0xdeadbeefULL); // create an iterator using an invalid handle
auto end_it = wil::reg::key_heap_string_nothrow_iterator();

for ([[maybe_unused]] const auto& key_data : wil::make_range(it, endit)) {
    std::cerr << "Did not expect this to happen!\n";
}
if (FAILED(it.last_error())) {
    std::cerr << "Error: " << std::hex << it.last_error() << "\n";
}
```
This prints:
```
Did not expect this to happen!
Error: 80070006
```
I adapted this code from the example in the source code here: https://github.com/microsoft/wil/blob/1e810ec4c86999c8da49ffb1cad33487c5b0b4c8/include/wil/registry.h#L203-L211

My question is whether the example demonstrates correct usage.
* for nothrow iterators, should I check `.last_error()` before the loop? (the example doesn't, and arguably it makes the ergonomics unnecessarily worse)
* is this perhaps a bug that the constructor doesn't go through `m_data.make_end_iterator()` on error?

dunhor:
Fixed by #646

PR #646, opened 2026-06-24
Fix error path on non-throw registry iterator construction

aursulis:
I'm pretty sure the behaviour I described on #645 is a bug. So this PR should fix it.


aursulis:
@microsoft-github-policy-service agree company="HP, Inc."

dunhor (approved):
Looks good - thanks for the report and fix!

PR #652, opened 2026-07-17
Use robots to update doc comments

dunhor:
Ensuring doc comments (1) exist, (2) are accurate, and (3) are kept up to date has not been a strong suit of ours... I'm going to start going through and let the robots clean things up. Plan is to issue one PR per header file so that the changes are not overwhelming.

Long term - would be nice to have GHCP do this on PRs, issuing change suggestions as necessary.

### DIFF
--- a/.github/skills/README.md
+++ b/.github/skills/README.md
@@ -1,0 +1,46 @@
+# Copilot CLI skills
+
+This directory holds repository-scoped [skills](https://docs.github.com/copilot) for coding agents (GitHub Copilot CLI
+and compatible tools). A *skill* is a folder containing a `SKILL.md` file that gives an agent focused, reusable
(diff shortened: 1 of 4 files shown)


jonwis (approved):
Seems ok; it'd be interesting to get the bot to update the wiki repo as well. Not sure how often people read that vs the header, but it's useful stuff.

dunhor:
> Seems ok; it'd be interesting to get the bot to update the wiki repo as well. Not sure how often people read that vs the header, but it's useful stuff.

Yes, the wiki is on the list of things to investigate. This just seemed simpler and easier to do/validate right now.

Personally, I don't care much about doxygen, however many editors use the doxygen comments when showing previews/autocomplete, so it's at least useful for that.

PR #656, opened 2026-07-21
Using robots to fix/update docs part 5: coroutine.h

dunhor:

Copilot fumbled the ball a bit with this header, so some things were touched up by hand.

I also made the judgement call to document the `task_base` type even though it exists in the `details` namespace. Otherwise, there wouldn't be any documentation for the `resume_any_thread` or `resume_same_apartment` member functions.

### DIFF
--- a/include/wil/coroutine.h
+++ b/include/wil/coroutine.h
@@ -677,9 +677,8 @@ A move-only awaitable object that can be awaited at most once; `co_await` takes 
 
 By default, awaiting a `wil::task` resumes on an arbitrary thread. Override this before the `co_await` with
 @ref resume_same_apartment() to resume in the COM apartment that was current when the await began:
-`co_await task_function().resume_same_apartment()`.
-Or change the default by converting to @ref com_task: `co_await wil::com_task(task_function())`
-Wait synchronously with @ref get(): `co_await task_function().get()`.
+`co_await task_function().resume_same_apartment()`. Or change the default by converting to @ref com_task:
+`co_await wil::com_task(task_function())`. Wait synchronously with @ref get(): `co_await task_function().get()`.
 
 The task cannot be cancelled, and an unobserved exception from an abandoned coroutine (its task destroyed without ever being
 awaited) is ignored. `wil::task` supplements PPL and C++/WinRT rather than replacing them.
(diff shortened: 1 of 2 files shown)


jonwis (approved):

include/wil/coroutine.h
```diff
@@ -618,11 +632,21 @@ struct task_base
...
+
+    //! Deleted: a task must be awaited as an rvalue, not an lvalue.
+    //! Awaiting an lvalue is rejected at compile time; use `co_await wistd::move(t)`. (Awaiting consumes the task, which is why
+    //! the task types are move-only.)
     cannot_await_lvalue_use_std_move operator co_await() & = delete;
 
-    // You must #include <synchapi.h> (usually via <windows.h>) to enable synchronous waiting.
+    //! Synchronously waits for the task to complete and returns its result.
+    //! Blocks the calling thread (the usual caveats about synchronously waiting on an STA thread apply) and must be called on an
+    //! rvalue: `wistd::move(t).get()`.
```
oldnewthing:
```suggestion
    //! rvalue: `something().get()` or `wistd::move(t).get()`.
```
The common source of an rvalue is a function call return value, not a variable that is explicitly moved-from.

include/wil/coroutine.h
```diff
@@ -634,22 +658,56 @@ struct task_base
...
 };
 } // namespace wil::details::coro
-/// @endcond
 
 namespace wil
 {
 // Must write out both classes separately
 // Cannot use deduction guides with alias template type prior to C++20.
+/** An awaitable type, whose `co_await` resumes on an arbitrary thread by default.
+A move-only awaitable object that can be awaited at most once; `co_await` takes ownership, so an lvalue must be moved in
```
oldnewthing:
```suggestion
A move-only awaitable object that can be awaited at most once; `co_await` takes ownership, so an lvalue must be moved from
```

include/wil/coroutine.h
```diff
@@ -634,22 +658,56 @@ struct task_base
...
+(`co_await wistd::move(t)`).
+
+By default, awaiting a `wil::task` resumes on an arbitrary thread. Override this before the `co_await` with
+@ref resume_same_apartment() (resume in the COM apartment that was current when the await began), or change the default by
+converting to @ref com_task. Wait synchronously with @ref get().
```
oldnewthing:
```suggestion
@ref resume_same_apartment() to resume in the COM apartment that was current when the await began:
`co_await task_function().resume_same_apartment()`.
Or change the default by converting to @ref com_task: `co_await wil::com_task(task_function())`
Wait synchronously with @ref get(): `co_await task_function().get()`.
```

include/wil/coroutine.h
oldnewthing:
```suggestion
 *   in which case unobserved exceptions are ignored.
```

There was a late change to the implementation and this part of the docs was not updated. Note that abandon() does not std::terminate() any more. This is necessary to avoid fail-fasts when a coroutine throws an exception before it can run down its active tasks.

```cpp
auto task = task_function();
do_other_stuff_in_parallel(); // throws!
co_await std::move(task);
```

include/wil/coroutine.h
```diff
@@ -634,22 +658,56 @@ struct task_base
...
+
+The task cannot be cancelled, and an unobserved exception from an abandoned coroutine (its task destroyed without ever being
+awaited) is fatal. `wil::task` supplements PPL and C++/WinRT rather than replacing them.
```
oldnewthing:
```suggestion
awaited) is ignored. `wil::task` supplements PPL and C++/WinRT rather than replacing them.
```

include/wil/coroutine.h
```diff
@@ -664,16 +722,50 @@ struct task : details::coro::task_base<T>
...
 
+    //! Awaits the task, resuming the caller on an arbitrary thread (shorthand for `resume_any_thread()`).
     auto operator co_await() && noexcept
     {
         return wistd::move(*this).resume_any_thread();
     }
 };
 
+/** An awaitable type, whose `co_await` resumes in the same COM apartment by default.
+A move-only awaitable object that can be awaited at most once; `co_await` takes ownership, so an lvalue must be moved in
```
oldnewthing: Please incorporate other changes mutatis mutandis from wil::task.

include/wil/coroutine.h
dunhor: I _almost_ made that change myself, but talked myself out of it for some reason.

include/wil/coroutine.h
dunhor: Note that the call patterns are present in the code example, but good to explicitly inline it too

ChrisGuzak (approved):

PR #657, opened 2026-07-22
Using robots to fix/update docs part 6: cppwinrt* headers

dunhor:

Figured that I should probably group all the C++/WinRT headers into a single change. This is another one Copilot had a bit of trouble producing good/correct documentation to & required some TLC.

Additionally, while doing this I noticed a number of potential issues. I'll note them in comments as follow-ups; I want to leave these changes as doc updates only.

### DIFF
--- a/include/wil/cppwinrt.h
+++ b/include/wil/cppwinrt.h
@@ -224,6 +224,7 @@ inline HRESULT __stdcall ResultFromCaughtException_CppWinRt(
 
 namespace wil
 {
+/// @cond
 inline std::int32_t __stdcall winrt_to_hresult(void* returnAddress) noexcept
 {
     // C++/WinRT only gives us the return address (caller), so pass along an empty 'DiagnosticsInfo' since we don't
@@ -255,6 +256,7 @@ inline void WilInitialize_CppWinRT()
         }
     }
 }
+/// @endcond
 
 /// @cond
 namespace details
@@ -271,52 +273,97 @@ namespace details
 } // namespace details
 /// @endcond
 
-// Provides an overload of verify_hresult so that the WIL macros can recognize winrt::hresult as a valid "hresult" type.
+//! Lets WIL's error-handling macros accept a `winrt::hresult` where they expect an `HRESULT`.
+//! This overload is found by the `RETURN_IF_FAILED` / `THROW_IF_FAILED` family (and friends) so a `winrt::hresult` result can be
+//! checked directly, without first converting it to an `HRESULT`.
+//! @param hr The `winrt::hresult` to validate.
+//! @return `hr`, as an `HRESULT`-compatible value.
 inline long verify_hresult(winrt::hresult hr) noexcept
 {
     return hr;
 }
 
-// Provides versions of get_abi and put_abi for genericity that directly use HSTRING for convenience.
+//! Returns the raw ABI representation of a C++/WinRT projected object, for passing across the ABI boundary in generic code.
+//! This overload set exists mainly because `winrt::get_abi` no longer references any Windows SDK types: for a `winrt::hstring` it
+//! returns a bare `void*` rather than an `HSTRING`, so WIL provides these as an alternative that at least knows about `HSTRING`
+//! (the `winrt::hstring` overload returns the concrete `HSTRING`). For every other type it simply forwards to `winrt::get_abi`.
+//! For a projected interface or runtime class `winrt::impl::abi_t<T>` resolves to a `void*` ABI interface pointer; for a
+//! projected value type it is instead the `winrt::impl::abi_t<T>` ABI struct, by value (layout-compatible with, but distinct
+//! from, the classic/SDK `ABI::` type). Paired with @ref put_abi.
+//! @tparam T A C++/WinRT projected type.
+//! @param object The object whose ABI representation to return.
+//! @return For an interface or runtime class, a borrowed (not AddRef'd) `void*` interface pointer; for a value type, the
+//!         `winrt::impl::abi_t<T>` ABI struct by value.
 template <typename T>
 auto get_abi(T const& object) noexcept
 {
     return winrt::get_abi(object);
 }
 
+//! Returns the raw `HSTRING` that backs a `winrt::hstring`.
+//! @param object The string whose `HSTRING` to return.
+//! @return The `HSTRING` backing `object` (may be null for an empty string).
 inline auto get_abi(winrt::hstring const& object) noexcept
 {
     return static_cast<HSTRING>(winrt::get_abi(object));
 }
 
+//! Returns a null-terminated wide-character pointer to a `winrt::hstring`'s buffer.
+//! This is the `winrt::hstring` overload of WIL's generic `str_raw_ptr` set, which normalizes the string types WIL supports (raw
+//! `PCWSTR`, `std::wstring`, `HSTRING`, and so on) to a `PCWSTR` for use by WIL's generic string helpers.
+//! @param str The string to read from.
+//! @return A `PCWSTR` to `str`'s contents, valid for as long as `str` is alive and unmodified.
 inline auto str_raw_ptr(const winrt::hstring& str) noexcept
 {
     return str.c_str();
 }
 
+//! Returns the address of a C++/WinRT object's ABI pointer, for receiving it as an out-parameter in generic code.
+//! Like @ref get_abi, this exists mainly because `winrt::put_abi` no longer references any Windows SDK types: for a
+//! `winrt::hstring` it yields a bare `void**` rather than an `HSTRING*`, so the `winrt::hstring` overload here yields the
+//! concrete `HSTRING*`. For every other type it forwards to `winrt::put_abi`. For a projected interface or runtime class the
+//! result is `void**` (the address of the object's `void*` ABI interface pointer), and the held interface is released first, so
+//! the object need not already be empty; for a projected value type it is `winrt::impl::abi_t<T>*`. Paired with @ref get_abi.
+//! @tparam T A C++/WinRT projected type.
+//! @param object The object that receives the ABI value; for an interface or runtime class, any held interface is released first.
+//! @return For an interface or runtime class, a `void**`; for a value type, a `winrt::impl::abi_t<T>*`.
 template <typename T>
 auto put_abi(T& object) noexcept
 {
     return winrt::put_abi(object);
 }
 
+//! Returns the address of a `winrt::hstring`'s backing `HSTRING`, as an `HSTRING*`.
+//! @param object The (empty) string that receives the `HSTRING`.
+//! @return The address of `object`'s backing `HSTRING`, suitable as an out-parameter.
 inline auto put_abi(winrt::hstring& object) noexcept
 {
     return reinterpret_cast<HSTRING*>(winrt::put_abi(object));
 }
 
+//! Returns the raw `IUnknown` ABI pointer for a C++/WinRT object, without taking a reference.
+//! @param ptr The object whose ABI pointer to return.
+//! @return The `IUnknown*` backing `ptr` (borrowed, not AddRef'd; may be null).
 inline ::IUnknown* com_raw_ptr(const winrt::Windows::Foundation::IUnknown& ptr) noexcept
 {
     return static_cast<::IUnknown*>(winrt::get_abi(ptr));
 }
 
-// Needed to power wil::cx_object_from_abi that requires IInspectable
+//! Returns the raw `IInspectable` ABI pointer for a C++/WinRT object, without taking a reference.
+//! This overload supplies the `IInspectable` that `wil::cx_object_from_abi` requires.
+//! @param ptr The object whose ABI pointer to return.
+//! @return The `IInspectable*` backing `ptr` (borrowed, not AddRef'd; may be null).
 inline ::IInspectable* com_raw_ptr(const winrt::Windows::Foundation::IInspectable& ptr) noexcept
 {
     return static_cast<::IInspectable*>(winrt::get_abi(ptr));
 }
 
-// Taken from the docs.microsoft.com article
+//! Constructs a C++/WinRT projected object from a raw ABI interface pointer by querying for the projected type.
+//! Calls `QueryInterface` for `T`'s default interface and returns the resulting projection.
+//! @tparam T The C++/WinRT projected type to construct.
+//! @param from The interface pointer to convert from; must not be null.
+//! @return A `T` referring to the same object as `from`.
+//! @note Throws a `winrt::hresult_error` if the `QueryInterface` call fails.
 template <typename T>
 T convert_from_abi(::IUnknown* from)
 {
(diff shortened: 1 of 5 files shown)


include/wil/cppwinrt.h
```diff
@@ -271,52 +273,97 @@ namespace details
...
 
-// Needed to power wil::cx_object_from_abi that requires IInspectable
+//! Returns the raw `IInspectable` ABI pointer for a C++/WinRT object, without taking a reference.
+//! This overload supplies the `IInspectable` that `wil::cx_object_from_abi` requires.
+//! @param ptr The object whose ABI pointer to return.
+//! @return The `IInspectable*` backing `ptr` (borrowed, not AddRef'd; may be null).
 inline ::IInspectable* com_raw_ptr(const winrt::Windows::Foundation::IInspectable& ptr) noexcept
 {
     return static_cast<::IInspectable*>(winrt::get_abi(ptr));
 }
```
dunhor: One thing that might be worth noting is that these are not identity pointers (two pointers to the same object might not compare equal) - that's something that would require a QI.

include/wil/cppwinrt.h
dunhor, line 459: It's potentially concerning that this type does not have an `operator=` and I'm _pretty_ sure it gets the compiler generated default, which is NOT what we probably want here.

It might be worth doing a preliminary investigation where we explicitly `= delete` the copy assignment operator to see if anyone is using it, and if they are, either they have a bug we'll be fixing OR they happen to function correctly (perhaps by accident, perhaps by explicitly working around this limitation) and we may need to be cautious not to break them.

include/wil/cppwinrt_authoring.h
dunhor, line 532: This should be `this->m_propertyChanged` because the member comes from `notify_property_changed_base`, however that type uses CRTP and if the derived type is itself a template, then the base becomes dependent and this will fail under proper two phase name lookup.

include/wil/cppwinrt.h
oldnewthing: Maybe, but that's also just a fundamental rule of COM. To test object identity, you must use the canonical IUnknown.

Wait, these are just overloads to the existing com_raw_ptr to add C++/WinRT support. We should defer to com_raw_ptr for caveats etc.

include/wil/cppwinrt.h
oldnewthing: The default assignment operator is a nop (there are no members to assign), which is correct behavior. The assigned-to and assigned-from objects remain unchanged. Each incremented the module lock when they were respectively constructed, and they will decrement when respectively destructed.

In fact, the `operator=` must *not* be deleted because that would prevent assignability of derived classes, such as in the example given in other parts of the docs:
```cpp
struct NonWinrtObject : wil::winrt_module_reference
{
    int value;
};
```

include/wil/cppwinrt.h
```diff
@@ -271,52 +273,97 @@ namespace details
 } // namespace details
 /// @endcond
 
-// Provides an overload of verify_hresult so that the WIL macros can recognize winrt::hresult as a valid "hresult" type.
+//! Lets WIL's error-handling macros accept a `winrt::hresult` where they expect an `HRESULT`.
+//! This overload is found by the `RETURN_IF_FAILED` / `THROW_IF_FAILED` family (and friends) so a `winrt::hresult` result can be
+//! checked directly, without first converting it to an `HRESULT`.
+//! @param hr The `winrt::hresult` to validate.
+//! @return `hr`, as an `HRESULT`-compatible value.
```
oldnewthing: This is a lot of documentation for a function people aren't expected to call directly.
Maybe just
```cpp
//! Adds `winrt::hresult` support to the `verify_hresult` helper function.
//! This allows macros like `RETURN_IF_FAILED` to be used with `winrt::hresult`.
```

include/wil/cppwinrt.h
dunhor: Oh, right, duh. It's not possible to "steal" a reference from another object or construct/release/reset one, so there's nothing to do there.

include/wil/cppwinrt.h
```diff
@@ -271,52 +273,97 @@ namespace details
...
+//! This overload set exists mainly because `winrt::get_abi` no longer references any Windows SDK types: for a `winrt::hstring` it
+//! returns a bare `void*` rather than an `HSTRING`, so WIL provides these as an alternative that at least knows about `HSTRING`
+//! (the `winrt::hstring` overload returns the concrete `HSTRING`). For every other type it simply forwards to `winrt::get_abi`.
+//! For a projected interface or runtime class `winrt::impl::abi_t<T>` resolves to a `void*` ABI interface pointer; for a
+//! projected value type it is instead the `winrt::impl::abi_t<T>` ABI struct, by value (layout-compatible with, but distinct
+//! from, the classic/SDK `ABI::` type). Paired with @ref put_abi.
+//! @tparam T A C++/WinRT projected type.
+//! @param object The object whose ABI representation to return.
+//! @return For an interface or runtime class, a borrowed (not AddRef'd) `void*` interface pointer; for a value type, the
+//!         `winrt::impl::abi_t<T>` ABI struct by value.
```
oldnewthing: Too much inside baseball. How about
```cpp
//! Returns the raw ABI representation of a C++/WinRT projected object.
//! This is an improvement over `winrt::get_abi` because it returns `HSTRING`
//! rather than `void*` when passed a `winrt::hstring`.
```

include/wil/cppwinrt.h
dunhor: I'll leave undocumented for now under the "this should be obvious to the consumer" assumption.

include/wil/cppwinrt.h
```diff
@@ -271,52 +273,97 @@ namespace details
...
 template <typename T>
 auto get_abi(T const& object) noexcept
 {
     return winrt::get_abi(object);
 }
 
+//! Returns the raw `HSTRING` that backs a `winrt::hstring`.
```
oldnewthing:
```cpp
//! Overload of `wil::get_abi` for `winrt::hstring` which returns the underlying ABI `HSTRING`.
```

include/wil/cppwinrt.h
```diff
@@ -271,52 +273,97 @@ namespace details
...
+//! @param object The string whose `HSTRING` to return.
+//! @return The `HSTRING` backing `object` (may be null for an empty string).
 inline auto get_abi(winrt::hstring const& object) noexcept
 {
     return static_cast<HSTRING>(winrt::get_abi(object));
 }
 
+//! Returns a null-terminated wide-character pointer to a `winrt::hstring`'s buffer.
```
oldnewthing:
```
//! Adds `winrt::hstring` support to the `str_raw_ptr` function.
//! See `str_raw_ptr` for details.
```

include/wil/cppwinrt.h
```diff
@@ -271,52 +273,97 @@ namespace details
...
+//! This is the `winrt::hstring` overload of WIL's generic `str_raw_ptr` set, which normalizes the string types WIL supports (raw
+//! `PCWSTR`, `std::wstring`, `HSTRING`, and so on) to a `PCWSTR` for use by WIL's generic string helpers.
+//! @param str The string to read from.
+//! @return A `PCWSTR` to `str`'s contents, valid for as long as `str` is alive and unmodified.
 inline auto str_raw_ptr(const winrt::hstring& str) noexcept
 {
     return str.c_str();
 }
 
+//! Returns the address of a C++/WinRT object's ABI pointer, for receiving it as an out-parameter in generic code.
```
oldnewthing: Again, talking too much.

```cpp
//! Returns the address of a C++/WinRT object's ABI pointer, for receiving it as an out-parameter.
//! This is an improvement over `winrt::get_abi` because it returns a pointer to `HSTRING`
//! rather than a pointer to `void*` when passed a `winrt::hstring`.
```

include/wil/cppwinrt.h
```diff
@@ -271,52 +273,97 @@ namespace details
...
+//! @tparam T A C++/WinRT projected type.
+//! @param object The object that receives the ABI value; for an interface or runtime class, any held interface is released first.
+//! @return For an interface or runtime class, a `void**`; for a value type, a `winrt::impl::abi_t<T>*`.
 template <typename T>
 auto put_abi(T& object) noexcept
 {
     return winrt::put_abi(object);
 }
 
+//! Returns the address of a `winrt::hstring`'s backing `HSTRING`, as an `HSTRING*`.
```
oldnewthing:
```cpp
//! Overload of `wil::put_abi` for `winrt::hstring` which returns a pointer to the underlying ABI `HSTRING`.
```

include/wil/cppwinrt.h
```diff
@@ -271,52 +273,97 @@ namespace details
...
+//! @param object The (empty) string that receives the `HSTRING`.
+//! @return The address of `object`'s backing `HSTRING`, suitable as an out-parameter.
 inline auto put_abi(winrt::hstring& object) noexcept
 {
     return reinterpret_cast<HSTRING*>(winrt::put_abi(object));
 }
 
+//! Returns the raw `IUnknown` ABI pointer for a C++/WinRT object, without taking a reference.
```
oldnewthing:
```cpp
//! Adds C++/WinRT support to the `com_raw_ptr` function.
//! See `com_raw_ptr` for details.
```

include/wil/cppwinrt.h
```diff
@@ -271,52 +273,97 @@ namespace details
...
+//! @param ptr The object whose ABI pointer to return.
+//! @return The `IUnknown*` backing `ptr` (borrowed, not AddRef'd; may be null).
 inline ::IUnknown* com_raw_ptr(const winrt::Windows::Foundation::IUnknown& ptr) noexcept
 {
     return static_cast<::IUnknown*>(winrt::get_abi(ptr));
 }
 
-// Needed to power wil::cx_object_from_abi that requires IInspectable
+//! Returns the raw `IInspectable` ABI pointer for a C++/WinRT object, without taking a reference.
```
oldnewthing:
```cpp
//! Adds C++/WinRT support to the `com_raw_ptr` function.
//! See `com_raw_ptr` for details.
```

include/wil/cppwinrt.h
```diff
@@ -271,52 +273,97 @@ namespace details
...
+//! This overload supplies the `IInspectable` that `wil::cx_object_from_abi` requires.
+//! @param ptr The object whose ABI pointer to return.
+//! @return The `IInspectable*` backing `ptr` (borrowed, not AddRef'd; may be null).
 inline ::IInspectable* com_raw_ptr(const winrt::Windows::Foundation::IInspectable& ptr) noexcept
 {
     return static_cast<::IInspectable*>(winrt::get_abi(ptr));
 }
 
-// Taken from the docs.microsoft.com article
+//! Constructs a C++/WinRT projected object from a raw ABI interface pointer by querying for the projected type.
```
oldnewthing: This doesn't construct an object. It just gets the C++/WinRT representation of an existing object.

```cpp
//! Converts a raw ABI interface pointer to a C++/WinRT projected object.
//! @tparam T the C++/WinRT projected type to convert to.
//! @param from The interface pointer to convert from. Must not be null. The function does not take ownership of this pointer.
//! @returns A `T` that represents the same object as `from`.
//! @note Throws a `winrt::hresult_error` if the conversion fails.
```

include/wil/cppwinrt.h
```diff
@@ -325,19 +372,43 @@ T convert_from_abi(::IUnknown* from)
     return result;
 }
 
-// For obtaining an object from an interop method on the factory. Example:
-// winrt::InputPane inputPane = wil::capture_interop<winrt::InputPane>(&IInputPaneInterop::GetForWindow, hwnd);
-// If the method produces something different from the factory type:
-// winrt::IAsyncAction action = wil::capture_interop<winrt::IAsyncAction, winrt::AccountsSettingsPane>(&IAccountsSettingsPaneInterop::ShowAddAccountForWindow, hwnd);
+/** Calls an interop method on a C++/WinRT type's activation factory and returns the object it produces.
+Retrieves the activation factory for `WinRTFactory`, queries it for `Interface`, invokes `method` with `args`, and wraps the
+result as a `WinRTResult`. Use this with factory interop interfaces such as `IInputPaneInterop`.
```
oldnewthing: Don't need to say how it does it.
```suggestion
Use this with factory interop interfaces such as `IInputPaneInterop`.
```

include/wil/cppwinrt.h
```diff
@@ -325,19 +372,43 @@ T convert_from_abi(::IUnknown* from)
...
 {
     auto interop = winrt::get_activation_factory<WinRTFactory, Interface>();
     return winrt::capture<WinRTResult>(interop, method, std::forward<Args>(args)...);
 }
 
-// For obtaining an object from an interop method on an instance. Example:
-// winrt::UserActivitySession session = wil::capture_interop<winrt::UserActivitySession>(activity, &IUserActivityInterop::CreateSessionForWindow, hwnd);
+/** Calls an interop method on an existing C++/WinRT object and returns the object it produces.
+Queries `obj` for `Interface`, invokes `method` with `args`, and wraps the result as a `WinRTResult`. Use this with per-instance
+interop interfaces such as `IUserActivityInterop`.
```
oldnewthing:
```suggestion
Use this with per-instance interop interfaces such as `IUserActivityInterop`.
```

include/wil/cppwinrt.h
```diff
@@ -403,15 +474,18 @@ winrt::fire_and_forget ContinueBackgroundWork()
 */
 struct [[nodiscard]] winrt_module_reference
 {
+    //! Takes a reference on the host C++/WinRT module, incrementing its lock count.
     winrt_module_reference()
     {
         ++winrt::get_module_lock();
     }
 
+    //! Takes an additional reference on the host module; every live copy holds its own lock.
```
oldnewthing:
```suggestion
    //! Takes a reference on the host C++/WinRT, module, incrementing its lock count.
```

include/wil/cppwinrt.h
```diff
@@ -462,6 +536,11 @@ struct winrt_conditionally_implements : Implements
 {
     using Implements::Implements;
 
+    //! Overrides C++/WinRT's interface lookup so conditionally-implemented interfaces are hidden when disabled.
```
oldnewthing: This method is not meant to be called by clients. Should remain an internal method, not a documented one.

include/wil/cppwinrt_authoring.h
```diff
@@ -117,9 +160,25 @@ struct single_threaded_rw_property : single_threaded_property<T>
 #define __WIL_CPPWINRT_AUTHORING_INCLUDED_ICLASSFACTORY
 /// @endcond
 
+/** A COM class factory (`IClassFactory`) that creates instances of a C++/WinRT (`winrt::implements`) class.
+Creates instances of `T` with `winrt::make_self<T>`, and implements `LockServer` against the C++/WinRT module lock so the host
+module stays loaded while the factory is locked. Hand it to whatever needs an `IClassFactory` (for example
+`CoRegisterClassObject`, or a `DllGetClassObject` implementation); @ref register_com_server wraps the `CoRegisterClassObject` case
+for you.
+@tparam T The C++/WinRT class the factory creates.
+@tparam Rest Additional C++/WinRT `implements` markers for the factory (not interfaces to implement), appended to its
```
oldnewthing: Why not interfaces for the factory to implement?

include/wil/cppwinrt_authoring.h
```diff
@@ -117,9 +160,25 @@ struct single_threaded_rw_property : single_threaded_property<T>
...
 template <typename T, typename... Rest>
 struct class_factory : winrt::implements<class_factory<T, Rest...>, IClassFactory, winrt::no_weak_ref, Rest...>
 {
+    //! Creates an instance of `T` and returns the requested interface.
+    //! Implements `IClassFactory::CreateInstance`. Aggregation is not supported, so a non-null `outer` fails with
+    //! `CLASS_E_NOAGGREGATION`.
+    //! @param outer The aggregating object's controlling `IUnknown`; must be null, since aggregation is unsupported.
+    //! @param iid Interface to return in `result`.
+    //! @param result Receives the requested interface on success, or null on failure.
+    //! @return `S_OK` on success; `CLASS_E_NOAGGREGATION` if `outer` is non-null; otherwise a failure `HRESULT`.
```
oldnewthing: Don't document this publicly. Clients do not call it directly. This is implementing a method that is called by COM itself.

include/wil/cppwinrt_authoring.h
dunhor: While this type is not `final`, it doesn't seem like the intent is for consumers to derive from it (which would be required if they were to list interfaces here). I suppose they could supply types that implement the interface methods, but then the comment would still _technically_ be correct.

I can augment this comment to just say it's "typically" used for markers and keep it intentionally semi-vague.

include/wil/cppwinrt.h
dunhor: It's useful when trying to figure out how to call the function, however with the example and explanation of template arguments, I agree that this section can be more brief.

include/wil/cppwinrt.h
dunhor: Consequence of using "no warnings" as a termination condition. I should let it be more aggressive when adding `@cond`/`@endcond` blocks

ChrisGuzak (approved):

PR #658, opened 2026-07-28
Using robots to fix/update docs part 7: filesystem.h

dunhor:

This also incorporates some feedback from #657 to relax documentation requirements, particularly on overloads and more "implementation detail" types/functions

### DIFF
--- a/include/wil/filesystem.h
+++ b/include/wil/filesystem.h
@@ -33,14 +33,22 @@
 namespace wil
 {
 //! Determines if a path is an extended length path that can be used to access paths longer than MAX_PATH.
+//! @param path The path to inspect.
+//! @return `true` if `path` begins with the `\\?\` extended-length prefix; otherwise `false`.
 inline bool is_extended_length_path(_In_ PCWSTR path)
 {
     return wcsncmp(path, L"\\\\?\\", 4) == 0;
 }
 
 #if (_WIN32_WINNT >= _WIN32_WINNT_WIN7)
-//! Find the last segment of a path. Matches the behavior of shlwapi!PathFindFileNameW()
-//! note, does not support streams being specified like PathFindFileNameW(), is that a bug or a feature?
+//! Finds the last segment of a path.
+//! Returns a pointer just past the last backslash in `path`, ignoring a single trailing backslash (which is kept as part of the
+//! returned segment), so `C:\dir\file` yields `file` while `C:\dir\sub\` yields `sub\`. If `path` has no backslash (apart from a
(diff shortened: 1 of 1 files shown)


include/wil/filesystem.h
```diff
@@ -430,73 +488,87 @@ struct next_entry_offset_iterator
...
-    T* current_;
+    T* current_; //!< Pointer to the current element, or null for the end iterator.
 };
 
+//! Creates a @ref next_entry_offset_iterator over the `NextEntryOffset`-chained range beginning at `p`.
+//! @tparam T The element type; it must expose a `NextEntryOffset` member.
+//! @param p Pointer to the first element in the buffer.
+//! @return A @ref next_entry_offset_iterator that can be used directly in a range-based `for` loop.
 template <typename T>
 next_entry_offset_iterator<T> create_next_entry_offset_iterator(T* p)
```
jonwis, line 529: It'd be neat to add a type-deduction helper here to say `for (auto const& q : next_entry_offset_iterator(r.get()))` ... no change required.

include/wil/filesystem.h
```diff
@@ -1295,11 +1479,8 @@ inline wil::unique_hfile open_or_create_file(
...
 }
 
-/** create using CREATE_ALWAYS, returns the file handle, throws on error.
-~~~
-auto handle = wil::open_or_truncate_existing_file(filePath.c_str());
-~~~
-*/
+/** Creates a new file or truncates an existing one using `CREATE_ALWAYS`, throwing on failure.
+The exception-based counterpart to @ref try_open_or_truncate_existing_file; see that function for details. */
 inline wil::unique_hfile open_or_truncate_existing_file(
```
jonwis, line 1484: Seems like this is `create_or_truncate_existing_file` ? It either creates the file with zero length, or opens one and sets its length to zero first...

jonwis (approved):

include/wil/filesystem.h
dunhor: From what I understand, a deduction guide isn't actually necessary here and that should just work in C++17 and beyond. E.g. the following compiles fine:
```c++
template <typename T>
struct test
{
    test(T*);
};

void doit()
{
    int x;
    test t(&x);
}
```

include/wil/filesystem.h
dunhor: While I agree that's probably a better name, a search across GitHub shows that this function is used in existing projects. I can file a follow-up issue to rename this function and add a forwarding deprecated function with this name if you'd like.

PR #665, opened 2026-08-13
basic_zstring_view: preserve tail substr and backport contains

mmthomas:
## Summary

- Add a one-argument `basic_zstring_view::substr(pos)` overload that returns `basic_zstring_view`, since a tail of a null-terminated view remains null-terminated.
- Add an explicit two-argument `substr(pos, count)` overload returning `std::basic_string_view`, since an arbitrary slice may not be null-terminated.
- Backport the C++23 `contains()` overloads for earlier language modes, gated on `__cpp_lib_string_contains`.

This is the focused, independently reviewable subset discussed in #635. The non-null variant and its traits design are intentionally excluded and can be considered separately.

## Validation

- MSVC Debug C++17 `witest.exe "[zstring_view]"`: 96 assertions passed.
- MSVC Debug C++23 `witest.cpplatest.exe "[zstring_view]"`: 98 assertions passed.

### DIFF
--- a/include/wil/stl.h
+++ b/include/wil/stl.h
@@ -166,6 +166,9 @@ inline wil::unique_bstr make_bstr(std::wstring_view source)
     * A zstring_view can be converted implicitly to a std::string_view because it is always safe to use a nul-terminated
       string_view as a plain string view.
     * A zstring_view can be constructed from a std::string because the data in std::string is nul-terminated.
+    * substr(pos) returns a zstring_view because the tail remains nul-terminated. substr(pos, count) returns a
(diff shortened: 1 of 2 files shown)


mmthomas:
@dunhor @dmachaj @jonwis — this is the focused follow-up to #635, limited to the `substr` overload changes and the `contains()` polyfill. The non-null variant and traits design have been left out as discussed. When you have time, I’d appreciate another look.

dmachaj (approved):
:shipit:

dunhor (approved):

PR #666, opened 2026-08-13
Reserve apartment contexts before push_back to reduce reallocations

zengelbrecht61-lgtm:
## Summary

Pre-size the apartment context collection before pushing entries into the vector so it avoids repeated reallocation while the apartment lock is held.

## Why

The apartment migration code collects context pointers and then appends them back into a vector each time it enumerates the apartment set. The vector grew one element at a time, which caused repeated reallocation and copy churn under the lock.

## What changed

- computed the total number of contexts to capture once
- reserved that capacity before push_back
- left the rest of the logic unchanged

## Validation

- builds and existing WIL tests continue to pass with the change


dunhor:
@ChrisGuzak IIRC you wrote most of this code, so you'd be able to better comment on this than me.

From the code, not _all_ contexts are pushed into the vector, so this is likely going to over-allocate. That's probably fine - vector is likely going to over-allocate when it doubles its size the final time - however your description is missing the very important _why_? What problem are you trying to solve?

